PR #64, opened 2015-10-13
extracted common context from agent repo

elf-pavlik:


fosterlynn:
@elf-pavlik I happily merged this, but I have a technical question on it that I'm finally getting around to asking: How does it work that we reference 

   "vf": "https://w3id.org/valueflows/",

inside what I think is our context?  Seems circular.

elf-pavlik:
we define a *prefix* "vf" and then use it in [CURIEs](http://www.w3.org/TR/curie/) like "vf:subject" - which expands to "https://w3id.org/valueflows/subject"

all vocabulary terms need to map to URI and thanks to prefixes we just have a shorthand method for writing them

fosterlynn:
Oh I get it, thanks!  Not telling you to go there for the vocab definitions, just defining the "vf" shorthand.  Actually that clears up a few things for me.  (Duh I guess.)

PR #65, opened 2015-10-13
added role vocab - example of generic properties

elf-pavlik:
this will prevent need for defining properties like 'member', 'contributor', 'mentor' on vf: level and give us chance to experiment with integration of 'vertical' JSON-LD contexts

### DIFF
--- a/context.jsonld
+++ b/context.jsonld
@@ -1,19 +1,23 @@
 {
-  "@context": {
-    "owl": "http://www.w3.org/2002/07/owl#",
-    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
-    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
-    "rdfs:label": { "@container": "@language" },
-    "vf": "https://w3id.org/valueflows/",
-    "Relationship": "vf:Relationship",
-    "url":  { "@id": "vf:url", "@type": "@id" },
-    "image": { "@id": "vf:image", "@type": "@id" },
-    "displayName": "vf:displayName",
-    "displayNameMap": { "@id": "displayName", "@container": "@language" },
-    "subject": { "@id": "vf:subject", "@type": "@id" },
-    "relationship": { "@id": "vf:relationship", "@type": "@id" },
-    "object": { "@id": "vf:object", "@type": "@id" }
-  },
+  "@context": [
+    "https://w3id.org/role/",
+    {
+      "owl": "http://www.w3.org/2002/07/owl#",
+      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+      "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+      "rdfs:label": { "@container": "@language" },
+      "role": "https://w3id.org/role/",
+      "vf": "https://w3id.org/valueflows/",
+      "Relationship": "vf:Relationship",
+      "url":  { "@id": "vf:url", "@type": "@id" },
+      "image": { "@id": "vf:image", "@type": "@id" },
+      "displayName": "vf:displayName",
+      "displayNameMap": { "@id": "displayName", "@container": "@language" },
+      "subject": { "@id": "vf:subject", "@type": "@id" },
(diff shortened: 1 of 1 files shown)


ahdinosaur:
eh. there's a few reasons i'm skeptical of this, related to my feelings in https://github.com/valueflows/agent/issues/38#issuecomment-146721078, but also:

-1 for "grab-bags", or domain-less / global vocabs. it's a pattern that comes up in javascript module ecosystems, but most recommend not to do it (e.g. [mattdesl/module-best-practices](https://github.com/mattdesl/module-best-practices)). if we're [going to support common relationship verbs](https://github.com/valueflows/agent/issues/38), i'd rather we not step outside of our domain / local scope.

-0 for "role". in Holodex we're trying to [move away from role nouns and move towards relationship verbs](https://github.com/holodex/enspiral-data/blob/compost/relationshipTypes.csv). but this is still a field that needs more experiments before i have a definite opinion.

bhaugen:
I think this is related to domain ontologies vs upper ontologies (those that start with something like Thing).

REA is a domain ontology.  Bill McCarthy has been adamant about that from 1980 or so. You may want to connect it to an upper ontology, and he might even help you, but upper ontologies come and go, and REA will remain a domain ontology.

The history of projects that started by trying to define an upper ontology (e.g. Cyc) is many years, little progress.

elf-pavlik:
@ahdinosaur can you make alternative PR which proposes **one stop shop** for generic relationships for Agent <-> Agent and Agent <-> Resource to start with? I see need for them in examples we add, and while I would encourage people to experiment with various *snowflake terms* defined in their namespaces (not commonly shared ones) and not used by other systems, I also would like to include in our work use of some shared vocabs which define relationships (precisely `"@type": ["rdf:Property", "owl:ObjectProperty"]`)

Once again, we still work on `0.?` stage and shouldn't consider things we add as stable by any means, I see it mostly as removing road blocks for adding examples and building prototype software. I would really like that we try as much as possible let examples and code using them drive our work, and when we don't agree with someone's proposal we do our best to come up with alternative one and focus on discussing alternative proposed approaches until we all feel comfortable with *giving one of them a try* by using them in our examples and code (which all have version control and we can go back and explore other path if we don't want to continue on one we decided to try)

bhaugen:
@elf-pavlik - both Lynn and I are pushing back against the idea of assuming that Agent <-> Agent and Agent <-> Resource relationships are the same, and do not need to constrain both the domain and the range. "Snowflake" is pejorative in that issue. While we have some of those in living networks, that is a separate issue from the semantics of Agent:Agent vs Agent:Resource. I will start writing examples today.

elf-pavlik:
> both Lynn and I are pushing back against the idea of assuming that Agent <-> Agent and Agent <-> Resource relationships are the same, and do not need to constrain both the domain and the range. "Snowflake" is pejorative in that issue.

@bhaugen IMO we need to start somewhere, so far we have [only one .jsonld file with example of a resource](https://github.com/valueflows/resource/tree/master/examples) and it's relationship to agents. I will add few more .jsonld files with resources and agents, most likely starting with [ouikit](https://github.com/valueflows/valueflows/blob/master/use-cases/ouikit.md). This will give us possibility to work on **concrete, real world examples**. I would also encourage you to also 'hand write' JSON-LD based on use cases from your system, besides not needing to wait for writing auto generation code it can offer opportunity to exercise using this serialization.

While I understand various worries and concerns about different approaches each of us can suggest here. The most constructive approach (which I can currently think of) to work on them, requires **proposing** modeling by providing .jsonld examples which use constructs which so far we can agree on. I also do expect that we will need to go sometimes little back and forth to try things out and learn what we see working and what we see not working.

I will happily close this PR without merging it as soon as someone proposes a clear alternative supported by .jsonld example file(s) which use it.

bhaugen:
As I wrote above and elsewhere, I will start writing **concrete, real world examples** today,  but I will start with fairly natural language descriptions. I am not yet competent to hand write json-LD.

elf-pavlik:
I'll just define "role" prefix locally in .jsonld examples which I contribute ...

Issue #34, opened 2015-10-06
consistent labelling system across our vocab


ahdinosaur:
as mentioned in https://github.com/valueflows/agent/pull/30#issuecomment-145809856 and https://github.com/valueflows/agent/issues/9#issuecomment-144392445, i think we should aim for a consistent labelling system across all our domains. i'm bringing this issue up here and now so i don't block https://github.com/valueflows/agent/pull/30 and https://github.com/valueflows/agent/issues/31.

features we need in our labels:
- primary names
- [nick names](https://github.com/valueflows/agent/issues/9#issuecomment-144098333)
- long descriptions
- language maps (as supported by [JSON-LD language maps](http://www.w3.org/TR/json-ld/#dfn-language-map))
- [audible labels](https://github.com/valueflows/agent/issues/31#issue-109958938)
- ???

possible options:
- [`skos` labels](https://en.wikipedia.org/wiki/Simple_Knowledge_Organization_System#Labels_and_notations)
- `as` [`displayName`](http://www.w3.org/TR/activitystreams-vocabulary/#dfn-displayname), [`title`](http://www.w3.org/TR/activitystreams-vocabulary/#dfn-title), and [`summary`](http://www.w3.org/TR/activitystreams-vocabulary/#dfn-summary)
- `rdfs` [`label`](http://www.w3.org/TR/rdf-schema/#ch_label) and [`comment`](http://www.w3.org/TR/rdf-schema/#ch_comment)
- `dcterms` [`description` ](http://dublincore.org/documents/dcmi-terms/#terms-description)
- ???

elf-pavlik:
awesome summary! thanks @ahdinosaur :smile:

fosterlynn:
Another one: 

dcterms:description

ahdinosaur:
i propose we adopt the following [labels](http://www.w3.org/TR/2009/NOTE-skos-primer-20090818/#seclabel) and [documentation](http://www.w3.org/TR/2009/NOTE-skos-primer-20090818/#secdocumentation) properties for every domain:

- [`skos:prefLabel`](http://www.w3.org/TR/skos-reference#prefLabel)
- [`skos:altLabel`](http://www.w3.org/TR/skos-reference#altLabel)
- [`skos:hiddenLabel`](http://www.w3.org/TR/skos-reference#hiddenLabel)
- [`skos:note`](http://www.w3.org/TR/skos-reference/#note)

and for only the type objects:

[...]

PR #70, opened 2015-10-17
use skos label and note

ahdinosaur:
first pull request towards https://github.com/valueflows/valueflows/issues/34.

changes:

- remove special handling of `rdfs:label`
- add `label` which points to `skos:prefLabel`
- add `labelMap` which is language map for `label`
- add `note` which points to `skos:note`
- add `noteMap` which is language map for `note`
- remove `displayName` and `displayNameMap`
- `Relationship` is also a `skos:Concept`

### DIFF
--- a/context.jsonld
+++ b/context.jsonld
@@ -3,66 +3,61 @@
     "owl": "http://www.w3.org/2002/07/owl#",
     "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
     "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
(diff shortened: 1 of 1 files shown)


elf-pavlik:
:+1: thanks @ahdinosaur 

@fosterlynn if you also :+1: let's merge it 

we all should also take care to update all the .jsonld examples accordingly

*workflow note*: we don't need to create PR for updating .jsonld examples that we manage, except when we really want to get feedback from others on changes we made

fosterlynn:
merged, although I need to apply it in practice now to understand it completely!  (all good)

Issue #71, opened 2015-10-17
Valuing as Agent -> Resource relationship


elf-pavlik:
Following https://github.com/valueflows/valueflows/issues/69#issuecomment-148892583

I would propose that we stay very careful wherever we introduce anything related to some agent **valuing** some resource. If we take direct relationship

Agent - wants -> Resrouce

(eg. elf Pavlik wants a piece of cake)

We can reify it and than qualify it with anything related to valuing 

Agent <- subject - Want - object -> Resource

(eg. Want of elf Pavlik for a cake, as long as it uses vegan and fair trade ingredients)

So **valuing** related qualifications would refer to a Want (or an Offer) not to the Agent or to the Resource!

See also:
* #63 
* #62

bhaugen:
@elf-pavlik in the contexts we work in, assigning values to resources is a community effort. This is necessary for agreed-upon value equations and distribution of resources.

As such, value becomes an attribute of a resource, assigned by a community according to whatever rules they prefer.  It can also then be summed up at different stages of a value flow, assuming the community can arrive at a common unit of value. Which could be peanuts, doesn't need to be money.

Does this make sense to you? Or not?

elf-pavlik:
> in the contexts we work in, assigning values to resources is a community effort. This is necessary for agreed-upon value equations and distribution of resources.

Agent (Community) - values -> Resource

So we can reify 'values' edge and qualify it with peanuts or whatever we want to use. This way another agent (possibly also community) can feel free to value same resource in a different way.

> As such, value becomes an attribute of a resource, assigned by a community according to whatever rules they prefer. It can also then be summed up at different stages of a value flow, assuming the community can arrive at a common unit of value. Which could be peanuts, doesn't need to be money.

As I explained above, I see _assigning values directly to resources_ a **major misunderstanding** in how people commonly model economic states. Valuing comes from agents not resources themselves. One could try arguing 'resource has value' as - a hypothetical average (normal?) agent values this resource in such and such way. The only problem I see with that - in real world I never met a Person or a Group which would represent such 'average (normal?) agent'

Once we get to examples which include aspect of valuing, I feel confident that I will always find an agent, who has the actual valuing based relationship with a particular resource.

bhaugen:
I'm totally okay with modeling values in that way in the valueflows vocab. 

[...]

PR #149, opened 2016-08-19
Context agents

fosterlynn:
For discussion: this is a first draft to try to clarify context agents.  Please comment!  I'll do a picture next to show everything that can have a context, and add that to the PR - as part of that, I need to figure out how the new transfers might fit with context agent.


elf-pavlik:
Thank you @fosterlynn !

I think we can start with process having a context agent and clarify the implication of that. We should also clarify if single process can have 'context' relation to more than one agent.

elf-pavlik:
> Connecting value creation to a context agent creates the ability to do the accounting for a context.

I would see agents to heave 'valuing' relationships with resources #71 . Resources come as output of processes and can have traceable IPO history. All those processes took resources as inputs, again resources with which various agents have various 'valuing' relationships. Having clear IPO chains allows different 'valuing' of all the resources involved in that IPO history by different agents.

Describing a project as process which has sub-processes should allow to get exactly the same list of events and related resources and people as the list you get by using 'context agent' approach.

bhaugen:
One difference between a project as a context agent as a project as a process is that a context agent has agency. One of the features of a context agent that Lynn did not list is that the relationships between context agents are those of negotiation: that is, they make agreements with each other, they cannot command and control each other. Each context is self-governing, at least as much as possible. 

One context may only request resources from another context. Within a context, the resources may be issued and used at will, subject possibly to some other access rules.

That is the membrane that Lynn wrote about.

In NRP value equation disributions (which do not necessarily need to be adopted by VF, although they do come from a lot of experience), when a distribution passed over from one context to another, the distribution at that boundary was given in total to the new context to distribute according to their own value equation.

The exception to that rule was when both contexts used the same value equation, which would happen only if they inherited their value equations from the same parent context.

Likewise I think in planning, one context may not plan inside another context, but just pass on requests to the other context.

The history of contexts in my experience came from a previous supply chain software project where I was discussing planning problems with a friend who worked on airline "trip repair": rescheduling airplanes, crews and passengers during weather emergencies etc.

The airlines used contexts they called "planospheres", which was the sphere of one planner, which was usually one or more hubs in a hub-and-spoke system. One planner could not reschedule anything in another planner's sphere, they could only request changes or maybe just pass on a part of the overall problem.

fosterlynn:
>Describing a project as process which has sub-processes should allow to get exactly the same list of events and related resources and people as the list you get by using 'context agent' approach.

I don't see why it has to be either-or.  If you see your context agent as also one big process, then that is fine.  Then the context agent of your big process is the agent side of the picture, and they can relate in that way.

There are a lot of contexts that just aren't big processes, except as a force fit.  Like the time banks for example, just a big bundle of exchanges, by a group of people who have agreed on some governance for themselves and use specific tools to facilitate their exchanges.  On some higher level regional model, it might be useful, but operationally for the time bank, it seems out of place.

fosterlynn:
>One difference between a project as a context agent as a project as a process is that a context agent has agency. One of the features of a context agent that Lynn did not list is that the relationships between context agents are those of negotiation: that is, they make agreements with each other, they cannot command and control each other. Each context is self-governing, at least as much as possible. 

Yes, good reminder that context agents tend to have some level of self-governance.  I will add to the md file.

elf-pavlik:
> Likewise I think in planning, one context may not plan inside another context, but just pass on requests to the other context.

What do you mean by 'plan inside a context'? Do see each resource as 'belonging to one context'?

fosterlynn:
>What do you mean by 'plan inside a context'? 

I suppose @bhaugen could speak to this; but you can usually do planning inside a context fairly freely.  (Irrespective of ownership probably.)  If you want something from another context, you may need to coordinate, negotiate, etc.

>Do see each resource as 'belonging to one context'?

Good question.  In NRP, we don't have that.  But it is possible that they do in reality.  (NRP has an assumed context for the resources of the installation scope.)  Belonging to a context may not apply to all resources, but I'm pretty sure it does to many of them.

More thought needed here.....

elf-pavlik:
> I suppose @bhaugen could speak to this; but you can usually do planning inside a context fairly freely. (Irrespective of ownership probably.) If you want something from another context, you may need to coordinate, negotiate, etc.

If various agents own resources independently from any 'context agent', in practice the agent planning a process still needs to coordinate, negotiate, get allocations etc. I don't know what marking an agent as context of a process changes here. If resources don't belong to a context, what do you mean by "something" in "If you want something from another context"?

fosterlynn:
I added some commits - picture and governance note.  Picture is very draft!  (Got some syntax errors as well as needs to be broken out into much more detail, IMO.)

fosterlynn:
>If various agents own resources independently from any 'context agent', in practice the agent planning a process still needs to coordinate, negotiate, get allocations etc.

An example: In Sensorica, people will bring their personal equipment to the lab, where it is shared with others.  They can specify rules if they like.  The person is the owner.  But the equipment can be used, scheduled, etc. within the context of Sensorica, or possibly a project within Sensorica.

fosterlynn:
Another example:  In DHEN, people or small businesses own the herbs all the time.  But it is all managed by the network based on agreements by the people in the network.  All of the herbs are in the database, visible to everyone who wants to see them.  There are calls for people to go out and harvest particular herbs together.  Etc.

elf-pavlik:
> The person is the owner. But the equipment can be used, scheduled, etc. within the context of Sensorica, or possibly a project within Sensorica.

What type of relationship that resource has with 'Sensorica' agent?

bhaugen:
By the way, we are explaining how we did things in NRP, which does not need to apply to VF in every detail. In this case, it is just to give you several actual use cases for context agents. Although not all written up as use cases yet, they could be. For now, it is just to give you more of an idea of what we mean and what the concept is good for.

fosterlynn:
>If resources don't belong to a context, what do you mean by "something" in "If you want something from another context"?

Resource.  Or future resource.

fosterlynn:
>What type of relationship that resource has with 'Sensorica' agent?

Nothing in the software, but it is implied by the installation.  I'm thinking we do need a (optional) context agent for resources.  Owner or manager or whatever resource agent relationship doesn't seem to really cut it by itself.  Although Owner would work for private property oriented businesses.

Still thinking....

bhaugen:
> "If you want something from another context"?

This will happen in planning when your recipe wants something that you do not have on hand and your recipe cannot produce. We don't enforce this across contexts internal to a network, that is, internal to one instance of NRP, but we should. The reason we don't is that initially we thought that NRP instances would be federated, one parent context per instance. It didn't work out that way because Sensorica wanted many networks in the same instance. 'Twas easier for them. So we made it happen. That was Lynn's agent relationship model, which became the starting point for the same thing in VF.

elf-pavlik:
In diagram included in this PR
io-context-agent
*vf:InputOrOutput* gets used to **relate agent with a resource**, which in practice would look

screen shot 2016-08-19 at 4 37 04 pm

Since we discussed @fosterlynn proposal to use *vf:InputOrOutput* without attaching it to any process ...

I propose to focus in this PR just on proposed *vf:context* relation between *vf:Process* and *foaf:Agent* and leave this proposal of *vf:InputOrOutput* for separate discussion.

fosterlynn:
>vf:InputOrOutput gets used to relate agent with a resource, which in practice would look

OK.  I couldn't get my list of domain references to work. :(  And was just trying to show context agent relationships, although I see why you might want to show more since the InputOrOutput is an "edge".

>I propose to focus in this PR just on proposed vf:context relation between vf:Process and foaf:Agent and leave this proposal of vf:InputOrOutput for separate discussion.

In that case, do you have any objections to vf:Process having a relationship to foaf:Agent called vf:context?

elf-pavlik:
> OK. I couldn't get my list of domain references to work. :( And was just trying to show context agent relationships, although I see why you might want to show more since the InputOrOutput is an "edge".

Let's discuss those 'orphan' *vf:InputOrOutput* case in https://github.com/valueflows/process/issues/22

> In that case, do you have any objections to vf:Process having a relationship to foaf:Agent called vf:context?

I would like to focus on discussing implications of that process <-> agent relationship in this issue.

fosterlynn:
> We should also clarify if single process can have 'context' relation to more than one agent.

It cannot, as far as I know.

>I would like to focus on discussing implications of that process <-> agent relationship in this issue.

OK.  Since I have a bunch of other stuff in the PR, I'll just leave it for now though if that is OK with you.

elf-pavlik:
Q: does *vf:context* process <-> agent relation only affects the process itself or has also some implications on resources related to that process through *vf:InputOrOutput* ?

fosterlynn:
>Q: does vf:context process <-> agent relation only affects the process itself or has also some implications on resources related to that process through vf:InputOrOutput ?

We haven't had the occasion to deal with this in code, but I suspect the answer is yes.  I could be proven wrong, but tentatively I think that the resources that are input need to be in the context, and the output resource would be in the context based on the process.  The moving across boundaries, if needed, would need to happen in some sort of process/transfer event(s).  

The resource-agent relationships could be several different things I suspect, from ownership to some level of management or control for coordination purposes.  So it wouldn't be derivable from that necessarily.  But still, maybe there should be a connection.

I feel a need to explore the event sequences in some yaml, it feels slightly fuzzy to me in our current model.  And we haven't addressed any of the accounting type output use cases we defined earlier.  Those might be helpful too.

elf-pavlik:
> I could be proven wrong, but tentatively I think that the resources that are input need to be in the context, and the output resource would be in the context based on the process. 

Can resource has *vf:context* relationship to more than one agent? We haven't discuss yet *vf:context* as agent <-> resource relationship and it seems that this *vf:context* agent <-> process relationship relies on it.

Then we have *vf:context* as agent <-> relationship relationship (yes, we want to put relationships between agents in some context) and maybe *vf:context* agent <-> agent relationship?

I would like to avoid including not clearly defined terms like *vf:context* on *vf:underlyingResource* on which later people may want to 'piggyback' and use them for all kind of unpredictable purposes, instead of trying to define and standardize a clearly defined model.

stevebosserman:
@fosterlynn you wrote:

> There are a lot of contexts that just aren't big processes, except as a force fit. Like the time banks for example, just a big bundle of exchanges, by a group of people who have agreed on some governance for themselves and use specific tools to facilitate their exchanges. On some higher level regional model, it might be useful, but operationally for the time bank, it seems out of place.

I'm very interested in utilizing the timebanking transaction platform to capture time spent by members on tasks related to a project.  This assures documentation in the system of contributions made by members.  Then, if a project converts the value generated into money, those who participated can get a share of the gain based on time spent and other conditions as recorded in the governance agreement.  An example would be writing a grant proposal.  Oftentimes, many more proposals are submitted than are awarded.  With hours recorded, at least there's an acknowledgement for who did what and how much time was invested.  Then, if a proposal gets awarded and there's an allowance built-in that covers general operating expenses for the organization (also a member of the timebank) fronting the proposal, participants can be compensated based on time spent and the governance agreement.  

Does that approach "fit" into this discussion?

elf-pavlik:
@stevebosserman great use case! I think you could make a new issue documenting it and clarifying details like: If 5 different project teams - 4 people each (20 people all together) write 5 proposals and just 2 of those proposals result in income. All that income will get distributed among all the 20 people writing proposals plus all the other organization members who did some time contributions to any of that organization projects in 'that time'? Or each income from each proposal that yielded it gets distributed among the participants of the team who wrote it?

> @fosterlynn 
I could be proven wrong, but tentatively I think that the resources that are input need to be in the context, and the output resource would be in the context based on the process.

> @elf-pavlik
Can resource has vf:context relationship to more than one agent?

I don't think we can put all resources 'in context of some agent'. For example every person can make work contributions to any number processes no matter which agents manage them. Similar owner of a tool can offer its usage in any number processes no matter which agents manage them. Even owner of big pile of apples can allow their consumption in any number of processes no matter which agents manage them.

In similar way, a construction group from one village can manage construction process in another village which takes as inputs work, usage of tools, and material resources from independent people living in any number of villages around to repair a house owned by particular family which has no affiliation with that construction group. I don't see here a clear 'context agent' which has all the input resources, the process and the output resource in its context.

bhaugen:
>  I don't see here a clear 'context agent' which has all the input resources, the process and the output resource in its context.

That's not necessary. Phil's microscope is in the context of his university, which owns his lab and microscope. He is the custodian. Individuals often bring their own tools to a job, without the tools ever being part of the context of the job.  

Context agents are not necessary. Only if they exist in real life. Phil's university does, as does Sensorica, although Sensorica has no "legal" existence. They think of themselves as a context for projects and processes, and act like one.

fosterlynn:
>Does that approach "fit" into this discussion?

@stevebosserman  Not sure, but I agree with @elf-pavlik , great use case!  I think all of your carefully considered input on timebanks is extremely useful to VF.

>I don't think we can put all resources 'in context of some agent'.

Quite possibly not.  I wasn't thinking of the vf:Resources like work, just trying to think of resources that people would generally account for in a context, say for a traditional balance sheet.  Maybe all of that can handled by events and/or the agent-resource relationship, and we don't even need to think about it?

I have no desire to push resources needing a context agent.  If we run into it in a use case, we'll see then.  In the meantime, my attempts to think about your original question aren't being very useful without a concrete use case, so I hereby drop them entirely, and say resources don't have a context agent!  :)

But I do want to address how we think about context agents a bit.

>For example every person can make work contributions to any number processes no matter which agents manage them.
>Similar owner of a tool can offer its usage in any number processes no matter which agents manage them. Even owner of big pile of apples can allow their consumption in any number of processes no matter which agents manage them.
>In similar way, a construction group from one village can manage construction process in another village which takes as inputs work, usage of tools, and material resources from independent people living in any number of villages around to repair a house owned by particular family which has no affiliation with that construction group.

Managing a process does not make an agent the context agent.  I did work in the contexts of VF, Mikorizal, FairCoop, our small housing coop, our household, Madison MAN, and probably others, all in the last few days.  And I wasn't particularly managed in any of them.  But I always know what context I am working in, and who else is part of that context in some way (agent relationships).  And some of these agents may want to account for economic activity that occurs in their context.  And I think all of the above have agency.  It's just "we" agency instead of "me" agency.

Neither is ownership the same as being the context agent necessarily.  The tool or apples are put into a context if they will be used or consumed.  There might even be an explicit event that does that.  For example, the owner of the pile of apples might give/issue/transfer/whatever some apples to put them into a process that is occurring in another context.  

I think managership and ownership should be handled with agent relationships and agent-resource relationships.  Context agent should not imply either.

elf-pavlik:
>  I did work in the contexts of VF, Mikorizal, FairCoop, our small housing coop, our household, Madison MAN, and probably others, all in the last few days.

Did you **work** become an input to something and added up to some outputs that could become (or already did) inputs to something else?

fosterlynn:
>Did your work become an input to something and added up with other inputs to yield some outputs that could become (or already did) inputs to yet something else?

None of those contexts has a computerized system.  Well, the housing coop has a minimal exchange based system just so we can file taxes.  I'm trying to get across how I think of context agents, that is all.  If they had computerized systems, and anyone cared about keeping track, then yes it would be like that.

elf-pavlik:
should we just close this PR?

Issue #173, opened 2017-02-04
How to model Service


fosterlynn:
We have discussed this off and on in different contexts, but I can't find an issue dedicated to it, so I'm starting one.  This is triggered by the work done on exchange, here: https://github.com/valueflows/exchange/issues/32.  The transfer model (https://github.com/valueflows/exchange/issues/32#issuecomment-267834962) makes it a lot easier to model Service in a simpler way that also more corresponds to reality, at least when the transfer of the service needs to be recorded. 

This relates to both Process and Exchange, providing a service, and 'transferring' that service to one or more agents.  Treating a service like a material resource leads to clunkiness - you create some kind of "service resource", which then is transferred to another agent(s).  The process side is pretty clean, since services tend to involve at least a work input and often more than that.  But in fact once the process is complete the service has been transferred, like it or not.  Once someone's hair is cut, it is cut.

The exchange model we have been working on uses events rather than resources.  This gives freedom to for example exchange work for something else.  It also in similar manner should give the freedom to exchange a service without creating a somewhat mythical resource to exchange.

Proposal:  We create a new vf:action verb (type of event) for service outputs, called something like provideService, corresponding to vf:produce for material items, i.e. the output of the process that provides the service. The vf:affectedResource referenced by the event will just have information on the model and/or category, there will be no instantiated vf:Resource.

(Note this is separate from the question of does the person's hair or whatever go into and out of the process, like an automobile that is being repaired might do, which I'm not really addressing here.)

When a service will be exchanged for something else, the new exchange model works well. The output provideService event will create a commitment for the reciprocal event (like payment for the haircut).

The problem we ran into when discussing this (https://github.com/valueflows/exchange/issues/32#issue-193362583) is what do we do when a service is output of a process, and also input into another process, rather than being exchanged.  The example is when Elf Consulting Services provides a service of participating in a ValueFlows meeting.  Elf Consulting Services has a process where Pavlik does some work, uses a computer and internet, and outputs the service.  The ValueFLows process of conducting the meeting wants to input that service.  The output and the input are the same thing.  Proposal: We use the same event as output of one process and input to another process. This doesn't follow the basic value flow pattern of I > P > O > R > I > P > O.  So that seems unfortunate.  But it also seems like the best representation of reality.  Something like I > P > O/I > P > O.  And should be easy to track in a chain, even though it is an exception.

Are we close to finalizing the "service" question?  Or not yet?

I can write some yaml.

bhaugen:
[...]

PR #176, opened 2017-02-15
First draft UML for an actual release

fosterlynn:
Maybe 0.1? :)  

This is to request feedback on this UML model of what we have either decided or have discussed a lot or just seems like it can be here at this point.  There are some things that will look different in a RDF based model - vowl coming after feedback on this.  Like UML model requires properties to be part of a class, so some of these will look a bit different. 

I have not been totally diligent on looking for all the properties of some of the entities, will do some scanning of issues, but welcome anyone to notice missing ones.


fosterlynn:
When we have this cleaned up, I'd like to take it and create a vowl diagram, and create all the class and property documentation, as in this: https://github.com/valueflows/valueflows/blob/master/release-doc-in-process/README.md.  

Then I'd like to replace the valueflows/valueflows readme with this release doc.  Might need to make it shorter and just link to the detail.

What do you all think?

djodjoni:
I really like this initiative and I am trying to follow the discussions, however I have not been able to actively participate too much in the rather sophisticated discussions :) I have 2 main concerns on this PR:

1) I never liked UML I think it kills OO design rather than helping it. This case in particular makes the system quite incomprehensible for an average human being like myself, apart from UML being ugly. As you are not defining some industrial/bank software I think VOWL is much better choice.

2) there are many cases where the predicate is the class at the same time. this doesn't really make sense for me or am I missing something ?

Anyway you're all doing a great job and I admire the efforts you put in here 👍  :)

djodjoni:
I will fast answer myself for the first one: I am missing some thing -  they are different. 

Isn't it better to have for example hasAgreement rather than just agreement? or it is already not possible to change the ontology ?

fosterlynn:
@djodjoni, thanks for the feedback!

>I never liked UML I think it kills OO design rather than helping it. This case in particular makes the system quite incomprehensible for an average human being like myself, apart from UML being ugly.

Sorry about that, I use it because 1) I think a lot faster in UML, I am newbie to linked open data.  2) people I want to talk to about the vocab in the near future won't understand LOD  3) I think we need both for documentation, because what you understand depends on your background, and because although we focus on LOD, we want the vocab to work for LOD plus other methods of interoperability.  VOWL will be next, so hopefully that will help you and others.  But it will a bit painfully slow for me so probably won't be this week. :)

>or it is already not possible to change the ontology ?

Nope not at all, it is wide open to changes!  So your feedback is appreciated, feel free to jump in any time!  Well, some of it is not completely wide open because there has been agreement among the core - but there is nothing that is totally finalized.  Especially names.

djodjoni:
@fosterlynn great :) I was a little bit exaggerating with 'incomprehensible', however if it is helpful for someone it's great :)

>  VOWL will be next, so hopefully that will help you and others.

I can help there as I am preparing some basic draft ontology for food and food actors (which can take advantage of the one you created here) and I am using WebVOWL to generate the diagram.

as regards to the simple naming things shall I make a PR in the ontology's repos? It is not a big deal but IMO improves the clarity when you see it in a diagram or elsewhere.

fosterlynn:
@djodjoni now that you mention food, I remember where I saw your name: https://github.com/ouisharelabs/food-dashboard/issues/1#issuecomment-279256735.  Nice!

@elf-pavlik who does a lot of work on ValueFlows is also participating in that discussion.

>I can help there as I am preparing some basic draft ontology for food and food actors (which can take advantage of the one you created here) and I am using WebVOWL to generate the diagram.

Great!  Would it be useful to compare the vocabularies and see if we want to move them closer together?  ( @bhaugen did some software for food networks several years ago, using the REA ontology, which this is also based on.)

As to WebVOWL, I have used it some and need to get more comfortable with it.  (Actually not the tool, but more the syntax.)  I was thinking to wait until we clean up this, but we don't have to.  And all help is appreciated!  Should I give it a first shot, and get you to review it?  Or do you want to give it a try?  I will work on this more next week, at a conference tomorrow thru Monday.

>as regards to the simple naming things shall I make a PR in the ontology's repos?

You could do that.  Or you can just make suggestions in comments in this one if that is easier. Your choice. :)

I'll change the vf:agreement ones, but am thinking of what it should be.  (Pavlik made that general comment earlier too.)

djodjoni:
Yep I am participating the fooddashbord now more actively and yea more than a year ago @elf-pavlik  introduced us @cristianvasquez and myself to this as we had started to work on something similar in Belgium and this is when we created https://github.com/openthings-cc/FarmFood-Interoperability and started to test softwares and vocabs, but I didn't spend too much time after that apart from working with farmers and see how the whole process actually works. However, I'm back on track and hope to have initial proposal ready soon so we can discuss it in fooddashboard as there is not too much activity towards this direction also. 
As regards to getting closer the 2 vocal together the food only related part has not much in common however the whole process of offering, ordering, arranging pick/delivery, combining deliveries from different flows is naturally based on the work you're doing here.

As Regards to the ontologies I can just wrap them in one ttl in some branch and then we can paly around and visualize it smth like this http://visualdataweb.de/webvowl/#iri=https://raw.githubusercontent.com/valueflows/process/master/transport.ttl

Shall I make a PR to your branch?

fosterlynn:
>Shall I make a PR to your branch?

You just went past my github workflow knowledge.... :)  so do what seems cleanest to you, it will be fine!

fosterlynn:
I committed an updated diagram to the branch, with suggested updates and also added to Commitment and Intent, which weren't finished.  Please feel free to suggest removal of things you aren't comfortable with, I made some stuff up.

fosterlynn:
>As Regards to the ontologies I can just wrap them in one ttl in some branch and then we can paly around and visualize it smth like this http://visualdataweb.de/webvowl/#iri=https://raw.githubusercontent.com/valueflows/process/master/transport.ttl

Cool!!

fosterlynn:
Committed another update, added to EventTemplate, tried to make more readable.

djodjoni:
added a pull request to this branch, when it is merged it will appear here for merging to master along with the UML. However, as I explained there it is not yet finished so ill push some commits in the following days.

fosterlynn:
>added a pull request to this branch, when it is merged it will appear here for merging to master along with the UML. However, as I explained there it is not yet finished so ill push some commits in the following days.

Awesome!  This won't be merged to master for at least a week, I'm guessing.  And it can wait for yours too.

elf-pavlik:
Hi @djodjoni :wave: Thank you for PR to this PR :smile: 

@fosterlynn I suggest to merge this PR right away and continue iterating with small PRs which address small specific issue. I can think already of 2-3 changes, but I would prefer do it by making bite size PRs after we merge this one.

djodjoni:
@fosterlynn @elf-pavlik glad I could help :) 
I personally like iterating with small increments. What I will probably be able to finish tonight it is to complete the rdf to the level of the UML diagram that and vowl shows the same thing. I can do that PR anyway you prefer :)

elf-pavlik:
@djodjoni AFAIK @fosterlynn and @bhaugen focus on Mutual Aid Network meeting next 3-4 days so we won't merge it in those days. Please feel free to get the VOWL in sync with UML and when we start iterating we should also modify both to keep them in sync.

@fosterlynn could you please *Export as -> XML ...* from draw.io and add source of that UML to this PR?

djodjoni:
@elf-pavlik for keeping it in sync I can find/write some converter for the draw io xml format <=>rdf as I do with vowl json format

djodjoni:
the PR is done:
---
Added more classes and properties to match the UML. 
TODO:
- add DataProperties 
- clean up unused properties and classes
- revise properties namings
- revise domain/ranges od ObjectProperties as not too clean now
---

As I mentioned before I try to do some vocabulary for food and food actors and I try to relate it to VF. There I mix OWL and SKOS in order to allow easier development and reuse of different decentralized vocabs. 
you can check it here: https://cdn.rawgit.com/openthings-cc/openthings-food/master/ofd-core.ttl
I think here similar principle can be used.

pmackay:
@fosterlynn I notice the UML has a whole bunch of Template classes :) I'm still wondering if these arent very similar to Types? :) However some descriptions and examples would really help there (which takes more time to document of course...) similar to the content on https://github.com/valueflows/valueflows/tree/master/release-doc-in-process.

bhaugen:
@pmackay 
> I notice the UML has a whole bunch of Template classes :) I'm still wondering if these arent very similar to Types? :)

They used to be called Types, and in the REA ontology, they are called Types. Those are mostly the REA Types renamed as Templates.

fosterlynn:
@pmackay

>>    I notice the UML has a whole bunch of Template classes :) I'm still wondering if these arent very similar to Types? :)

>They used to be called Types, and in the REA ontology, they are called Types. Those are mostly the REA Types renamed as Templates.

I believe @elf-pavlik preferred to call them Templates; I tend to prefer Types, but I don't have a strong feeling about it and I am fine with compromising on Templates, as I think he felt more strongly about it.   (Correct me if I am wrong, Pavlik.)

djodjoni:
> I believe @elf-pavlik preferred to call them Templates; I tend to prefer Types, but I don't have a strong feeling about it and I am fine with compromising on Templates, as I think he felt more strongly about it. (Correct me if I am wrong, Pavlik.)

@elf-pavlik yea really why not using Type? I also don't mind any name but since there so much discussion around it, it could be easier for adoption :)

elf-pavlik:
Instances all the VF classes would not reference instances of those Templates as `rdf:type` / (`@type` in JSON-LD). Also I don't think we would define those Templates as instances of *owl:Class*. All those Templates stay optional and only get used when someone want to define Recipe. I find it very confusing to call them 'Types' (especially that currently those Templates come into play when creating Recipes). Also those Templates don't come organized in some common Taxonomy.

BTW one of my first follow up PRs will propose changing *vf:RelationshipTemplate* to *rdf:Property* since it supposed to reflect https://www.w3.org/TR/activitystreams-vocabulary/#connections

fosterlynn:
>I find it very confusing to call them 'Types' (especially that currently those Templates come into play when creating Recipes).

I think we should go with Template, unless someone has strong objections. 

>Also I don't think we would define those Templates as instances of owl:Class.

Why not?  Would you use rdfs:Class, or do you think they shouldn't be classes? @elf-pavlik 

>BTW one of my first follow up PRs will propose changing vf:RelationshipTemplate to rdf:Property since it supposed to reflect https://www.w3.org/TR/activitystreams-vocabulary/#connections

This came about in my struggle to see how to define both OO based and LOD based views.  It is quite possible that the relationship template won't be there in LOD.  Or possibly it is a subclass of rdf:Property.  Anyhow.... I can wait for your PR.  :blush:

bhaugen:
> All those Templates stay optional and only get used when someone want to define Recipe. 

They will also frequently be used in exchanges, processes, and inventories. And the goodrelations:Model came from the e-commerce domain.

> Also those Templates don't come organized in some common Taxonomy.

Not necessarily common taxonomies, but they do often appear in taxonomies of one kind or another, and those taxonomies get shared among trading and supply chain partners. That was particularly true in food networks, but also appears in substitute components or ingredients in manufacturing.

djodjoni:
> > Also I don't think we would define those Templates as instances of owl:Class.

> Why not? Would you use rdfs:Class, or do you think they shouldn't be classes? @elf-pavlik

I don't know if for the same reasons but in the other [issue](https://github.com/valueflows/valueflows/issues/179#issuecomment-281351735) I proposed something similar:

> For LOD specificities I would propose subclass the Templates and Categories from SKOS Concept as it provides nice way to build those structures and also compatible with all other SKOS thesauri.

The reason for me is that people can create SubClasses and Individuals in different vocabs for the same thing and then it becomes a mess and not really reusable. So I started to like to have a few base classes and the rest related Concepts, unless of course one needs too much restrictions in the ontology. 

> > Also those Templates don't come organized in some common Taxonomy.

> Not necessarily common taxonomies, but they do often appear in taxonomies of one kind or another, and those taxonomies get shared among trading and supply chain partners. That was particularly true in food networks, but also appears in substitute components or ingredients in manufacturing.

Absolutely my view as well :) I found that more than 2/3 of product templates can be shared in online food stores.

fosterlynn:
>The reason for me is that people can create SubClasses and Individuals in different vocabs for the same thing and then it becomes a mess and not really reusable. So I started to like to have a few base classes and the rest related Concepts, unless of course one needs too much restrictions in the ontology.

OK, I'm learning, will take a closer look at skos.

pmackay:
@fosterlynn about the diagram, I noticed there is not a Location class/box - is that intentional or just left out?

djodjoni:
a question here about Agent.
there is vf:Agent/Person/Group exactly the same and referencing foaf. shall we just use foaf without creating specific vf Classes ?

bhaugen:
> is vf:Agent/Person/Group exactly the same and referencing foaf. shall we just use foaf without creating specific vf Classes ?

Agent in VF is the same as REA:EconomicAgent, which is maybe a subset of foaf:Agent with some additional properties. Economic Agents are able to affect Economic Resources, make economic Commitments, etc. I don't know if that means we should have a specific vf:Agent or not, but VF needs Agents to do anything. They are not optional.

fosterlynn:
>@fosterlynn about the diagram, I noticed there is not a Location class/box - is that intentional or just left out?

I left it out for now because we don't know what it looks like at all.  This was meant to represent what we basically agree on, or have discussed enough to be close to agreeing on.  I'd like to get location in at some point.  There might be other things missing too....

If someone specifically needed location for a project, we could focus on it.

djodjoni:
It is a design choice and lets do more strict restrictions and reasoning. ex: 
having:
```
   ofd:deliversTo a owl:ObjectProperty ;
    rdfs:label "deliversTo" ;
    rdfs:domain vf:Agent ;
    rdfs:range foaf:Agent .
```

and

`xx:A ofd:deliversTo xx:B .`

you can infer that A is VF Agent, while B is FOAF

fosterlynn:
>there is vf:Agent/Person/Group exactly the same and referencing foaf. shall we just use foaf without creating specific vf Classes ?

I meant to just use foaf, but I'm interested in @elf-pavlik 's opinion, and will go with whatever he and the other LOD experts think.

The properties are somewhat different, but it seems like a shame to create yet another agent structure.

djodjoni:
> The properties are somewhat different, but it seems like a shame to create yet another agent structure.

Indeed. The 'issue' would be mostly infering and that we cannot nicely make some restrictions on the new and old properties otherwise we can add any properties on the domain and range foaf:Agent. IMO at this stage I would keep is simple and generic.

djodjoni:
I have another set of thoughts :)
Let's say we have a service - delivery service, where "delivery service" is the category right? now This does not really need Template (as your example with the room) fo we have 
Resource: 'delivery from A to B'
Category: 'delivery service'
still right ?

now this is done we have a fulfillment and we're all cool?

How about we have a delivery contract and this service happens every week for example.

- Are there going to be multiple individual resources ?
- Does it make sense to have a template then ?

bhaugen:
Delivery service might want a Process Template.

djodjoni:
ok thx that seems the better option.

fosterlynn:
>How about we have a delivery contract and this service happens every week for example.

    >Are there going to be multiple individual resources ?

I think so, one for each week.  Especially if you have inputs to each delivery service that you want to keep track of, creating value flows.

Another thought, taking you down the slippery slope :), we are in the midst of a discussion of if a service process will create a vf:Resource, or if just the output event is enough.  Or maybe the vocabulary could support it either way.  Here: https://github.com/valueflows/valueflows/issues/173

djodjoni:
> I think so, one for each week. Especially if you have inputs to each delivery service that you want to keep track of, creating value flows.

indeed, in fact it is a chain of services delivery->picking->delivery done by different actors however.

> we are in the midst of a discussion of if a service process will create a vf:Resource, or if just the output event is enough. Or maybe the vocabulary could support it either way. Here: #173

IMO it is both, I was thinking about it yesterday. a process/services can be delivery thus have as in/output the same resource, can be picking, cooking which produce another product which contains the inputs and there can be eating which produces delight and energy which may trigger another complete unrelated Resource.

Th whole flow can be fairly complicated and currently, it is quite nicely handled by the Event, which can reference to any old or new Resource(s). The resources themselves can relate using (contains).

However now I ma wondering why do we need Process? Isn't that a special case of Service Resource?

bhaugen:
> However now I ma wondering why do we need Process? Isn't that a special case of Service Resource?

Processes are for converting input resources to output resources. Like manufacturing or cooking or bike repair. You might have many inputs and more than one output. Then some of the outputs become inputs to other processes, and you get IPO chains or resource flow networks.

elf-pavlik:
I have impression that we go in this PR all over the place. I suggest merging it and following up with smaller more focused PRs, Preferably they will include examples in YAML (JSON-LD) with concrete examples to make conversations more concrete.

fosterlynn:
ok i merged my own PR :)

djodjoni:
@elf-pavlik  agreed I'm making issues for some of the questions

PR #189, opened 2017-03-02
eliminate InputOutput, add action verbs, other tweaks

fosterlynn:
I made some changes to the ttl, haven't regenerated the VOWL diagram or tested.  Please review, especially @elf-pavlik .

The fulfillment stuff is not completely fixed, we need to review all of that some time.  Probably other things to fix too.

(InputOutput is an old name for Event.)

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -98,10 +98,6 @@ vf:Transportation  a     owl:Class ;
         rdfs:subClassOf  vf:Process .
 
 
-vf:InputOrOutput  a      owl:Class ;
-        rdfs:label       "vf:InputOrOutput" .
-
-
 vf:Transformation  a     owl:Class ;
         rdfs:label       "vf:Transformation" ;
         rdfs:subClassOf  vf:Process .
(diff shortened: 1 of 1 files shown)


elf-pavlik:
I'll take a look at it tomorrow

> (InputOutput is an old name for Event.)

Not exactly, we used InputOrOutput to refer to all Intent, Commitment, Intent, as in this diagram

copy of process-oriented flow

I think we should still think about possibility of having a common super class for Intent, Commitment and Event.

elf-pavlik:
given this definition
```ttl
 vf:provider  a       owl:ObjectProperty ;
         rdfs:domain  vf:Event, vf:Intent, vf:Commitment ;
         rdfs:label   "provider" ;
         rdfs:range   vf:Agent .
```

RDFS reasoner will apply it on instance 
```ttl
:foo a vf:Intent ;
  vf:provider :Alice .
```

it will infer additional types from *rdf:domain* resulting in
```ttl
:foo a vf:Intent, vf:Event, vf:Commitment ;
  vf:provider :Alice .
```

we have few options

1. define some common superclass and use it as rdfs:domain
```ttl
vf:InputOrOutput a owl:Class .
vf:Intent a owl:Class ;
    rdfs:subClassOf vf:InputOrOutput .
vf:Commitment a owl:Class ;
    rdfs:subClassOf vf:InputOrOutput .
vf:Event a owl:Class ;
    rdfs:subClassOf vf:InputOrOutput .

vf:provider a owl:ObjectProperty ;
   rdfs:domain vf:InputOrOutput .
```

2. use *owl:unionOf* in rdfs:domain definition, similar to first option just uses anonymous class defined as union of other classes instead having them using *rdfs:subClassOf*
```ttl
vf:provider a owl:ObjectProperty ;
    rdfs:domain [
        a owl:Class ;
        owl:unionOf( vf:Intent, vf:Commitment, vf:Event )
    ] .
```

3. use *schema:domainIncludes* (and *schema:rangeIncludes*)
```ttl
 vf:provider  a       owl:ObjectProperty ;
         rdfs:domainIncludes  vf:Event, vf:Intent, vf:Commitment ;
         rdfs:label   "provider" ;
         rdfs:rangeIncludes   vf:Agent .
```

those 'relaxed' variant from *schema.org* don't entail anything in formal way and just act as 'hints' for documentation purposes. reasoners will not infer anything based on them.

I suggest to start with option 3 since it seems closed to current definitions (we just need to globally replace *rdfs:range* with *schema:rangeIncludes* and *rdfs:domain* with *schema:domainIncludes). We also currently don't give much focus to formal entailments and reasoning so using those 'relaxed' variants may keep us on a safe side.

fosterlynn:
>I suggest to start with option 3 since it seems closed to current definitions

Thanks @elf-pavlik ! I always forget that inference stuff.

djodjoni:
> We also currently don't give much focus to formal entailments and reasoning so using those 'relaxed' variants may keep us on a safe side.

Indeed and I also like 3) the most.

elf-pavlik:
I'll merge this one and start new PR which applies proposal 3)

PR #190, opened 2017-03-02
remove RelationshipTemplate

elf-pavlik:
this aligns the definitions with https://www.w3.org/TR/activitystreams-vocabulary/#dfn-relationship
we can eventually reuse AS2.0 terms once it becomes REC


fosterlynn:
Yes, I saw RelationshipTemplate being just on the UML, not on the LOD based stuff.  Thanks.

Issue #123, opened 2016-05-09
graph != tree


elf-pavlik:
Since we may start writing more concrete examples, possibly in YAML #122, I think we should keep in mind the graph structure of data and avoid seeing it as tree. For example @ahdinosaur wrote 
snippet:
```yaml
'@type': vf:Transportation
resource: ex:resource/those-delicious-crates-of-apples-i-got-while-dumpster-diving
dimension: 20
start: ex:locations/the-zig-zag
end: ex:locations/the-lowlands
process:
  '@type': vf:Process
  inputs:
    - '@type': vf:Consume
      resource: ex:resources/coop-petrol
      dimension: 1 L
    - '@type': vf:Use
      resource: ex:resource/coop-vehicle
      dimension: 1
```

we should consider it **equal** to

```yaml
'@type': vf:Process
inputs:
  - '@type': vf:Consume
    resource: ex:resources/coop-petrol
    dimension: 1 L
  - '@type': vf:Use
    resource: ex:resource/coop-vehicle
    dimension: 1
'@reverse':
  process:
    '@type': vf:Transportation
    resource: ex:resource/those-delicious-crates-of-apples-i-got-while-dumpster-diving
    dimension: 20
    start: ex:locations/the-zig-zag
    end: ex:locations/the-lowlands
```

Both after parsing will have identical serialization and when serialized using [RDF Dataset Normalization](http://json-ld.github.io/normalization/spec/) will produce identical documents with the same checksum (hash). If someone needs particular shape of JSON, than one can use [JSON-LD Framing](http://json-ld.org/spec/latest/json-ld-framing/) but in my opinion we shouldn't privilege any particular frame and always rely directly on the structure of represented graph.

PR #191, opened 2017-03-02
clearer I/O relation between Process and I/C/E

elf-pavlik:
this PR IMO simplifies Input/Output relationships between Processes and Intents/Commitments/Events

Instead of relying on interpretation of 'action verb' the which relates Intent/Commitment/Event to a Process itself captures the input/output aspect.

Since we consider defining service as Intent/Commitment/Event which can reference two processes
* one as *outputOf*
* other as *inputOf*
Approach below makes it possible, since in case where we need to relatet I/C/E to two processes we can't just rely on 'action verb'.

Not sure if we need distinct properties for Intent, Commitment and Event or enough to just have *inputOf* & *outputOf* and rely on the type of instance using them to make distinction between Intent, Commitment and Event expressing Inputs and Outputs.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -22,8 +22,8 @@ vf:ResourceTemplate  a           owl:Class ;
 vf:ProcessTemplate  a           owl:Class ;
         rdfs:label           "vf:ProcessTemplate" .
 
-vf:EventTemplate  a           owl:Class ;
-        rdfs:label           "vf:EventTemplate" .
+vf:IntentTemplate  a           owl:Class ;
+        rdfs:label           "vf:IntentTemplate" .
 
 vf:ExchangeTemplate  a           owl:Class ;
         rdfs:label           "vf:ExchangeTemplate" .
@@ -125,32 +125,22 @@ vf:Currency  a           owl:Class ;
 
 vf:action
         a                   owl:ObjectProperty ;
-        rdfs:domain         vf:EventTemplate , vf:Intent , vf:Commitment , vf:Event ;
+        rdfs:domain         vf:IntentTemplate , vf:Intent , vf:Commitment , vf:Event ;
         rdfs:range          vf:Action .
 
 vf:inputOf
         a                   owl:ObjectProperty ;
-        rdfs:domain         vf:EventTemplate , vf:Event ;
-        rdfs:range          vf:ProcessTemplate , vf:Process .
+        rdfs:domain         vf:Event ;
+        rdfs:range          vf:Process .
 
 vf:outputOf
         a                   owl:ObjectProperty ;
-        rdfs:domain         vf:EventTemplate , vf:Event ;
-        rdfs:range          vf:ProcessTemplate , vf:Process .
+        rdfs:domain         vf:Event ;
+        rdfs:range          vf:Process .
 
 vf:commitedInputOf
         a                   owl:ObjectProperty ;
         rdfs:domain         vf:Commitment ;
-        rdfs:range          vf:Process .
-
-vf:intendedOutputOf
-        a                   owl:ObjectProperty ;
-        rdfs:domain         vf:Intent ;
-        rdfs:range          vf:Process .
-
-vf:intendedInputOf
-        a                   owl:ObjectProperty ;
-        rdfs:domain         vf:Intent ;
         rdfs:range          vf:Process .
 
 vf:commitedOutputOf
@@ -158,10 +148,30 @@ vf:commitedOutputOf
         rdfs:domain         vf:Commitment ;
         rdfs:range          vf:Process .
 
-vf:eventTemplate
+vf:intendedInputOf
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:Intent ;
+        rdfs:range          vf:Process .
+
+vf:templatedOutputOf
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:IntentTemplate ;
+        rdfs:range          vf:ProcessTemplate .
+
+vf:templatedInputOf
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:IntentTemplate ;
+        rdfs:range          vf:ProcessTemplate .
+
+vf:intendedOutputOf
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:Intent ;
+        rdfs:range          vf:Process .
+
+vf:intentTemplate
         a                   owl:ObjectProperty ;
         rdfs:domain         vf:Intent , vf:Commitment ;
-        rdfs:range          vf:EventTemplate .
+        rdfs:range          vf:IntentTemplate .
 
 vf:processTemplate
         a                   owl:ObjectProperty ;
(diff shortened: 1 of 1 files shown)


release-doc-in-process/all_vf.TTL
```diff
@@ -128,14 +128,34 @@ vf:action
...
+vf:inputOf
         a                   owl:ObjectProperty ;
         rdfs:domain         vf:EventTemplate , vf:Event ;
         rdfs:range          vf:ProcessTemplate , vf:Process .
 
-vf:willOccurIn
+vf:outputOf
         a                   owl:ObjectProperty ;
-        rdfs:domain         vf:Intent , vf:Commitment ;
+        rdfs:domain         vf:EventTemplate , vf:Event ;
```
djodjoni: I am not sure EventTemplates really fits in here. I will explain

djodjoni:
I like in/output and I like distinct properties. 
However there are some things which I don't feel right when trying to think about concrete SW implementation. It is related to EventTemplates and how we do the planning.

So first there is Intent then there is Commitment(on both ends) to put and consume some Resources into/from a Process.

So the actual in/outputs of a Process are Resources as in the many examples. The Commitments are to provide those to the Processes. On the other hand, you have Events when input happens and when output happens, which is in the runtime/observation phase and IMO it is different from the Templates of those, meaning the Properties in/out are semantically different. I will explain.

Processes are pipes, what flows in/out are Resources. Events are triggered when something happens.

so when you connect pipes you have Source(Resources)->Process->Process->Sink(Resources).
So in the planning phase it is important to identify what input a Process accepts and what output it produces and then link the flow having the Source and Sink is the Commitment/Intent. So I would do:

vf:acceptsRes
        rdfs:domain         vf:ProcessTemplate  ;
        rdfs:range          vf:ResourceTemplate .

vf:producesRes
        rdfs:domain         vf:ProcessTemplate  ;
        rdfs:range          vf:ResourceTemplate .

so then it is easier to visualize and plan/make Intents and Commitments to provide those Resources. Currently this is linked but not that clear.

Then when we build the flow shape putting Intent/Commitment (Offer) as source and Intent/Commitment (Demand) and processes or intermediate offers/demands in between.
and these can be mapped with commited/intendedIn/OutputOf properties.

When we monitor in runtime we will have the events on both ends of a process so we need just Actual Instances:

vf:inputOf
         rdfs:domain        vf:Event ;
         rdfs:range          vf:Process .

vf:outputOf
        rdfs:domain        vf:Event ;
        rdfs:range          vf:Process .

in other words, EventTemplate doesn't seems to be needed as it is covered by Intent/Commitment.

Does this makes sense at all to you or I missed some point of what you have in mind.

release-doc-in-process/all_vf.TTL
elf-pavlik: Please notice that git gives here inaccurate diff, I replaced *vf:occursIn* with two properties *vf:inputOf* and *vf:outputOf*. so they have same domain & range.

djodjoni: yes i know, but i meant for that, please check the loong description :)

elf-pavlik:
Thanks @djodjoni to do things step by step, maybe I will just remove *vf:EventTemplate* from domains of *inputOf* and *outputOf* and add two extra properties.

```ttl
vf:templatedInputOf
        rdfs:domain         vf:EventTemplate ;
        rdfs:range          vf:ProcessTemplate .

vf:templatedOutputOf
        rdfs:domain         vf:EventTemplate ;
        rdfs:range vf:ProcessTemplate .
```

EventTemplates still needs a lot of thinking, it only comes to play when defining a Recipe. Also quantities they reference can get multiplied to 'scale' the Recipe.

fosterlynn:
I also like splitting into input and output. :+1: 

As to how to do it.... in lots of examples, we have a Process with container of Events (as an example), and I like an input container and an output container.

Your examples put the reference on the other end, from the event (say) to the process.

It's like the OO - relational mismatch.  I don't know which is best for this?

(Just replying to @elf-pavlik original PR, @djodjoni might be talking about this too, but I need to study that comment.)

fosterlynn:
>in other words, EventTemplate doesn't seems to be needed as it is covered by Intent/Commitment.

It is not needed as a reference from I/C/E.  But it is needed because it is used to automatically generate plans from a recipe, which comes before the planning.  (And if I'm not understanding your context, please feel free to ignore...)

djodjoni:
> But it is needed because it is used to automatically generate plans from a recipe, which comes before the planning. 

I don't really get that. How would you generated plans if you don't know the offers/demands ?

elf-pavlik:
@fosterlynn in JSON-LD context we can have aliases
```json
"hasInput": { "@reverse": "inputOf" },
"hasOutput": { "@reverse": "outputOf }
```
which can make it simpler to reference events from the process. We should consider all properties as bidirectional. Turtle has and advantage over JSON-LD that it allows write triples more directly since we serialize graph, with JSON we may have this graph - tree mismatch #123 

N3 has even more natural way to use properties in either direction without need to always start with subject `:process123 is vf:inputOf of :event234203` === `:event234203 vf:inputOf :process123`

fosterlynn:
>>>in other words, EventTemplate doesn't seems to be needed as it is covered by Intent/Commitment.

>> But it is needed because it is used to automatically generate plans from a recipe, which comes before the planning.

>I don't really get that. How would you generated plans if you don't know the offers/demands ?

I might not understand what you are saying, if so, sorry!  Are you saying we don't need EventTemplate (or whatever we end up calling it) at all?  That's what I understood.  

Yes, you also would need some kind of demand to plan for.  And of course recipes are optional, they just save a bunch of time.

bhaugen:
@djodjoni (and all)
> How would you generated plans if you don't know the offers/demands ?

See https://hillside.net/plop/plop97/Proceedings/haugen.pdf

Most demands are dependent demands, dependent on some independent demand. So if you want some pizza, that's the independent demand. That causes dependent demands for the dough, sauce, toppings, oven time, pizza maker time, etc. And the demand for dough causes dependent demands for flour, water, mixer time, etc etc.

The offers for pizza maker and oven time and flour and sauce and toppings, etc, get matched up in a dependent demand explosion, which generates a plan by following a recipe.  

I hope that is all clear. If not, read that article, and if it's still not clear, I'll walk through the code with anybody who wants to do that.

djodjoni:
@bhaugen  thx for the link! btw funnily this :
> Most demands are dependent demands, dependent on some independent demand. So if you want some pizza, that's the independent demand. That causes dependent demands for the dough, sauce, toppings, oven time, pizza maker time, etc. And the demand for dough causes dependent demands for flour, water, mixer time, etc etc.

The exact same concept (with different examples) I have heard from a presentation of the GoodRelations guy :)
this is clear now I was just confused by the name of the EventTemplate which is fixed :)

fosterlynn:
Good to merge from my perspective...  @djodjoni ?

bhaugen:
> The exact same concept (with different examples) I have heard from a presentation of the GoodRelations guy :) probably he read the paper :)

The GoodRelations guy has been deep into this stuff by himself. He probably read all the papers I read before I wrote mine. People get certified to prove they understand this shit.

djodjoni:
@fosterlynn yep for me is ok too :+1: 
@bhaugen indeed

PR #192, opened 2017-03-02
quantities used by Intent/Commitment/Event

elf-pavlik:
First commit proposes *vf:incrementQuantity* and *vf:decrementQuantity*. Using those distinct properties we don't need to rely on 'action verb' to know if resource referenced with *vf:affects* should get incremented or decremented.

Similar to #191 we need to figure out if we want to use same properties on Intent/Commitment/Event all call them differently *eg *vf:incrementedQuantity*, *vf:commitedIncrementQuantity*, *vf:intendedIncrementQuantity*.

Most likely we also need to do the same for *vf:affects* - reuse on Intent/Commitment/Event or define variants of this property for each one.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -239,6 +239,36 @@ vf:decrementQuantity
         rdfs:domain         vf:Event ;
         rdfs:range          qudt:QuantityValue .
 
+vf:commitedIncrementQuantity
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:Commitment ;
+        rdfs:range          qudt:QuantityValue .
+
+vf:commitedDecrementQuantity
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:Commitment ;
+        rdfs:range          qudt:QuantityValue .
+
+vf:intendedIncrementQuantity
+        a                   owl:ObjectProperty ;
+        rdfs:domain         vf:Intent ;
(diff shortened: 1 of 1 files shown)


fosterlynn:
Thinking.... one down side is that if you want to query events for example and put them in a tabular report, it would be nice to fine the same property name for the same thing in all events.

How about a separate property for resourceEffect in the event?  (Brainstorming, not proposing.)

And does this mean you don't want to have the property hierarchy we have for vf:Action verbs?  Or more just trying to get derived info where it is easier to access?

elf-pavlik:
> Thinking.... one down side is that if you want to query events for example and put them in a tabular report, it would be nice to fine the same property name for the same thing in all events.

Depends of action, quantity means something different, so I don't really see them as "the same thing". When it comes to something usage of car, one may want to include both time & distance so it would require two different quantities on the same event. Maybe for usage of accesspoint, one may want to include time & bandwidh, for usage of charging station time & KWh etc.

fosterlynn:
>Most likely we also need to do the same for vf:affects - reuse on Intent/Commitment/Event or define variants of this property for each one.

At this point, I like variants for I/C/E, like "committedQuantity" etc.

elf-pavlik:
> Thinking.... one down side is that if you want to query events for example and put them in a tabular report, it would be nice to fine the same property name for the same thing in all events.

Such query requirement can also get satisfied with a common super property, as explained in http://euclid-project.eu/modules/chapter2.html section: *Inferences based on sub-property relationships*

Issue #185, opened 2017-02-26
templates vs types


fosterlynn:
@elf-pavlik and I had a long chat in gitter the other day that I want to bring into focus here.  It involves whether the things we currently call XTemplate, in the top layer are "templates" or "types".  This was after the also extended discussion of template/type/model in the [this PR](https://github.com/valueflows/valueflows/pull/176) and in #179 .

In the PR, I suggested we call those things in the top layer Template, because I thought we had agreed that although they are "type objects", the name Type is confusing because of the use of "type" as in rdf:type for a LOD class/instance kind of relationship.

Then in the gitter chat I found out that Pavlik thought template != type, while I thought template == type. I think we are both representing things the vocabulary needs, we just need to figure out what to call them, and where they fit in the model, and whether they go in the "top" layer or as "plan" layer, which aren't part of the vocab strictly, but do give good conceptual structure I think.  I think it involves whether the Templates can be used with some things "filled in" already - for example agents and quantities in an exchange template, so they can be used like an agreement for things they want to commit to later. 

Thus, this issue to figure it out.  There are a bunch of subtleties involved here.

ahdinosaur:
can see why Type is overloaded, since it has many meanings across programming paradigms that might be confusing. can also see why Template doesn't exactly convey the correct meaning.

other potential words for the soup: abstract, blueprint, prototype, concept, scheme, formula?

bhaugen:
I'm doing a bunch of research to differentiate type, template, class, category, and taxonomy, but it's pretty squishy and will take me at least until to tomorrow to disambiguate.

But as a preliminary, I think we need to remember that our domain is economic relationships, and not programming languages (although the programming language usage is informative).

So, for example, I think type in economic relationships tells you what something is. As in exchanges, if somebody gives you a Grade A Organic Haralson Apple, you pretty much know what that is.

On the other hand, a Template as (for example) an element in a recipe is used to specify (or as a factor in computing) a lot of the attributes of an input or output. For example, the resource type (which, in the context of this conversation is a Type, not a Template), and the quantity (which is not meant to be used as-is, which is a little un-template-like, but extended by the quantity to be produced: for example, if 2 Apples specified in the Applesauce input recipe element, and 10 units of Applesauce are to be produced, then 20  Apple inputs are required).

I'll continue with the research. This is a preliminary opinion.

fosterlynn:
hi @ahdinosaur really good to see you pop in over here :heart:

[...]

PR #193, opened 2017-03-05
changed Templates to TaxonomyItems

elf-pavlik:
based on discussion in #185 

please pay attention to possibly confusing git diff, maybe i shouldn't have changed order of definition and edit it at the same time... you can always View version of the file in this PR without the diff.


elf-pavlik:
I added commit removing `vf:intentTemplate` and `vf:IntentTemplate`. We may or may not need to add something like that again once we dive into nuances of Recipes.

Issue #195, opened 2017-03-11
Event should be renamed (back) to EconomicEvent


bhaugen:
I know the argument that it's vf:Event and we can define it any way we want.  It will still cause confusion. I've seen this over and over again since I first started working REA in 1996.

I try to stay out of the naming arguments, but I think this one is important and went astray.

Everybody thinks of some other meaning of event that is not EconomicEvent: ooh, it's an event, can I attend? Or, something happened! (Of some unspecified nature.) Maybe it's a javascript event. Onclick. Can I set up an event handler?

In the committees where REA became a standard, we got very careful about always saying "Economic Event". Otherwise, people went crazily off into whatever meaning of event was on their mind.

Besides conceptual confusion, there's also the issue of divorcing ourselves from the REA community. I think it will become increasingly important to stay connected, as people try to figure out how create sustainable, cooperative, community economies. REA is the distillation of accounting and the transformation from accounting for corporations to accounting for economic networks. Accounting will be important for community economies: what resources do we have, what happened with them, how are they doing? What resources do we need? Who needs what? Who can provide what?

Most of the practitioners in the REA community are not yet thinking on that level, but some of them are starting to.

I wrote this [overview of a whole economy](https://cryptpad.fr/pad/#/1/edit/fdqFAy57bZaYLsQzr9awYQ/NyflJGu9+dISW4tlmSmykv+R) last year. Or look at [peerconomy](http://peerconomy.org/wiki/Main_Page) for more details.

pmackay:
I confess I had a hiccup recently where the naming of just "Event" caused confusion with my project....

bhaugen:
Event is one of the most overloaded words I know of. Resource is another.

fosterlynn:
Is there any reason to shy away from the term "economic"?  I don't see that it necessarily has overtones of any particular economic system.  Or any of the allergic reactions that people have to things like "accounting" (boring! or none of your business!) etc.  It's just a statement of the domain, i.e. not political, not social, not technical, etc.

fosterlynn:
The only place I think there isn't confusion is with Agent.  Even if we could technically think of some exceptions, I think it is ok.

elf-pavlik:
Please go ahead and make PR

PR #198, opened 2017-03-12
change name to EconomicEvent

fosterlynn:
Made that change just here, if accepted, I'll fix it everywhere it occurs.

(Also got rid of the text in this file, because it is now elsewhere.  Hoping to generate these eventually.)

### DIFF
--- a/docs/specification/classes/event.md
+++ b/docs/specification/classes/event.md
@@ -1,16 +1,8 @@
-### vf:Event
+### vf:EconomicEvent
 
 <table>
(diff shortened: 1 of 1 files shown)


fosterlynn:
Another note:  Since we haven't agreed on the existence of "transfer" yet, I left the process stuff in the definition for now.

bhaugen:
Lynn says I am not allowed to merge this. Phooey!

fosterlynn:
Ref. some discussion here: #195 - but the participation of @elf-pavlik is required!  :)  And of course anyone else who has an opinion not yet expressed is very welcome!

PR #201, opened 2017-03-14
Working on ttl file, moved around text

fosterlynn:
No mergy please!!  I'm doing this PR so people can look at my humble addition to the ttl file.  I added comments from the other doc; and added a bunch of properties (the ones that are datatype properties or reference non vf stuff).  Moved all of the text that wasn't a rdfs:comment into more general files, although it is not really integrated yet, just parked.

This ttl file won't load in webvowl; but it does seem to validate.  So I'm not sure what is wrong with it, but I'm sure something.  I also would like a review, in particular I'm not sure how to do the ranges of the datatype-y properties.  But also in general.

And I have some temporary comments when we have more than one non-hierarchical class as domain.  I fixed a few of those, but more to be figured out.

And I'm sure I'm still missing some properties, I just made sure all the ones from the existing doc got in there.


docs/introduction/resources.md
```diff
@@ -22,6 +22,15 @@ is a model. Your library may have two copies that you can check out. Those are r
 
 *Names TBD.
```
pmackay: Is this a unclosed italic? Should it have a `*` at the end as well? Everything after is appearing as italics :)

docs/introduction/resources.md
```diff
@@ -22,6 +22,15 @@ is a model. Your library may have two copies that you can check out. Those are r
...
 
+Resource categories are a Taxonomy. That means they can be defined very broadly and generally and maybe vaguely, or they can be defined very narrowly, but fit into broader categories. 
+
+So, for example, you may want an apple. Or you may want a green apple. Or you may want a Granny Smith apple.
+Or you may want a Granny Smith apple from your neighboring farm.
```
pmackay: Is this last one a valid example? I dont think the location is relevant for the resource category.

fosterlynn:
@pmackay fixed, thanks!

fosterlynn:
Made another commit.  Removed Intent from the ttl and uml, I think it will be a bit before we have that fleshed out.  And did a few other misc. things.

I'd like to work towards merging this, will wait a couple days for any further review / objections.  And won't merge until I can get a vowl diagram to work.

And may get the generated doc for classes and properties in there (from ttl file); if not, will do right after, so you can ignore all the class and property definition files, the ttl file is the important one.

docs/introduction/core.md
```diff
@@ -1,17 +1,37 @@
...
+* vf:use - for example a tool used in process, after the process, the tool still exists
+* vf:consume - for example an ingredient composed into the output, after the process the ingredient is gone
+* vf:cite - for example a design file, neither used nor consumed, the file remains available at all times
+* vf:work - labor power towards a process
+* vf:accept - in processes like repair or maintentance, same resource will appear in output with *vf:improve* verb
+* vf:load -  transported resource enters the process
+
+###### Output
+
+* vf:create - new resource created in that process
```
elf-pavlik: I think we used *vf:produce* here to match *vf:consume*

fosterlynn:
@elf-pavlik thanks, fixed "produce"

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+        rdfs:label      "image" .
+
+vf:url a owl:DatatypeProperty ;
+        rdfs:label      "url" .
+
+vf:name a owl:DatatypeProperty ;
+        rdfs:label      "name" ;
+        rdfs:comment    "An informal or formal textual identifier for an object. Does not imply uniqueness." .
+
+vf:note a owl:ObjectProperty ;
```
djodjoni: vf:note is declared twice causing vowl json converter to fail. One needs to be removed.

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+
+vf:url a owl:DatatypeProperty ;
+        rdfs:label      "url" ;
+        rdfs:domain     vf:Agent .
+
+vf:name a owl:DatatypeProperty ;
+        rdfs:label      "name" ;
+        rdfs:comment    "An informal or formal textual identifier for an object. Does not imply uniqueness." .
+
+vf:note a owl:DatatypeProperty ;
```
djodjoni: this is probably the one we need to remove, unless we want the range to be a datavalue and vf:note is a subproperty of skos:note. actually do we need vf:note ?

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+        rdfs:label      "note" ;
+        rdfs:range      skos:note ;
+        rdfs:comment    "A textual comment or description." .
+
+vf:trackingIdentifier a owl:DatatypeProperty ;
```
djodjoni: DatatypeProperty could limit future uses, as ID could be represented by an individual resource. something like skos labels  had suffered from and skos-xl emerged to solve this.

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+vf:duration a owl:ObjectProperty ;
+        rdfs:label     "duration" ;
+        rdfs:domain     vf:EconomicEvent ;
+        rdfs:comment   "The observed duration of the economic event." .
+
+vf:image a owl:DatatypeProperty ;
+        rdfs:label      "image" ;
+        rdfs:domain     vf:Agent .
+
+vf:url a owl:DatatypeProperty ;
```
djodjoni: use foaf?

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+        rdfs:label      "url" ;
+        rdfs:domain     vf:Agent .
+
+vf:name a owl:DatatypeProperty ;
```
djodjoni: use foaf?

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+        rdfs:label     "start" ;
+        rdfs:domain     vf:EconomicEvent ;
+        rdfs:comment   "The starting date and optionally time of the economic event." .
+
+vf:duration a owl:ObjectProperty ;
+        rdfs:label     "duration" ;
+        rdfs:domain     vf:EconomicEvent ;
+        rdfs:comment   "The observed duration of the economic event." .
+
+vf:image a owl:DatatypeProperty ;
```
djodjoni: use foaf?

release-doc-in-process/all_vf.TTL
```diff
@@ -6,255 +6,854 @@
...
+        rdfs:range      qudt:QuantityValue ;
+        rdfs:comment    "The quantity and unit of the event. This is the quantity that could be used to increment or decrement a resource, depending on the type of resource and type of event." .
+
+vf:currentLocation a owl:ObjectProperty ;
+        rdfs:label   "current location" ;
+        rdfs:domain  vf:EconomicResource ;
+        rdfs:range   vf:Location ;
+        rdfs:comment        "The current place an economic resource is located.  Could be at any level of granularity, from a town to an address to a warehouse location.  Should be mappable." .
+
+vf:primaryLocation a owl:ObjectProperty ;
```
djodjoni: foaf or shema org?

djodjoni:
@fosterlynn  I can add a travis/codeship script and a jar to automate the creation of html, rdf formats and vowl json, where failure to generate those will be noted when a PR is made so we can make sure there is clean convertible ontology. Also some MDs/gitbook pages might be generated depending on the comment/description in the ttl. 
I can make a separate PR for that actually

release-doc-in-process/all_vf.TTL
fosterlynn: seems to be ok now; must have fixed earlier, possibly by accident :)

fosterlynn:
>+vf:trackingIdentifier a owl:DatatypeProperty ;

>DatatypeProperty could limit future uses, as ID could be represented by an individual resource. something like skos labels had suffered from and skos-xl emerged to solve this.

This is interesting, I'm not sure how to think about it.  In that case should we use owl:Thing?

fosterlynn:
>@fosterlynn I can add a travis/codeship script and a jar to automate the creation of html, rdf formats and vowl json, where failure to generate those will be noted when a PR is made so we can make sure there is clean convertible ontology.

Not a bad idea.  Although right now webvowl generator has been down all morning, as far as I can tell.  Do you think it would be too limiting?  Especially if people are working in branches?  I don't feel qualified to judge this.

djodjoni:
> This is interesting, I'm not sure how to think about it. In that case should we use owl:Thing?

Rather ObjectProperty

> Although right now webvowl generator has been down all morning, as far as I can tell. Do you think it would be too limiting?

I use the java library the web tool is using so i test offline and it also spits out nice readable errors :)
This in the java ugly thing i mentioned before :)

About branches that doesn't matter. The generated files might be pushed/uploaded after merging

fosterlynn:
>>This is interesting, I'm not sure how to think about it. In that case should we use owl:Thing?

>Rather ObjectProperty

But for our known use cases, it is a string - a lot number or serial number....

>The generated files might be pushed/uploaded after merging

Maybe that is better.  I set it up to generate each time.  But I don't know how that will behave inside of gitbook.... haven't tested that way.

elf-pavlik:
I think we can merge this one and continue in more atomic PRs that will follow up

fosterlynn:
>I think we can merge this one and continue in more atomic PRs that will follow up

Yeah, I think that will be easier... I'll pull the specific review comments from @djodjoni for the foaf issue (those are the ones not dealt with yet).

djodjoni:
:+1:  del the branch ? :)

PR #205, opened 2017-03-22
added all_vf.json. a json file created for webvowl by owl2vowl

bshambaugh:
Processed all_vf.ttl with https://github.com/bshambaugh/OWL2VOWL/tree/master/target/owl2vowl.jar . Fed the resulting file, all_vf.json into https://github.com/VisualDataWeb/WebVOWL  via Ontology>Custom Ontology > {"upload from file"} > Upload .


elf-pavlik:
Thanks @bshambaugh but I don't think we need this file committed here, what do you think @fosterlynn ?

fosterlynn:
No probably not, I was using an older beta version of webvowl, causing all of my problems.  @bshambaugh I owe you!!

Issue #52, opened 2015-10-09
owl:inverseOf


elf-pavlik:
I will use this issue to explain various nuances of owl:inverseOf
- http://richard.cyganiak.de/blog/2006/06/an-rdf-design-pattern-inverse-property-labels/
- `https://github.com/mnot/I-D/issues/39#issuecomment-89789736`

In examples which I added I use only single instance of rdf:Property - _vf:member_
- https://github.com/valueflows/agent/blob/master/examples/hackers4peace.jsonld#L13-L15
- https://github.com/valueflows/agent/blob/master/examples/elf-pavlik.jsonld#L14-L17

As I understand @ahdinosaur currently uses in holodex two different properties _hdx:hasMember_ and _hdx:isMemberOf_. To formalize that they have **exactly the same semantics**, but only reverse position of _rdf:subject_ and _rdf:object_ in _rdf:Statement_, one can use _owl:inverseOf_

I see need to explore those two ways of approaching the issue. As I see it now it may not make major difference when one traverses the graph, but it may complicate querying it by matching graph patterns!

With defining only one property on vocabulary level, both individual and group will have a single way to describe member relations

![member](https://docs.google.com/drawings/d/1pi6iy2vPXlEq_Y05ANOkf8YxeOD_lLbYFVfk3Z48oJM/pub?w=1237&h=567)

With defining both directions on vocabulary level, we end up in situation where individual and group can describe their relation in 3 different ways!

 ![member-inverse-a](https://docs.google.com/drawings/d/1b1_xeJwwYFfO3f-K76wvRqqwI9v-aJ6hh7eO-cibflc/pub?w=1237&h=567)

![member-inverse-b](https://docs.google.com/drawings/d/1UBCbgFKDbs_GSFZXBKQvOPprF6yzx7L1mjPCAIM-Sq4/pub?w=1237&h=567)

![member-inverse-c](https://docs.google.com/drawings/d/1ana7ep9uXh0zb2AUn1MPXDZ8hK7zuzXWNV8pdZ7n4ZI/pub?w=1237&h=567)

1) Please notice that last diagram (using only _hasMember_) differs from the very first one (_member_) only by naming of the property/predicate.
2) While from given individual perspective of given group perspective it may make more sense to use _hasMember_ or _isMemberOf_ in statements. When we describe something from 3rd person perspective, we can choose directions very arbitrary.

[...]

Issue #182, opened 2017-02-24
vf:Agent or foaf:Agent


djodjoni:
shall we have a custom vf:Agent or shall we just reference to the well know foaf:Agent and its subclasses

pmackay:
Pragmatic reuse seems preferable IMHO, so can VF use other vocab's classes, properties where possible? I see the ESS Global spec did this somewhat, and I really like how the Popolo Project [approached reuse](http://www.popoloproject.com/specs/#standard-reuse).

fosterlynn:
Yeah, if we can do "pragmatic reuse" that is great.  Good Relations did this too.  And yes, the Popolo Project documention is very good.  I'll have to look at their semantics, to see how they formally do mapping or reuse.

@djodjoni what is your opinion?  (implied but not stated :blush:   )

djodjoni:
Implied indeed and even if a statement here could be redundant :) I am also for "pragmatic reuse" this what I try to apply for the food vocab i'm doing :)

fosterlynn:
>Implied indeed and even if a statement here could be redundant :) I am also for "pragmatic reuse" this what I try to apply for the food vocab i'm doing :)

thanks just making sure....  :)

fosterlynn:
OK, we decided (I think) to just use foaf straight out for agents, right?  As opposed to referencing other vocabs as sameAs?  I didn't do this in the current round of doc for gitbook, I can change it.

Although I was noticing yesterday that foaf:Group is not what we have really.  To them it is any grouping or categorization of people, like "all the monarchs of England", not necessarily something that people have actually organized.  Maybe vf:Group is a subclass of foaf:Group?  This is a bit true for foaf:Agent, which is broader than economic agents, but that seems more like close enough somehow?

Now @djodjoni raises the same question for common properties like url, name, image.

Some of these may only apply to one of our classes, seems good to have that domain defined so it shows up there.... I don't know if we can do that if it is foaf?  The UML can always show everything, but that is really secondary documentation.

Please give your opinions, LOD people, I am traveling in the dark here.

pmackay:
>OK, we decided (I think) to just use foaf straight out for agents, right? 

+1

Could redefine Group if it has tighter semantics.

>Some of these may only apply to one of our classes, seems good to have that domain defined so it shows up there.... I don't know if we can do that if it is foaf?

Not sure I quite understand this, can you expand a bit?

djodjoni:
[...]

PR #206, opened 2017-03-22
Trying to clarify Agent

fosterlynn:
Please don't merge without some discussion, and also fixing of the .ttl file.  

This PR is to see if we can agree on the Agent portion of the model for release 0.1.  I tried to add text to explain how "we" (of differing views) see it.   I tried to honor the most recent discussion, as well as the very long discussions we had early in the VF project.  I tried to be clear, but of course welcome feedback and clarifications!

*Excerpt*
We also want to acknowledge that some people prefer to think of themselves as independent and decentralized agents who interact freely in different places in the economy as individuals, and some people think of themselves more as members of different groups and networks and communities and interact more in the context of those groups and networks and communities.  Many experiments are going on as people strive towards a better economy.  We want to support all these experiments, so want to support both of these ways of thinking and organizing ourselves.  The agent vocabulary is very flexible, and will support these as well as current conventional structures.

So, if people want to form a group that has agency as a group, fine.  If people want to consider that their group does not have agency as a group, also fine.  Not all groups, and especially not all networks, will be vf:Agents. That depends on the agreement of the people in the group.  Note that within the vocabulary, network formations will appear, as agents have economic interactions with each other in the world.  This does not mean that the network is necessarily a vf:Agent.
*End excerpt*

I removed Elio's network of networks picture from the readme, to back off the emphasis and because it has specifics I think we don't want to include.  Left Bob's in the Flows of Resources page so the concept is represented.

I also changed to foaf.  

And I included foaf:Agent, foaf:Group, foaf:Organization, foaf:Person.  I hope that we can agree that many people see this as useful and worth including for that reason.  I'm also OK with vf:GroupAgent, but think it is less important to be crystal clear than to re-use fairly standard terms - at least in this case. :)

I added explicit verbs for the main behaviors of agent relationships - member of, part of, peer of.  I called these AgentRelationshipVerb, and would like a better name!  I added them to the .ttl file for discussion, but haven't gotten the inverses in yet, see comments in https://github.com/valueflows/valueflows/issues/52.  Want to do that, but figured I'd get this out there for feedback before spending more time.  Probably need to fix something too, doesn't load in webvowl, and I'm using the right version now...  :(

I changed the name from vf:Relationship to vf:AgentRelationship, expecting we will have other kinds of relationships that will look pretty different, and to make the name very clear to minimize confusion.

OK!  Let me have it! :)


elf-pavlik:
Thank you @fosterlynn 

I would really prefer to start with less and add more only when we clearly need it. Also besides `foaf:Organization` we can reuse `org:Organization` which has status of W3C REC https://www.w3.org/TR/vocab-org/ BTW `org:Organization owl:equivalentClass foaf:Organization`.

Currently I vocab only requires distinction between `foaf:Person` and Non-Person where `org:Organization` seems perfectly fit. I'll add more comments inline to changed files.

fosterlynn:
I'm fine with org:Organization instead of foaf:Organization.

fosterlynn:
@elf-pavlik @ahdinosaur @djodjoni @gcassel I requested a review from you all.  Mostly I'm interested in the text and Agent subclass piece, but whatever else you have time for is great too.  Pavlik and I are going back and forth on it, it would be good to have some other opinions.  So if you have time, great; if not, no worries.

fosterlynn:
Re subclasses of foaf:Agent, @elf-pavlik and I just had a long discussion which brought no conclusions.  But the following criteria came up, so I'm summarizing them here, in case it helps us focus how to think about it.  These are of course not mutually exclusive, and please add if more are useful.

* for VF it matters what code needs to handle: thus far we have only identified specifically that we need to handle person and non-person, because only Persons can do vf:work (a possibly solution: foaf:Person and vf:NonPersonAgent)
* for VF it matters what people including developers and end users understand about the domain, what are the conceptual definitions of the classes that they will fit their groups and organizations into
* it matters that we do our best to fit in with existing standards, are good citizens of the standards world

docs/introduction/agents.md
```diff
@@ -1,24 +1,25 @@
-# Agent
+# Agents
 
-The agent vocabulary describes networks of people and groups, using agents and their relationships.
+The agent vocabulary describes networks of people, groups, and organizations, constructed using a simple but powerful model of agents and their relationships.
```
gcassel: This isn't very important, but I'd prefer:
>The agent vocabulary describes networks of people, organizations and networks

Reasons:
1.  To me, "group" is an extremely generic term including all organizations & networks, and isn't needed here.
2.  I perceive "organizations" to mean "structured groups" or "formally structured groups", including all "agentified groups" or "group agents" as well as many networks.   
3.  I perceive "networks" to broadly mean all sets of persistently interactive people (who can send and/or receive info, energy or matter to each other.)  Networks includes many huge groups of peers with no internal group structure.  To me, this seems to be both properly generic *and* distinctly contrasted with--although sometimes overlapping-- "organization".

Some people might dislike the intentional fractal redundancy created by explicitly referring to "networks of networks".  I personally think it's appropriate here.

docs/introduction/agents.md
```diff
@@ -1,24 +1,25 @@
...
+In ValueFlows, we are talking about economic agents, agents who can create or exchange value, and make agreements with each other - who have economic agency.  But we want to re-use existing vocabularies for commonly defined things, so we have elected to use those as much as possible here, even though they tend to be more broadly defined.
 
-#### Relationships
+We also want to acknowledge that some people prefer to think of themselves as independent and decentralized agents who interact in different places in the economy as individuals, and some people think of themselves more as members of different groups and networks and communities and interact more in the context of those groups and networks and communities.  Many experiments are going on as people strive towards another economy.  We want to support all these experiments, so want to support both of these ways of thinking and organizing ourselves.  The agent vocabulary is very flexible, and will support these as well as current conventional structures.
```
gcassel: This is "Relationships" section is a highly valuable addition!

fosterlynn:
I edited my last comment to try to make it more clear.

bhaugen:
"Nonperson" has some ugly connotations:
* https://en.wikipedia.org/wiki/Nonperson
* http://www.thefreedictionary.com/nonperson

fosterlynn:
@gcassel thanks!

>The agent vocabulary describes networks of people, organizations and networks

Here is an example of what I think is a group agent but not an organization and not a network, I'd be interested in your thoughts. :)

> what about people occupying a square, say Zuccotti Park in NYC, say a bunch of Occupy Wall Street people
>they ebb and flow, they don't have members, etc. - but they "use" a space resource for example
>they are not an org:Organization but they have economic agency, imo

fosterlynn:
added @pmackay as reviewer - if you have time, it would be helpful, if not, no worries

elf-pavlik:
> what about people occupying a square, say Zuccotti Park in NYC, say a bunch of Occupy Wall Street people
they ebb and flow, they don't have members, etc. - but they "use" a space resource for example
they are not an org:Organization but they have economic agency, imo

https://www.w3.org/TR/vocab-org/#org:Organization
> Represents a collection of people organized together into a community or other social, commercial or political structure. The group has some common purpose or reason for existence which goes beyond the set of people belonging to it and can act as an Agent. 

I think it fits the description of `org:Organization` so I don't understand why you suggest "they are not an org:Organization but they have economic agency"
I also have impression that you think of Organization more as https://www.w3.org/TR/vocab-org/#org:FormalOrganization

>* for VF it matters what code needs to handle: thus far we have only identified specifically that we need to handle person and non-person, because only Persons can do vf:work
>* for VF it matters what people including developers and end users understand about the domain, what are the conceptual definitions of the classes that they will fit their non-person agents into
> * it matters that we do our best to fit in with existing standards, are good citizens of the standards world

Anyone can define any more specific variants of both `foaf:Person` and `org:Organization` but for VF only that distinction matters based on the first point about *vf:work*. This makes easy decision between Person or non-person - Organization. Domain specific cases like *Single Parent* or *Refugee* would stay from VF perspective *foaf:Person*, cases like *Jolly Circle*, *Foursome*, *Fight Club* etc. can all stay from VF perspective an `org:Organization`. We still may need to take some time if we want to use *rdfs:subClassOf* do define more specialized versions of rely on something like *vf:category* which can reference items from taxonomies.

elf-pavlik:
http://visualdataweb.de/webvowl/
page-shot-2017-3-23 webvowl

FOAF defines *foaf:Person* as **disjoint** with *foaf:Organization* (and doesn't do the same for *foaf:Group*). We can also see this **disjoint** noted on this diagram

vocab-org

We actually need such **disjoint** classes definition for requirement related to *vf:work* so *foaf:Group* doesn't help in any way here and just creates confusing impression that one needs to choose between *org:Organization* and *foaf:Group*. And because lack of **disjoint** definition we can have valid instances with `"@type": ["foaf:Person", "foaf:Group"]` while `"@type": ["foaf:Person", "org:Organization"]` would contradict the **disjoint** definitions of those two classes.

fosterlynn:
Thanks @elf-pavlik 

>I think it fits the description of org:Organization so I don't understand why you suggest "they are not an org:Organization but they have economic agency"

I am mostly responding to the hierarchies in the existing standards that we are using, which says to me that they think there are Groups that aren't Organizations.  And which I'd like to reuse and respect, even though their models are not extremely rigorous.  As well as to people I have worked with that just don't like "organization" and are careful to define their networks in another way.  (I'm not one of those, btw, I like organization just fine.)

foaf has:
-foaf:Agent
--foaf:Person
--foaf:Organization
--foaf:Group

org has:
-foaf:Agent
--org:Organization
--foaf:Group
--foaf:Person

also, for completeness, schema has:
--schema:Person
--schema:Organization

gcassel:
>what about people occupying a square, say Zuccotti Park in NYC, say a bunch of Occupy Wall Street people

Those groups sometimes include identifiable organizations.  Often, however, many of the people present aren't part of those organizations.

The people in such huge gatherings have agency, but I don't think that they form an identifiably united agent.     

There are times that the whole gathering may seem to be swept over by an idea or energy, and act in a highly synchronized (and possibly mob-like) way.   But I don't think that makes them "an agent", any more than a group of water molecules become an agent when they're part of a wave which crashes on the shore.   -- A wave is an identifiable pattern of energy, distinct from the medium(s) it passes through.

elf-pavlik:
from gitter: 

btw RDFS reasoner seeing
```ttl
:Coop a org:Organization ;
   org:hasMember :Bob .
```
will infer
```ttl
:Coop a org:Organization, foaf:Group ;
   org:hasMember :Bob .
```
so as soon as organization org:hasMember it automatically also becomes foaf:Group
clue: rdf:domain of foaf:member.

IMO making people 'decide' between org:Organization and foaf:Group seems like wasting their time and just confusing them, since organization with even one member already becomes a foaf:Group but the (foaf/org):Organization makes it disjoint with foaf:Person by its definiton

fosterlynn:
>so as soon as organization org:hasMember it automatically also becomes foaf:Group

But we have no intention of using org:hasMember, so not an issue?

bhaugen:
>  preferably let ongoing software implementations to drive the requirements.

https://github.com/valueflows/vocabulator

elf-pavlik:
@bhaugen great to see you got started on https://github.com/valueflows/vocabulator but not exactly what I meant by 'implementations drive the requirements'. Vocabulator seems more trying to follow the vocab, while i meant implementation of real world applications that use the vocab. If you get data dumps from existing applications which consider adopting VF in a future then it would also fit 'implementations drive the requirements'.

bhaugen:
@elf-pavlik I fully understand the distinction, but I also expect vocabulator to drive some implementations. We'll see. Waiting for vientos.

fosterlynn:
>Very likely we will need to work on this PR in systematic way clarifying various current misunderstandings

I'd like to have something we can agree on in a 0.1 release mode, so the efforts that are starting up that will be using VF have something decent to look at.  And so our website reflects something of our current thinking, which it doesn't now (we would switch it to the gitbook).  Or if we are not ready for 0.1, at least let's get the gitbook out there with as much as we know, because stuff is happening!

Of course 0.1 is a long way from 1.0, and we would continue to iterate systematically.  And can drive more from new implementations, if they can have a starting point they can work with.

I'm quite happy to leave off the specific relationships for now, I put them there in a bit in a hurry because of @elf-pavlik concern.  We talked quite a lot about the general relationship model many months ago, it maps somewhat to AS pattern, and it is proven in software, as well as coming from modelers such as Martin Fowler.

So that leaves:
* do we include foaf:Group as a subclass of foaf:Agent?  I'm happy to go with the consensus, as long as the people who discussed it early on get a chance to chime in so we aren't pulling the rug out from anyone.
* can we rename vf:Relationship to vf:AgentRelationship so it is clear what it is about?
* do we put back the old RelationshipType, and if so, what do we call it?  I called it AgentRelationshipVerb, but that is a lousy name.  Maybe AgentRelationshipRole? 
* anything else?

elf-pavlik:
> I'm quite happy to leave off the specific relationships for now, I put them there in a bit in a hurry because of @elf-pavlik concern.

:+1: i think we can take some more time to figure that out

>* do we include foaf:Group as a subclass of foaf:Agent? I'm happy to go with the consensus, as long as the people who discussed it early on get a chance to chime in so we aren't pulling the rug out from anyone.

Currently, because of *vf:work* related difference in *behavior*, in code one needs to only distinguish for agents a person from non-person where we can reuse `foaf:Person` and `(foaf/org):Organization` already defined as **disjoint** subclasses of `foaf:Agent`. Organization can automatically become `foaf:Group` as soon as it specifies having members but this wouldn't matter for implementations which only MUST handle `foaf:Person` and `org:Organization` as disjoint.

We should clearly say that in general implementations should anticipate that in all cases `rdf:type` (`@type` in JSON-LD) can have multiple values. Which doesn't mean that vocab requires using multiple values of `rdf:type` in any place.

>* can we rename vf:Relationship to vf:AgentRelationship so it is clear what it is about?

:+1: 

>* do we put back the old RelationshipType, and if so, what do we call it? I called it AgentRelationshipVerb, but that is a lousy name. Maybe AgentRelationshipRole?

I don't think we should call it `vf:RelationshipType` since we use instances of `rdf:Property`, Type suffix may lead to expectation of `owl:Class` instances. Why don't we just reference `rdf:Property` directly at least to get started?


>Anyone can define any more specific variants of both `foaf:Person` and `org:Organization` but for VF only that distinction matters based on the first point about *vf:work*. This makes easy decision between Person or non-person - Organization. Domain specific cases like *Single Parent* or *Refugee* would stay from VF perspective `foaf:Person`, cases like *Jolly Circle*, *Foursome*, *Fight Club* etc. can all stay from VF perspective an `org:Organization`. We still may need to take some time if we want to use >*rdfs:subClassOf* do define more specialized versions of rely on something like *vf:category* which can reference items from taxonomies.
>* cooperative: https://www.wikidata.org/entity/Q4539
>* NGO https://www.wikidata.org/entity/Q79913
>* ...

I propose to capture above in dedicated issue and in this PR not attempt to handle such categorizations of agents which don't affect how they *behave* in a way relevant to VF vocab.

fosterlynn:
>I don't think we should call it vf:RelationshipType since we use instances of rdf:Property, Type suffix may lead to expectation of owl:Class instances. Why don't we just reference rdf:Property directly at least to get started?

:+1:  for LOD.  We will need something for the non-LOD implementations, since there aren't free-range properties there.  If there are no objections, I'll go with AgentRelationshipRole for now?

elf-pavlik:
> We will need something for the non-LOD implementations, since there aren't free-range properties there. If there are no objections, I'll go with AgentRelationshipRole for now?

Possibly we will not need it in data model but only some implementations might need it based on programming paradigm they follow. Looking at definition of `org:memberOf`
```ttl
org:memberOf a owl:ObjectProperty, rdf:Property;
    rdfs:domain foaf:Agent;
    rdfs:range  org:Organization.
```

It stays defined as instance of `owl:ObjectProperty`, `rdf:Property` so we shouldn't make impression that one can only use here instances of something more specific than `owl:ObjectProperty`. On the other hand if we define

```ttl
vf:relationship a owl:ObjectProperty, rdf:Property ;
    rdfs:domain vf:AgentRelationship ;
    rdfs:range  vf:AgentRelationshipProperty .
```

whenever `org:memberOf` gets used as object in statement using `vf:relationship` as predicate, RDFS reasoner based on its `rdfs:range` would infer

```ttl
org:memberOf a owl:ObjectProperty, rdf:Property, vf:AgentRelationshipProperty .
```

In other words if we define something like `vf:AgentRelationshipProperty rdfs:subPropertyOf rdf:Property`. We should make it clear that one can use any instance of `rdf:Property` or more precisely `owl:ObjectProperty` and don't expect that it has to have in its definition `some:property a vf:AgentRelationshipProperty`. People shouldn't get false impression - "this property wasn't defined as `vf:AgentRelationshipProperty` so I can't use it as 'value' of `vf:relationship`.

fosterlynn:
>In other words if we define something like vf:AgentRelationshipProperty rdfs:subPropertyOf rdf:Property. We should make it clear that one can use any instance of rdf:Property or more precisely owl:ObjectProperty and don't expect that it has to have in its definition some:property a vf:AgentRelationshipProperty

This would be fine.  But I'm also fine with just using rdf:Property in LOD, and naming the class in UML.  (We just still need a name in non-LOD data and modeling structures that don't support properties outside of a class.)

Missing element: we need the inverse label, I looked at issue #52 and didn't understand the options well enough, I leave it to you LOD people to pick the best way.  Does that affect use of rdf:Property?

elf-pavlik:
> Missing element: we need the inverse label, I looked at issue #52 and didn't understand the options well enough, I leave it to you LOD people to pick the best way. Does that affect use of rdf:Property?

We don't need 'inverse label' some properties have inverse counterparts defined in vocab. If not one may define inverse property as blank node just to attach label to it, as explained in http://richard.cyganiak.de/blog/2006/06/an-rdf-design-pattern-inverse-property-labels/

```ttl
skos:subject rdfs:label "has subject" .
skos:subject owl:inverseOf [ rdfs:label "is subject of" ] .
```

So properties need only `rdfs:label`, some will stay defined instances represented by named node and some as 'blank node' just to attach label to it.

fosterlynn:
> if we define something like vf:AgentRelationshipProperty

Actually, I'm re-thinking.  Do you have any objection to using AgentRelationshipRole, just for the UML diagram, and in the LOD spec it would just be a rdf:Property?  The name ...Property I think is confusing to non-LOD people, and at some point we might find we need it to be n-ary.

fosterlynn:
I committed the things we more or less agreed on and fixed some bugs (missing properties, mismatched names, etc.).

So I think we are just down to do we include foaf:Group, for this PR. :)

docs/introduction/agents.md
```diff
@@ -1,24 +1,33 @@
...
- 
+Agent relationships have many nuances, thus the ability to define one's own kinds of relationships.  For example people might "participate" with an organization by means of agreeing to terms and conditions.  Or people might have more active "membership" in a group or organization.  Or people might consider themselves members but want a more independently flavored term such as "affiliates".
+
+A relationship can be direct, like "steward", or more like a role, for example "grower" or "harvester" for a food network.
+
+There are a number of useful Properties in existing vocabularies that can be used. Or people can create their own as needed.
+
+Relationships have direction: For example, in "Michael is a member of Enspiral", Michael is the subject and Enspiral is the object.  In this case the inverse is also valid, "Enspiral has member Michael".
+
+Relationships can be in a context (or not): For example, "Kathy is mentor of Sam, in the context of Enspiral."
```
ahdinosaur: :+1:

release-doc-in-process/all_vf.TTL
```diff
@@ -22,13 +23,18 @@ vf:Action  a           owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "An action verb defining the kind of event, commitment, or intent." .
 
+#vf:AgentRelationshipRole  a  owl:Class ;
```
ahdinosaur: nice. is there a reason this is commented out?

release-doc-in-process/all_vf.TTL
```diff
@@ -218,31 +226,30 @@ vf:committedCategory
 #        rdfs:domain         vf:ExchangeAgreement ;
 #        rdfs:range          vf:ExchangeTaxonomyItem .
 
-vf:parent
-        a                   owl:ObjectProperty ;
-        rdfs:label          "parent" ;
-        rdfs:domain         vf:ResourceTaxonomyItem ;
-        rdfs:range          vf:ResourcetaxonomyItem ;
-        rdfs:comment        "A more general or broader category of resource." .
+#vf:parent
```
ahdinosaur: `skos:broader`?

release-doc-in-process/all_vf.TTL
```diff
@@ -82,26 +88,28 @@ vf:EconomicEvent  a            owl:Class ;
         rdfs:label       "vf:EconomicEvent" ;
         rdfs:comment   "An inflow or outflow of an economic resource in relation to a process and/or exchange. This could reflect a change in the quantity of a vf:EconomicResource. It is also defined by its behavior in relation to the vf:EconomicResource and a vf:Process (consume, use, produce, etc.)" .
 
-vf:Agent  a                  owl:Class ;
-        rdfs:label           "vf:Agent" ;
-        owl:equivalentClass  foaf:Agent ;
-        rdfs:comment   "A person or group with economic agency." .
+foaf:Agent  a                  owl:Class ;
```
ahdinosaur: what's the motivation for using `foaf` / `org` directly instead of subclassing as `vf`? i'm happy if there's a link to a discussion.

release-doc-in-process/all_vf.TTL
```diff
@@ -218,31 +226,30 @@ vf:committedCategory
...
 vf:underlyingResource
         a                   owl:ObjectProperty ;
         rdfs:label         "underlying resource" ;
-        rdfs:domain         vf:Resource ;
-        rdfs:range          vf:Resource ;
+        rdfs:domain         vf:EconomicResource ;
+        rdfs:range          vf:EconomicResource ;
         rdfs:comment        "A more concrete resource upon which a resource is based." .
 
 vf:relationship  a          owl:ObjectProperty ;
```
ahdinosaur: i'm very confused about how the domain and range works here, can anyone explain to a newbie?

release-doc-in-process/all_vf.TTL
ahdinosaur: nevermind, duh: https://github.com/valueflows/valueflows/issues/182

release-doc-in-process/all_vf.TTL
fosterlynn: Because @elf-pavlik prefers to just have the relationship be a rdf:Property, instead of creating our own subclass of the rdf:Property, and I have agreed to that.  The name would remain in the UML for non-LOD implementations.

release-doc-in-process/all_vf.TTL
fosterlynn: Discussion: #182

release-doc-in-process/all_vf.TTL
fosterlynn: domain is what class(s) it is a property for; range is what it is in the way of an object (think of like a foreign key)

release-doc-in-process/all_vf.TTL
ahdinosaur: okay cool

release-doc-in-process/all_vf.TTL
ahdinosaur: oh, so it's saying an AgentRelationship has a property "relationship" which points to a Property (presumably a AgentRelationshipRole)? yeah that makes sense, the terminology and how LOD works is confusing to me.

fosterlynn: you got it!  yeah, different terminology and some differences in the possible data structures too

fosterlynn:
>So I think we are just down to do we include foaf:Group, for this PR. :)

@ahdinosaur are you OK with removing foaf:Group, and having org:Organization as the only choice for a non-person economic agent?  You were a big part of the initial discussions where we said we wanted a concept for Group, so I want to make double sure it is OK with you.  And as it might apply to how Enspiral thinks of its various pods, etc.

So it would be:
foaf:Agent
--foaf:Person
--org:Organization

elf-pavlik:
> @ahdinosaur are you OK with removing foaf:Group, and having org:Organization as the only choice for a non-person economic agent? As in, if you are an economic agent, you are either a foaf:Person or an org:Organization.

Please keep in mind my comment that we still need a way for categorizing `foaf:Person` and `org:Organization` further. Still currently for requirements specific to the vocab itself we only need to distinguish a Person ('foaf:Person') from 'Non Person Agent' ('org:Organization') because only a Person can provide *vf:work*.

> Anyone can define any more specific variants of both foaf:Person and org:Organization but for VF only that distinction matters based on the first point about vf:work. This makes easy decision between Person or non-person - Organization. Domain specific cases like Single Parent or Refugee would stay from VF perspective foaf:Person, cases like Jolly Circle, Foursome, Fight Club etc. can all stay from VF perspective an org:Organization. We still may need to take some time if we want to use rdfs:subClassOf do define more specialized versions of rely on something like vf:category which can reference items from taxonomies.

* cooperative: https://www.wikidata.org/entity/Q4539
* NGO https://www.wikidata.org/entity/Q79913
* ...

elf-pavlik:
> So it would be:
foaf:Agent
--foaf:Person
--org:Organization

after we remove `foaf:Group` and have just those two :+1: to merge this PR

ahdinosaur:
> @ahdinosaur are you OK with removing foaf:Group, and having org:Organization as the only choice for a non-person economic agent?

yeah keen on that for now, seem like a great way to start and if it becomes problematic we can always change it later.

fosterlynn:
Merging, but will take a final look around for obvious bugs, and also try to get the LODE spec to work better before tagging.

Issue #207, opened 2017-03-24
Get gitbook working with lode


fosterlynn:
Ref. #184 -- this is a detail of that one.

@djodjoni introduced [LODE](http://www.essepuntato.it/lode) for formatting our class and property specs from the .ttl file, and it works great, is open source, love it.  @pmackay set up Gitbook for us to move all the VF doc into, works great, open source, love it.

I'm having trouble getting my generated LODE page to work inside Gitbook.  @djodjoni is this something you would like to put some time into while (or instead of) working on the .ttl validation on PRs?  It would be really helpful.

So far: I tried just calling our generated LODE page from a gitbook side bar in SUMMARY.md, and it generates the page, but as a new tab, maybe OK but not the best.  So then I saved the source from our generated page, made it a .md file, and connected it like any other page on the side bar in SUMMARY.md.  It stays in the gitbook structure fine, but lost its style.  I thought maybe it was just a matter of css relative references, but all of them are full path.  I think the problem is that the stylesheets are referenced hidden inside whatever div is created by gitbook?

One possible solution I found: https://github.com/chudaol/gitbook-plugin-addcssjs.  But maybe there is something much easier and more obvious, I don't know.

Anyhow, I'd love to hand over this task to someone a bit more techy than me. :)

djodjoni:
Just back from a few busy days and seeing lots going on 👍  
I'll try to make this work

fosterlynn:
hi @djodjoni thanks!  

I put it back to where it just opens the LODE spec in another tab, with its styling intact.  I'm actually getting used to that - well a little used to it. :)  Why don't you take a look and see what you think is worth spending time on. 

Another option is to create a front page which links into the gitbook (with spec removed) and the spec separately.  If we do that, we could do a prettier front page.

What direction do you think we should go?

djodjoni:
I took a look and it can work also if we put the styles in the book, however I think I like the LODE specs as a separate page indeed. 
In fact I would point https://w3id.org/valueflows/ to the LODE specs if the browser requests text/html or the .ttl in the case of text/turtle, .jsonld file in case of browser accepting "application/ld+json" etc. 

> Another option is to create a front page which links into the gitbook (with spec removed) and the spec separately. If we do that, we could do a prettier front page.

I personally like the gitbook generated site. it is simple and clear I don't think there is a need of 'prettier' page :)

fosterlynn:
OK let's go with your call @djodjoni unless others disagree....  :)

[...]

PR #211, opened 2017-03-28
add rdf and html formats for the ontology

djodjoni:
this adds various formats that will be opened from redirection from w3id
These are just to test if w3id redirection is ok.  

I will make another PR to generate those on PR or Merge later tonight or in the next days.


djodjoni:
oh yea, this is related to #207

Issue #187, opened 2017-02-27
agent - resource type/etc. - agent


fosterlynn:
Both @pmackay and @djodjoni have expressed a desire for something like agent - resource type - agent.  It would basically say something like "FarmerX provides carrots to FoodHubY".  You could probably get this from commitments or past events, but what if you don't want to implement all of that?  We talked a bit about maybe trying out Intents for this.  It is kind of like an intent, but not exactly.

We also use something like this in [LocEcon](http://locecon.org), which would be theoretically happy to get this kind of data from operational systems, but in fact mostly does not have access to it, and doesn't have a way to gather it yet.

@bhaugen and I talked about this over breakfast, and I want to propose we add this structure to VF, and figure out what would be included.

Something like vf:AgentResourceFlow?

Properties: vf:provider, vf:receiver, vf:category (?? - do we need anything more specific than category?).  Might need a vf:action, although that might be 2 action verbs, need to try some things out; or maybe as long as we have the direction, that is all that is needed, I don't know.  Might need some dates/durations kind of things.  Just brainstorming.

What do people think about including something like this?  On one hand, I want to try to get a minimal agreed upon vocab released for basic REA; on the other hand, here are some needs expressed, I think good to go for it while people are actively thinking about it.  And I think there will be even higher level constructs needed for community or regional planning, which also might or might not have a link to operational data.

pmackay:
Thanks for the suggestion @fosterlynn :) Just to check, do you see introducing vf:AgentResourceFlow as preferable to using Intents as discussed on Friday?

Also if start and end Locations needed to be defined, might they be part of this or are you seeing this only working with an Agent per Location?

fosterlynn:
>Just to check, do you see introducing vf:AgentResourceFlow as preferable to using Intents as discussed on Friday?

Well, on thinking about it after we talked, I do think Intents have a somewhat different meaning, and also wouldn't really generally include 2 agents (I think, could be wrong).  But what do you think?

>Also if start and end Locations needed to be defined, might they be part of this or are you seeing this only working with an Agent per Location?

Locations would be great.  Hmmm, would we want locations for the ends of the flow separate from the agents' locations?  Maybe for the example of unmanned recycling centers?  Or is that getting too far into the weeds for something that provides more general information?

pmackay:
[...]

PR #219, opened 2017-04-16
solution/discussionaction for #187

djodjoni:
this is meant as a basis for concrete discussion of #187

### DIFF
--- a/.scripts/deploy.sh
+++ b/.scripts/deploy.sh
@@ -5,6 +5,7 @@ SOURCE_BRANCH="master"
 TARGET_BRANCH="gh-pages"
 
 function doCheck {
+    mkdir .out
     java -jar .scripts/lod-converter.jar release-doc-in-process/all_vf.TTL TTL vf https://valueflows.github.io/valueflows/vf.ttl
 }
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -107,9 +107,44 @@ org:Organization  a         owl:Class ;
         rdfs:subClassOf      foaf:Agent ;
         rdfs:comment      "Usually a more formally defined structure than Group, often legally defined, can be a group or not." .
 
-vf:AgentRelationship  a       owl:Class ;
+vf:AgentRelationship  a  owl:Class ;
         rdfs:label       "vf:AgentRelationship" ;
         rdfs:comment    "An ongoing voluntary association between 2 Agents of any kind." .
+
+vf:subject  a        owl:ObjectProperty ;
+        rdfs:domain  vf:AgentRelationship ;
+        rdfs:label   "subject" ;
+        rdfs:range   foaf:Agent ;
+        rdfs:comment    "The subject of a relationship between 2 agents." .
+
+vf:context  a        owl:ObjectProperty ;
+        rdfs:domain  vf:AgentRelationship ;
+        rdfs:label   "context" ;
+        rdfs:range   foaf:Agent ;
+        rdfs:comment        "The larger context of a relationship between 2 agents, used where the relationship is not relevant outside of that context." .
+
+vf:object  a         owl:ObjectProperty ;
+        rdfs:domain  vf:AgentRelationship ;
+        rdfs:label   "object" ;
+        rdfs:range   foaf:Agent ;
+        rdfs:comment        "The object of a relationship between 2 agents." .
+
+vf:relationType  a    owl:ObjectProperty ;
+		rdfs:comment  "generic defined relationship that exists between 2 agents." ;
+		rdfs:domain   vf:relationType ;
+		rdfs:label    "relationship" ;
+		rdfs:range    owl:Thing .
+
+vf:AgentResourceFlow  a  owl:Class ;
+		rdfs:comment  "AgentResourceFlow relationship" ;
+        rdfs:subClassOf vf:AgentRelationship .
+
+vf:flowingResource  a          owl:ObjectProperty ;
+		rdfs:comment  "the resource that flows from subject to the object" ;
+		rdfs:domain   vf:AgentResourceFlow ;
+		rdfs:range   vf:ResourceTaxonomyItem .
+
+
 
 #vf:Fulfillment  a            owl:Class ;
 #        rdfs:label       "vf:Fulfillment" ;
@@ -247,12 +282,6 @@ vf:underlyingResource
         rdfs:range          vf:EconomicResource ;
         rdfs:comment        "A more concrete resource upon which a resource is based." .
 
-vf:relationship  a          owl:ObjectProperty ;
-        rdfs:label         "relationship" ;
-        rdfs:domain         vf:AgentRelationship ;
-        rdfs:range          rdf:Property ; # think about creating a subproperty for these
-        rdfs:comment        "A verb that describes a generic defined relationship that exists between 2 agents." .
-
 #vf:commitsTo
 #        a                   owl:ObjectProperty ;
 #        rdfs:domain         vf:Commitment ;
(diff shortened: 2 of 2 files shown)


fosterlynn:
```
+vf:relationType  a    owl:ObjectProperty ;
+		rdfs:comment  "generic defined relationship that exists between 2 agents." ;
+		rdfs:domain   vf:relationType ;
+		rdfs:label    "relationship" ;
+		rdfs:range    owl:Thing .
```

Question: the domain should be vf:AgentRelationship?

Comment: in terms of the name of the property, I'd like relationship instead of just relation... and I'm fine with Type, but if @elf-pavlik doesn't like it here, then I'm fine without it.  Thinking now, it would also be nice to have the "agent" in it too, so we don't get conflicts with other relationships later on.  So how about just  "vf:agentRelationship"?

I'm ok with the range of owl:Thing for now; but we have discussed at some point defining something about the base behaviors of different kinds of relationships.... but later is fine....

Also, checking with @pmackay does this work for your needs?

fosterlynn:
Hmmm, I forgot about this poor thing:
```
#vf:AgentRelationshipRole  a  owl:Class ;
#        rdfs:label           "vf:AgentRelationshipRole" ;
#        rdfs:subClassOf   rdf:Property ;
#        rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
```
In the release, I left it in the UML because non-rdf implementations would need something more than a property.  But commented it out of the ttl.  

I guess it is a question of how many constraints we want on this, and perhaps in general.  Clear definitions for interoperability vs more flexibility?  I obviously lean towards clearly defined constraints for interoperability, but I do get the freedom the folks with rdf background appreciate.

I suppose if we reinstated it, even with the property having a range of anything, people could use it if they want others to be able to have expectations of what to find there.  Would have to document it, there wouldn't be a way to know in the definition.

OK, just musing at this point.... thoughts?

djodjoni:
> Question: the domain should be vf:AgentRelationship?

absolutely correct :) my mistake! ill fix that

> Comment: in terms of the name of the property, I'd like relationship instead of just relation... and I'm fine with Type, but if @elf-pavlik doesn't like it here, then I'm fine without it. Thinking now, it would also be nice to have the "agent" in it too, so we don't get conflicts with other relationships later on. 

yer relationshipType is better that relationType. I just put the type as it is a type of relationship and could be part of taxonomy(for example in agrovoc there many cases like this) and to differ the property from the actual class. But I would not mind to change the workiding

> So how about just "vf:agentRelationship"?

This is confusing IMO as it has the same name as the object that is range of. I think it is better to distinguish in this case the class vf:AgentRelationship used as a binder to relate 2 agents and the type of relationshit they are binded with. For example in holodex app "relationshipType" is used :)

> In the release, I left it in the UML because non-rdf implementations would need something more than a property. But commented it out of the ttl.

I am confused vf:AgentRelationshipRole  as it is defined is a subclass of property. Also in the current relationshipType property the range cam be any complex instance of a possible relationship. 
I don't really see what is the use of AgentRelationshipRole :) could you explain please what do you want to uise it for :)

elf-pavlik:
> I am confused vf:AgentRelationshipRole as it is defined is a subclass of property. Also in the current relationshipType property the range cam be any complex instance of a possible relationship.
I don't really see what is the use of AgentRelationshipRole :) could you explain please what do you want to uise it for :)

An example

```yaml
'@context': https://w3id.org/valueflows/v1-dev
id: https://coop.example/cj1y86fyd0000t8lpfw0a0nfd#relationship
type: vf:AgentRelationship
subject: https://coop.example/#identity
object: https://alice.example/#identity
relationship: http://www.w3.org/ns/org#hasMember
```

Which pretty much represents qualified (in a way similar to how [Reification](http://patterns.dataincubator.org/book/reified-statement.html) works) version of

```yaml
'@context': https://w3id.org/valueflows/v1-dev
id: https://coop.example#identity
'http://www.w3.org/ns/org#hasMember': https://alice.example/#identity
```

djodjoni:
What I would like to avoid is exactly Reification-like-approach where it is not needed for example if we don't need additional info like context then I would use this
```
'@context': https://w3id.org/valueflows/v1-dev
id: https://coop.example#identity
'http://www.w3.org/ns/org#hasMember': https://alice.example/#identity
```
as Reification as it states in its description says:
> The obvious limitation of this approach is that the data model no longer contains a simple set of triples, instead it contains descriptions of triples. The triples themselves have not been explicitly asserted. This makes a reified model harder to query, more difficult to manipulate, and at least three or four times larger in terms of number of triples.

In other words I would define triples as usual having an already defined known ObjectProperty. In addition to that "vf:AgentRelationshipRole" does not need to be defined as a subclass of property as there is no advantage if used like that:

```
 '@context': https://w3id.org/valueflows/v1-dev
id: https://coop.example/cj1y86fyd0000t8lpfw0a0nfd#relationship
type: vf:AgentRelationship
subject: https://coop.example/#identity
object: https://alice.example/#identity
relationship: http://www.w3.org/ns/org#hasMember
```

elf-pavlik:
https://www.w3.org/TR/vocab-org/#class-membership

> Represents an n-ary relationship between an Agent, an Organization and a Role. It is possible to directly indicate membership, independent of the specific Role, through use of the org:memberOf property. 

I think we need something similar to `org:Membership` but more generic (so not only for membership in organization)

`vf:AgentRelationship` currently works similar to https://www.w3.org/TR/activitystreams-vocabulary/#dfn-relationship

You can also find examples of [N-ary relations](https://www.w3.org/TR/swbp-n-aryRelations/) in http://smiy.sourceforge.net/cco/spec/cognitivecharacteristics.html#sec-glance

cco which IMO gives interesting example of `owl:propertyChainAxiom`

Another interesting example we can find in https://www.w3.org/TR/vocab-org/#org:roleProperty
> The intended semantics is that a Membership relation involving the Role implies the existence of a direct property relationship through an inference rule of the form: `{ [] org:member ?a; org:organization ?o; org:role [org:roleProperty ?r] } -> {?a ?r ?o}`

I'd like to take a closer look at reasoning using rules expressed in N3 https://rubenverborgh.github.io/Semantic-Web-Reasoning/#

djodjoni:
> I think we need something similar to org:Membership but more generic (so not only for membership in organization)

I agree, in fact this pretty much what I have imagined also.
In our case  "vf:AgentRelationshipRole" is identical to "org:Role", which is a Concept which I also like as I imagine different role type to be some kind of taxonomy and there SKOS is perfect for that.
"org:roleProperty" is also very interesting as it may link to actual rdf:Property, which I think you try to achieve directly on "vf:AgentRelationshipRole". 
As regard to querying it would be indeed easier. I have been doing that with Jena and add rules to the reasoner and then query on the Model, its quite easy.

So shall I change "vf:AgentRelationshipRole" to Concept and add "vf:roleProperty" ?

elf-pavlik:
If we want to move away from how https://www.w3.org/TR/activitystreams-vocabulary/#dfn-relationship works (which tries to use naming similar to reification). We may need to also rename *vf:subject*, *vf:object* and *vf:relationship*. Let's take some time to study org-vocab more, including constructs like [org:transitiveSubOrganizationOf](https://www.w3.org/TR/vocab-org/#org:transitiveSubOrganizationOf) and if for such cases we would recommend using org-vocab directly or try to do it with *vf:AgentRelationship*.

djodjoni:
I am not sure I understand. I don't want to move away from that nor change namings.
If we abstract from the namings and the ranges of subject and object(member and org), we see that org-vocab and activity-streams are quite the same except that org-vocab has this option to link to actual Properties enhancing queries. 
For example org-vocab:role has range org-vocab:Role which is some Concept in activitystreams the equivalent is relationship property with range Object which is the base type in their context.

so again I see no changes in namings nor meanings just adding this roleProperty.

fosterlynn:
>In our case "vf:AgentRelationshipRole" is identical to "org:Role", which is a Concept which I also like as I imagine different role type to be some kind of taxonomy and there SKOS is perfect for that.

>so again I see no changes in namings nor meanings just adding this roleProperty. and change the subtype of AgentRelationshipRole

I like making AgentRelationshipRole into a taxonomy, and SKOS is ok with me. :+1:  

Just a note: vf:AgentRelationshipRole is broader than org:Role, because org:Role is only a role in an org:Organization.

Re. roleProperty, here is the org definition:
`This is a metalevel property which is used to annotate a org:Role instance with a sub-property of org:memberOf that can be used to directly indicate the role for ease of query.  The intended semantics is that a Membership relation involving the Role implies the existence of a direct property relationship through an inference rule of the form: { [] org:member ?a; org:organization ?o; org:role [org:roleProperty ?r] } -> {?a ?r ?o} `

I could use help understanding exactly how we would use something like that (not arguing; and I haven't studied SPARQL much).  Would it apply to all kinds of roles, not just those that are subs of membership, I presume? And is that the same basically as our role label (and inverse label)?

fosterlynn:
>I think we need something similar to org:Membership but more generic (so not only for membership in organization)

Question: what else would you have that is a membership that is not in an organization in our group-less model?

I agree the term "membership" seems not quite right to some people (like Sensorica) but not because of the organization part.  They prefer "affiliate" because it is looser sounding.

PR #223, opened 2017-05-13
Scope of events, commitments, eventually intents

fosterlynn:
For everyone's review.  This came out of recent gitter discussions.  Used to be called "context".  

If you have questions, then probably the doc isn't clear enough.  :)


docs/introduction/scope.md
```diff
@@ -0,0 +1,29 @@
+# Scope
+
+Scope defines a boundary or context for sets of economic events, commitments, and intents.  This can be for purposes of:
+* accounting
+* planning
+* distributing income
+
+Often the "scope" will be an organization (an agent), where work is done, where processes live, where value is created by groups of people. It may be formal or informal organizations, and will include the network(s) themselves.  A "scope" can also be a geographic area (a city, a bioregion, etc.), or a community, or a loose network, or a less formal shorter lived project, or any other bounded concept that is useful for grouping economic events.
```
elf-pavlik: I wouldn't say:
> Often the "scope" will be and organization (an agent), ...
> A "scope" can also be a geographic area (a city, a bioregion, etc.), ...

And look for something in direction of
> Often economic events and commitments will reference an organization as entity that defines their scope.
> Economic events and commitments can reference kind of entity as defining their scope, for example a geographic area ...

release-doc-in-process/all_vf.TTL
```diff
@@ -322,17 +322,35 @@ vf:subject  a        owl:ObjectProperty ;
...
         rdfs:label   "context" ;
         rdfs:range   foaf:Agent ;
         rdfs:comment        "The larger context of a relationship between 2 agents, used where the relationship is not relevant outside of that context." .
 
-vf:object  a         owl:ObjectProperty ;
-        rdfs:domain  vf:AgentRelationship ;
-        rdfs:label   "object" ;
-        rdfs:range   foaf:Agent ;
-        rdfs:comment        "The object of a relationship between 2 agents." .
+vf:scope  a        owl:ObjectProperty ;
```
elf-pavlik: If we try to clearly capture **the relationship** that this property represents. Maybe *inScopeDefinedBy* would stand as clearer property name?

elf-pavlik:
Thank you for preparing PR @fosterlynn!

Besides addressing two comments I added in review, I think we also should include somewhere a paragraph: *"Each economic events and commitment can reference any number of entities as defining scopes it belongs to"*

docs/introduction/scope.md
fosterlynn: sounds good, will fix later and commit

release-doc-in-process/all_vf.TTL
fosterlynn: kind of long.... how about *inScope*?

elf-pavlik: maybe *scopedBy* ?

docs/introduction/scope.md
```diff
@@ -0,0 +1,29 @@
+# Scope
```
elf-pavlik: Can we rename it to *Scoping* ? To avoid confusion that we have something with `type: vf:Scope`. But we can use references to anything to scope events and commitments.

gcassel:
A great basis for developing this important concept!  For what it's worth, "scope" seems to be used frequently and efficiently in Enspiral-related activity.

I use "context" broadly and frequently.  "Scope" seems highly appropriate for *intended context*-- and I think that "intended context" is the proper 'scope' of scope for VF.

Small suggestion, probably not worth a pull request by itself:

>Scope defines intended boundaries or context for sets of economic events, commitments, and intents. 

My reasoning for "boundaries" is that IMO, scope usually indicates (at least indirectly) at least 2-3 intended boundaries.   (For instance, a grouping of agents + a purpose within that group.)

fosterlynn:
@gcassel 

>My reasoning for "boundaries" is that IMO, scope usually indicates (at least indirectly) at least 2-3 intended boundaries. (For instance, a grouping of agents + a purpose within that group.)

I think you would have to get that defined as one thing in some way to reference it as scope, maybe one boundary that is defined by multiple criteria?

gcassel:
>I think you would have to get that defined as one thing in some way to reference it as scope, maybe one boundary that is defined by multiple criteria?

That depends utterly on how we define "boundary".   I think that "boundary" is useful in analyzing each criterion/ variable which is used to define a multivariate scope-- or, for that matter, all of our complexly "emergent" perceptions and definitions.   

I just spent 15 minutes typing longwinded examples to support my view that each "definitive" criterion contains one or more separately identifiable boundaries.   Instead of subjecting you all to that, however:  perhaps we could simply switch to something like simple like

>>Scope describes the intended context for sets of economic events, commitments, and intents.

Does that seem too simple...?  It would avoid the whole subject of indicating that contexts often describe several strictly delimiting criteria, but often are worded quite informally.

bhaugen:
I assume scope will be optionally referenced on events, commitments, and intents. So the boundary will be defined by whatever the people posting those things reference the scope on.

In real life, some scopes, like geographic areas, might have publicly defined boundaries. Other times, informal or conceptual or metaphoric or even looser. All ok with me.

fosterlynn:
>I assume scope will be optionally referenced

yes

>In real life, some scopes, like geographic areas, might have publicly defined boundaries. Other times, informal or conceptual or metaphoric or even looser. All ok with me.

Me too.  It can be defined however people want, but it does have to have a way to reference it as one thing, according to the model as it stands now anyhow.  It isn't a search, or series of and's and or's, or other complex kind of reference.  Just one URI.  Sorry for the more technical perspective, the looseness or plural-ness will need to be defined behind the one reference - which should be fine I think.

@gcassel If you feel strongly about the word "boundary", I'll take it off without complaint.  I personally think it adds to people's understanding though, from the discussions we had.

gcassel:
>@gcassel If you feel strongly about the word "boundary", I'll take it off without complaint. I

I'm okay with "boundary".  I prefer "boundaries" for many reasons which I've not described thoroughly here.   I would prefer "limits" to either of those, but didn't desire to complicate the discussion further.

PR #234, opened 2017-08-11
Made fulfillment a M:M between economic event and commitment

fosterlynn:
History: We did a process model then an exchange model, and they fulfilled commitments differently, with exchange being a M:M, and process being a 1:M.  This resolves that in favor of a M:M for all events and commitments, in the interest of simplifying the model.  We had a lot of discussion about this, and I believe @elf-pavlik was in favor of this solution, while I was concerned about people fulfilling commitments across processes with one event.  

I now would like to resolve and simplify, and think this is the safest and most flexible.  (This is triggered by the VF graphql api I'm working on for NRP/OCP.)

I'll leave the PR out for several days in case anyone wants to comment.


elf-pavlik:
LGTM thank you @fosterlynn 

It seems that UML diagram besides adding *vf:Fulfillment* also adds bunch of stuff around resource taxonomies. If intended could it go as separate PR?

fosterlynn:
@elf-pavlik good catch, I apparently started some resource taxonomy work a while ago and forgot about it.  I don't even like it anymore, so no separate PR.  Removed from UML.

PR #239, opened 2017-10-12
make consistent property names using owl:unionOf

fosterlynn:
This is not ready for a merge yet, but I'd like some RDF help here.  What I'm trying to do is to clean up and make consistent the property names for various relationships without causing incorrect inference rules in RDF. 

I also added the other direction of the relationships between Process / EconomicEvent and Process / Commitment.  Might want to do more of that in the future.

I think my syntax for the rdfs:domain using owl:unionOf is not correct, at least the webvowl validator doesn't like it.  Seems right from the references on the web though.  @elf-pavlik @djodjoni @pmackay any of you have time to take a quick look?  Not a big rush.

Here is an example:
```
vf:action
        rdfs:label         "action" ;
        rdfs:domain         owl:unionOf (vf:EconomicEvent, vf:Commitment) ;
        rdfs:range          vf:Action ;
        rdfs:comment        "Relates a process input or output to a verb, such as consume, produce, work, improve, etc." .
```


fosterlynn:
P.S. I plan on doing another PR to change vf:context to vf:scope for AgentRelationship, after this one gets cleaned up.

elf-pavlik:
To validate turtle online you can use: http://ttl.summerofcode.be/

With OWL, you can look up syntax at https://www.w3.org/TR/owl2-primer/#Complex_Classes but you have to go first in he same document to https://www.w3.org/TR/owl2-primer/#OWL_Syntaxes and click 'Show Turtle Syntax'

TL;DR
` rdfs:domain [ owl:unionOf (vf:EconomicEvent vf:Commitment) ] ;`

this creates a *blank node* defined as `owl:Class` (implied by domain of owl:unionOf), and states that this class represent union of those two other classes.

fosterlynn:
@elf-pavlik thank you very much for the links and info!!  Fixed it.

PR #252, opened 2017-12-09
resource currentQuantity, reference classification with uri (for now)

fosterlynn:


json-schemas/EconomicResource.json
```diff
@@ -9,13 +9,14 @@
...
         },
-        "quantity": {
+        "currentQuantity": {
             "description": "The current amount and unit of the economic resource.",
             "$ref": "https://raw.githubusercontent.com/valueflows/valueflows/master/json-schemas/QuantityValue.json"
         },
         "resourceClassifiedAs": {
             "description": "A resource's primary classification, as part of describing that resource.",
-            "$ref": "https://raw.githubusercontent.com/valueflows/valueflows/master/json-schemas/ResourceClassification.json"
+            "type": "string",
```
elf-pavlik: Does it work with a single IRI (string) or an array?
Actually with [JSON-LD Embedding](https://www.w3.org/TR/json-ld/#embedding) anywhere where IRI can appear also an object can appear with same IRI as `@id` and other properties.
In general wherever we have IRI reference, we can get
* IRI (string)
* Array of IRIs (strings)
* Object (with that IRI as `@id`)
* Array of Objects (each one with an IRI as `@id`)

fosterlynn: Good point on the single string vs array!  But for actually identifying a model of resource exactly, we will need to know which is that one.  For other purposes, we will need more than one.  I feel like we may need to re-think some of the resource - resource classification stuff.  And Intents will add a whole other set of use cases to it.

Of those choices, are you saying we should make it flexible so people can use any of those?  Or putting out some choices that we should look at?  

I actually made that change in the json-schema for the ssb group because we aren't ready to consider the recipe layer, and this way, we don't have to define ResourceClassification anywhere.  So sort of a practical short cut.

elf-pavlik: > Of those choices, are you saying we should make it flexible so people can use any of those? Or putting out some choices that we should look at?

People can use any of those, at least if they use RDF and serialize it to JSON-LD

Any comment doesn't only apply for this particular case but pretty much to any place where on can use IRI as value in JSON-LD. I think when we link those JSON Schemas from gitbook we should explain somewhere their purpose and that valid VF data serialized to JSON-LD does NOT have to conform to those JSON Schemas.

As alternative to JSON Schemas which may not fully account for flexibility of RDF serialized to JSON-LD I will look at writing some SHACL (data shapes) to go along VF vocab http://shacl.org/playground/

fosterlynn: :+1:   Thanks!  Yeah, I wouldn't think about adding them to the gitbook until we see how they actually work over in ssb, and also we write some different kinds of examples to reflect rdf and whatever else people seem to want to use.  I have no need to do it any time soon, got other higher priorities to work out for the rest of the vocab, like intents and recipes. :)  And totally agree on a bunch of explanation, when we do publish examples.

Issue #254, opened 2017-12-13
Add link to Github repo in docs


pmackay:
Just been browsing through most of the pages in https://www.valueflo.ws/ and it occurred that I could not see a link to the GitHub repo - would it make sense to add one somewhere?

Similarly would a Contributing page or note be useful?

bhaugen:
Yes, please. I guess this is what we got for contributing: https://github.com/valueflows/valueflows/blob/master/CONTRIBUTING.md

fosterlynn:
This was done awhile ago.

PR #255, opened 2017-12-15
Random things I have run into lately trying to use VF

fosterlynn:
Changed EconomicResource.quantity to currentQuantity.  
Changed AgentRelationship.context to scope.  
Added isFinished to Commitment.  
Added plannedStart and plannedDuration to Process in the ttl.  (Not an actual change, just a fix.)
Took off extraneous line in the UML, just a fix.

(If anyone disagrees with one of these, I'll remove it here and take it to a separate PR.)


elf-pavlik:
> Added isFinished to Commitment.

Could you explain this one, how does it supposed to work with *vf:Fulfillment*?

fosterlynn:
>>Added isFinished to Commitment.

>Could you explain this one, how does it supposed to work with vf:Fulfillment?

It assumes that the planned or estimated time could be wrong (often is wrong), so you can't go on that, but may need to close the commitment explicitly.  There also is often more than one event fulfilling a commitment, so you can't assume that because there is a fulfillment that is is complete.

fosterlynn:
And one other thing I forgot to say - if one of the rdf experts could check on my use of xsd:boolean, that would be great.  (That is for isFinished.  And I had to add the @prefix up top too.)

fosterlynn:
Also added link to github and welcom chat into the intro in the gitbook to resolve #254

fosterlynn:
Also added link to github and welcom chat into the intro in the gitbook to resolve #254 

Removed isFinished from the ttl for now until resolved.  Merging to get the other stuff in.

PR #256, opened 2017-12-26
example json event to create resource

fosterlynn:
including new resource data on the resource on the event... but no reference to an id of the resource, thinking this could be a way to indicate a new resource rather than affecting an existing resource.... (all experimental)

### DIFF
--- a/json-schemas/sample-create-resource.json
+++ b/json-schemas/sample-create-resource.json
@@ -1,0 +1,26 @@
+[
+    {
+        "id": "E10-event-to-create-resource",
+        "action": "take",
+        "start": "2017-12-01",
+        "affects": {
+            "resourceClassifiedAs": "https://www.wikidata.org/wiki/Q41777871",
+            "currentLocation": "Mikeys basement NW corner",
+            "note": "Last haralson apples of the season, good storage apples.",
+            "trackingIdentifier": "5nw-haralson"
+        },
+        "affectedQuantity": {
+            "numericValue": "10",
+            "unit": "kg"
+        },
+        "provider": {
+            "id": "A1",
+            "name": "Bob"
+        },
+        "receiver": {
+            "id": "A5",
+            "name": "Mikey"
+        }
+    }
+
+]


elf-pavlik:
Why not create a new resource first with `currentQuantity: 0` ?
If you have two events like above but intend it for two different resources, you may have hard time later to avoid ambiguity.
For cases where service allows clients to mint URLs (based on some guid component in that url) you could even just create `@id` for that new resource in that event and service would create the resource based on that event.

bhaugen:
@elf-pavlik thanks for commenting.  Hope I will make sense to you.
> Why not create a new resource first with currentQuantity: 0 ?

I think that would work, except that one of the principles of economic events is that an agent should be able to compute or recreate all of their resources from their event history, and validate their resource positions, if they instantiate resources and keep track of the current balances, by comparing to their events.

We're having conversation about that in SSB right now, where some people want to compute resource balances on the fly by summarizing events, because that fits their expectations of a gossip network better than persistently instantiated resources. That's what Bill McCarthy does, too.

> If you have two events like above but intend it for two different resources, you may have hard time later to avoid ambiguity.

Would need a unique identifier. 

> For cases where service allows clients to mint URLs (based on some guid component in that url) you could even just create @id for that new resource in that event and service would create the resource based on that event.

I think I follow, and I think that would work.

fosterlynn:
>>If you have two events like above but intend it for two different resources, you may have hard time later to avoid ambiguity.

>Would need a unique identifier.

In ssb they are talking about using the hash of the event message to create the resource id, maybe with a timestamp too, or something.  Still in discussion/experimentation.  

I left off the id in this json to indicate the resource needs to be created by this event.  If another event comes through to say subtract 5kg from the resource that got created, then it would need to reference the resource by its id, which by that time would be known.

fosterlynn:
P.S. I'm not sure how this would work best in linked open data.....

elf-pavlik:
> In ssb they are talking about using the hash of the event message to create the resource id, maybe with a timestamp too, or something. Still in discussion/experimentation.

IMO event message preferably should include the `@id` of resource, even if client just generated a new UUID or [CUID](https://github.com/ericelliott/cuid) based IRI.

Relying on hashing would introduce SSB specific interpretation of the data, depending on some SSB specific choice of hashing algorithms, serialization nuances etc.

fosterlynn:
>Relying on hashing would introduce SSB specific interpretation of the data, depending on some SSB specific choice of hashing algorithms, serialization nuances etc.

Yeah for sure, it would be technology specific.  So ssb would end up with a hash, in NPR for example you would end up with a url including a database key that was also created on insert... etc.  But in both of those cases, you don't know until it gets created.

fosterlynn:
> if client just generated a new UUID or CUID based IRI.

So in cases where the client generates a unique ID, it makes sense that apps using that technology would have to do it differently.  So far, I'm seeing that we can't standardize that kind of thing.  

For example the json created for this PR won't work for the graphql api I'm doing for NRP/OCP, because the incoming data is flat (just a list of parameters), so I did something like this, very flattened out: 
```
class CreateEconomicEvent(AuthedMutation):
    class Input(with_metaclass(AuthedInputMeta)):
        action = graphene.String(required=False) #but then must have a commitment
        input_of_id = graphene.Int(required=False)
        output_of_id = graphene.Int(required=False)
        provider_id = graphene.Int(required=False)
        receiver_id = graphene.Int(required=False)
        scope_id = graphene.Int(required=False) #but then must have a commitment
        affects_id = graphene.Int(required=False)
        affected_resource_classified_as_id = graphene.Int(required=False) #but then must have a commitment
        affected_numeric_value = graphene.String(required=True)
        affected_unit_id = graphene.Int(required=False) #but then must have a commitment
        start = graphene.String(required=False)
        fulfills_commitment_id = graphene.Int(required=False)
        url = graphene.String(required=False)
        note = graphene.String(required=False)
        create_resource = graphene.Boolean(required=False)
        resource_tracking_identifier = graphene.String(required=False)
        resource_image = graphene.String(required=False)
        resource_note = graphene.String(required=False)
        resource_current_location_id = graphene.Int(required=False)
```

So things end up looking different for different technologies, but still are using the vocabulary.

Thoughts?

elf-pavlik:
I hope we can at least agree to identify everything with IRIs
* `https:`
* `did:` - https://w3c-ccg.github.io/did-spec/
* `ni:` - https://tools.ietf.org/html/rfc6920 (Naming Things with Hashes)

APIs and protocols will differ to various extents but at least we should have possibility to use IRIs everywhere. Of course message with an event can't include its own hash, but it won't hurt if it has some GUID based IRI.

fosterlynn:
>I hope we can at least agree to identify everything with IRIs

I have no idea if that will work for all the technologies that work off the web.  But the links are interesting.  I don't think I will be technical enough, maybe @elf-pavlik you can comment over in ssb when we get there, I'll ping...

>so I did something like this, very flattened out

But to be clear, the queries are straight-up VF.

Issue #266, opened 2018-01-07
Agreement Template


elf-pavlik:
Capturing quickly from: https://github.com/valueflows/valueflows/issues/261#issuecomment-355834405

> I think reusable Agreement Recipe would cover cases like: https://en.wikipedia.org/wiki/List_price

While process recipes rely on physical dynamics, agreement recipes rely on social dynamics. This may imply that without implications of physical constraints, they may not come as 'reusable'. Example of *List price* seems to give at least one case for reusing such *Agreement Recipe* by many agents.

fosterlynn:
>Could you maybe share in #266 the use case with '3 transfer types' from Sensorica?

Exchange Type: Purchase
    Transfer Type: Receipt (of material items)
    Transfer Type: Expense (other costs of getting the material items)
    Transfer Type: Payment

(Note the terms Purchase, Receipt, Expense, and Payment are user defined.)

elf-pavlik:
could you please also write quickly all events for a observed exchange based on that recipe?

fosterlynn:
>could you please also write quickly all events for a observed exchange based on that recipe?

![exch-vf-ex](https://user-images.githubusercontent.com/3776081/34654356-3fef9d0e-f3c0-11e7-9d82-c21f73c0cbe6.png)

Somewhat confusing, not sure why they have both Expense and Additional Expense in the exchange type.

elf-pavlik:
> could you please also write quickly all events for a observed exchange based on that recipe? Including provider & recipient relations for those events.

On your screenshot I have impression that you have 3 agents:
* ORD solutions
* Digital Fab services
* Tibi

Does this record 3 way exchage betwen those 3 agents or 2 way and Tibi represents Digital Fab here?

Those agents mostly appear on Recipt and two Pay Purchase, and expense has only *receive* so no info who *give* there.

I asked for events with *provider* and *recipient* information since most likely that provides all the information to infer all the reciprocity details in the exchange under this agreement.

[...]

PR #275, opened 2018-04-07
Transfer

fosterlynn:
Please don't merge yet, this is to get some review and see if this is on the right track for everyone.  Once we have agreed on the details, I'll also do the UML diagram and a new webvowl diagram.


fosterlynn:
Almost forgot, here is a visual that might help.
transfer-webvowl

gcassel:
I hope Elf and Bob review this, but FYI it looks solid to me.   I think *transfer* has been a key element to get right in a fundamentally sound and flexible way, and it's looking good.

docs/introduction/exchanges.md
```diff
@@ -1,10 +1,46 @@
...
 This differs from (for example) the [Good Relations Conceptual Model](http://wiki.goodrelations-vocabulary.org/Documentation/Conceptual_model), which we like and want to intertrade with. But Good Relations assumes Compensation in the form of money; the compensation itself is not a separate promise; and the actual flow events are not part of the model. (That's not a criticism. Good Relations has different goals, and a more minimal model makes sense for them.)
 
 In ValueFlows, we want to track not only the offers and promises, but also the actual flows of resources in networks, in all directions. And we support exchanges that don't involve money as well as those that do.
 
-We also support non-reciprocal one-way transfers, such as in a gift economy.
+We also support non-reciprocal one-way transfers, such as in a gift economy.  An exchange implies at least two transfers.
+
+### Transfers
+
+We want to emphasize that although ownership is the primary right transferred in the current economic system (along with renting or leasing), we seek to focus on transfers not necessarily governed by the legal context of property ownership.  This puts the focus on economic coordination.
```
elf-pavlik: If Transfer supposed to handle transferring various kinds of rights and responsibilities, where do you specify what kind of them gets transferred?

docs/introduction/exchanges.md
```diff
@@ -1,10 +1,46 @@
...
+
+For example, perhaps someone has many apple trees, and plans on pressing apple cider. Another agent has an apple press and agrees to exchange use of the press (a resource) for a portion of the apple cider.  The use of the press involves some rights (to use the press for some period of time) and responsibilities (to not run it beyond its capacity and to clean it up before returning it), and the portion of apple cider is transferred in exchange.
```
elf-pavlik: I think we should add **Examples** section to the gitbook with YAML snippets which we can link to from other parts of the book. Could you imagine writing a YAML for this use case first and add *see data example* link here?

docs/introduction/exchanges.md
```diff
@@ -1,10 +1,46 @@
...
+
+Or in a library, a book can be checked out for a period of time.  The agent who checks it out is responsible for caring for it and returning it on time, and can read it, or use it however they want.
+
+Or let's say that a community has farmland and equipment held in common.  The community transfers that to some farmers to use and take care of.  The community also transfers seeds every year to the farmers, enough to grow the food the community needs.  During the year, the harvests are distributed (transferred) to the community members for their consumption.  In reciprocity, the community provides for other needs of the farmers.
```
elf-pavlik: This one sounds like diving into some more elaborate *access control* / *capabilities*. So far it seems like we had an agent which owns something and this agent can make any kind of *Commitment* on this thing - *use*, *consume* etc.

I this use case, what kinds of *Commitments* this farmer can make on that equipment. Once again what kinds of rights and responsiblities community *transfers* to this farmer?

bhaugen:
I think at least @fosterlynn and I have always considered Custody rather than ownership as one of the options for transfer, at least coming from the experience of Sensorica, which uses that concept, but previously in my experience with vendor-managed inventory.

The rights and responsibilities of custody are fairly well understood in common law. The responsibilities are can range from [fiduciary](https://en.wikipedia.org/wiki/Fiduciary) to [duty of care](https://en.wikipedia.org/wiki/Duty_of_care). 

But I don't think we need in ValueFlows to spell any of that stuff out. Just provide some hooks in the vocabulary where user groups can spell what they need.

bhaugen:
Re the farmer and the CSA: here is a handbook that explains lots of configurations:
https://urgenci.net/csa4europe/european-handbook-on-csa/

fosterlynn:
>But I don't think we need in ValueFlows to spell any of that stuff out. Just provide some hooks in the vocabulary where user groups can spell what they need.

I will make that more clear - I think all of that is in the agreements, which we reference but do not support in terms of vocabulary of the agreement (which can be quite complex).

docs/introduction/exchanges.md
elf-pavlik: I think we need to have concrete snippets for all three use cases you mention here. I would like to avoid that we claim that certain parts of the vocabulary can handle some use cases but we don't prove it by providing concrete examples.
To stay honest I can't see myself at this moment myself how the new terms this PR proposes supposed to handle those use cases.

elf-pavlik:
Currently agreements get referenced from events and commitments using *vf:under*. I don't really see how this superposed to work together with proposed here 'transfers of rights and responsibilities'.
I don't say we need to showcase it for all possible cases anyone can come up with, but we should at least show how intended usage for few use cases we consider more common.

Ownership - meaning 'all rights' / 'full access' - in practice agent can make any kind of commitment and stay entitled to any kind of events on that resource.

As I mentioned, 'rights and responsibilities' seem directly related to *access control* / *capabilities* which agents have on given resources. Some may only allow *use*, other ones *consume* as well as *issue* & *receive*. I think we need to consider how it all supposed to work together and back it with concrete examples before we can decide if we should include it in vocabulary.

elf-pavlik:
> I started working on these, and you are right, it is challenging me to blend the transfer concept cleanly into the way we are handling the whole model of exchanges based on events. The same issues are with processes too, I think.

I recall you mostly wanted *Transfer* for commitments and events which use  *produce* & *consume* / *issue* & *receive* action, not really for those which use *work* or *use* action.
I have impression that you mostly don't want to use *Process* just to tie *issue* & *receive* together, or would you also see it of some use for cases where Alice in a Process directly *consume* or *produce* on a resource which Bob controls? I like possibility where Bob makes commitment to Alice that she can *consume* directly from the resource he controls and she can reference it in events that happens as input to some process she runs. But actually between Bob's commitment (which may not reference concrete resource) and the event in a process which Alice runs, we can have a step where Bob transfers rights to certain quantity of certain resource to Alice. This would provide clear **allocation** step since as we discussed commitments don't have to reference particular resource and just based on that Alice may not directly *consume* from resource controlled by Bob. I thought that this kind of **allocation** could work just by commitment referencing particular resource but some scheduling requirement may require that more specific transfer of rights needs to occur before the actual event can happen.


docs/introduction/exchanges.md
fosterlynn: The farmer can use the equipment all year to work the farm.  No commitments are needed as far as I can see.  I agree there could be complexity in the agreement, but I think that is out of scope for VF.  I explain it some of it here because I'm trying to explain we don't have to think about ownership.

fosterlynn:
>would you also see it of some use for cases where Alice in a Process directly consume or produce on a resource which Bob controls?

Yes, also that scenario.  Like I was thinking that there are a couple ways people might want to do the apple press case.

1. Alice transfers the press (loans it) to Bob.
2. Bob uses it to produce cider.
3. Bob transfers the press back, say the next day when Alice picks it up.
4. Bob also transfers 2 gallons of cider to Alice in exchange for use of the press.

or

1. Bob uses Alice's press (which is sitting in their shared garage) to produce cider.  The use event is also an implied transfer.
2. Bob transfers 2 gallons of cider to Alice in exchange for use of the press.

elf-pavlik:
> 1. Alice transfers the press (loans it) to Bob.
> 3. Bob transfers the press back, say the next day when Alice picks it up.

Do you consider it 'creating value'? I think Bob values gaining access to use the press.

In this case Alice & Bob also may want to capture transporting the press with *load* (pick up) and *unload* (drop off) actions in events.

I guess Alice could transfer *use* rights of the press to Bob, and if he needs to take care of picking it up and dropping it off also *load* and *unload* rights. Again before that they could have commitments for all that but for more explicit allocation / scheduling transferring of rights to *use*, *load* and *unload* referencing specific resource and timeframes could come helpful.

fosterlynn:
>Do you consider it 'creating value'? I think Bob values gaining access to use the press.

I think when Bob uses the press, Bob's labor creates value (in the cider, not the press).  When Bob borrows the press, the press itself doesn't become more valuable, and nothing of value is created.  I'm trying to be clear from the perspective of the whole economy, because it affects what people prioritize.  For example, how much effort goes into developing an alternative currency to buy stuff created by capitalist enterprises vs developing cooperative production of goods and services.  Alternative currencies can be very useful, but they can't change the system by themselves.

The transportation thing is interesting.  It seems like an unload (or load, depending on the agreement) event could also imply a transfer (but not be a Transfer).

>I guess Alice could transfer use rights of the press to Bob, and if he needs to take care of picking it up and dropping it off also load and unload rights.

I think it is a mistake to think of rights just in terms of actions.  There are all sorts of complicated parts to rights and responsibilities.  I don't feel inclined to make that sort of data relationship in terms of defining transfer of rights.  Bob can borrow the cider press, then he can make cider, or he can use it for some photos he wants to take.  In either case, he basically just needs to return it in good shape.

elf-pavlik:
> When Bob borrows the press, the press itself doesn't become more valuable, and nothing of value is created. I'm trying to be clear from the perspective of the whole economy, because it affects what people prioritize.

> The transportation thing is interesting. It seems like an unload (or load, depending on the agreement) event could also imply a transfer (but not be a Transfer).

I think Alice transports the press based on Bob's request, she does something that he values (otherwise why he would request it), one could say transporting it 'created value'.

Maybe instead of terms of 'creating value' we could make distinction based on what changes the physical state of resources, and what happens just for the sake of governing resources. So far all the kinds of actions seem to affect physical state of resources, unless one would use *consume* & *produce* just to do 'virtual movements' of how things get managed in the inventory.
I also would like to clarify if 'transfer' always involves more than one agent or agents would also use it just for such 'virtual movements' which may relate to how resources get allocate to processes.
Since we don't seem to intend to use 'transfer' with *work* and *use*, we can't rely on transfers to capture all the reciprocity related aspects. 

> 1. Bob uses Alice's press (which is sitting in their shared garage) to produce cider. The use event is also an implied transfer, so it can be part of an exchange.

I think this version, which doesn't involve transport - no physical change to resource (the press), could come useful. I think we also should consider variants of it where commitments get created before any events can happen. We could also have variants where Alice would *vf:load* pick up her share of the juice (so physical change will happen, possibly some transportation process) as well as where Alice will directly *vf:consume* that juice from resource Bob manages as stock. In the later case, would Alice just need *Commitment* in agreement where Bob commits that she can *vf:consume* on that resource, or Bob would need to 'transfer' her rights to consume that quantity of juice from that resource. Then in agreement would Bob commit directly that she can *vf:consume* and her event would fulfil that commitment. Or Bob would commit to transfer rights to *vf:consume* and event of transferring those rights would fulfil that commitment even that Alice haven't consumed the juice. I really thing we have to consider 'transfering' rights together with Commitments and Agreements.

fosterlynn:
>Maybe instead of terms of 'creating value' we could make distinction based on what changes the physical state of resources, and what happens just for the sake of governing resources. So far all the kinds of actions seem to affect physical state of resources, unless one would use consume & produce just to do 'virtual movements' of how things get managed in the inventory.

This is OK with me, I can clarify in the text and commit.

>I also would like to clarify if 'transfer' always involves more than one agent or agents would also use it just for such 'virtual movements' which may relate to how resources get allocate to processes.

Very good question.  The idea below relates to some of that.  But if you mean 'virtual movements' like the cider press that is owned by Bob, stays in the shared garage, and used by Alice in a process - then I don't see a need for a Transfer.  The use event will be the event involved in the reciprocity, it is an "implied transfer" with no need for extra data.

>Since we don't seem to intend to use 'transfer' with work and use, we can't rely on transfers to capture all the reciprocity related aspects.

:+1:  That is why I keep saying "implied transfer" (little t).  So any type of event can be used in reciprocity.  (I can't think of any that wouldn't have a good example.)  So, we only need something new for events that don't involve transformation or transportation.

So, I've been toying with the various use cases and writing out some data (just for myself thus far).  I want to float an idea for discussion (not yet a proposal), that combines some of this discussion and also trying to minimize duplication of data (transfer/give-event/receive-event) but keep some consistency in the event model for simplicity.

event-transfer

In this model, EconomicEvent is as it is now.  TransferEvent and MoveEvent are subclasses of EconomicEvent.  TransferEvent has an extra resource (one for the give side, one for the receive side), and requires 2 agents.  There is no separate Transfer that operates at the same level as Process.  MoveEvent only needs one agent, and is for the scenario of just moving stuff from one stock resource to another, without involving different agents or implying anything about rights and responsibilities.  And if we like this direction, we can do MoveAgent now or later or not at all, it is a separate need.

This seems to me to be a reasonable compromise between duplicating data and consistency of model.  The reason we need Process is to mediate a many-to-many between events where transformation or transportation is happening - a physical change in the resource(s), possibly complex.  A transfer is simpler, doesn't need a m:m between the "give" and "receive" events.  So if we can get it into one owl:Class, that would be good.

Regarding the issue of who knows what data, it seems to me that we can not worry about it.  (That was one reason I originally wanted a give and a receive event as part of a transfer.)  I'm thinking since if it often the business of just one agent what the identifier of the stock resource is, in a TransferEvent if the giver and receiver both have their data, they can keep internally the resource identification data, and just the independent view data would be served out to the world, maybe just resource classification, tracking identifier if it exists perhaps.  (Hope this is somewhat clear.)

P.S. I know I still need to do some detailed data examples, but want to hear if this direction seems worth considering, especially from @elf-pavlik .

fosterlynn:
>For the cases when we do want to move certain quantity between two different resources. I would prefer to introduce something like StockOperation which would work just like process and have input and output events which would decrement one resource and increment other one.

Yeah, let's leave that one for later, it is really an edge case.

fosterlynn:
>I think we still need to clarify how to include what rights and responsibilities get transferred.

I think that is in the agreement.  It also could be in a resource that references the physical resource as the underlying resource - up to the user I think.  If you think about a rental agreement on an apartment, it is several pages.  I propose we just leave the vf:under as the hook and let other vocabularies define the agreement vocabulary.

But, what would be your proposal?  Maybe an example so I can understand what you are thinking?

bhaugen:
@elf-pavlik @fosterlynn you two have been arguing about transfers for a very long time.

Do you think you are capable of coming to agreement? A lot of the argument seems circular, as often happens in VF, with people restating essentially the same position over and over, maybe with slightly different language.

@elf-pavlik have you implemented anything like transfer in Vientos? If so, how does it work? If not, do you think Vientos will need it?

@fosterlynn needs it in the work she is doing with Ivan M and Luandro, among others.

If you two can come to agreement in the next week, I would say, do it.

Otherwise, I think it would be best if Lynn implements anything she thinks will work and we get some experience with it. But table the argument in VF until we get that experience.

I would say the same for Vientos if you need transfers. Do what you think will work.

Then VF can have some live examples to look at.

elf-pavlik:
> @fosterlynn needs it in the work she is doing with Ivan M and Luandro, among others.

Maybe we could discuss it around a concrete use case from that work?
Does this work assumes single dataset / URL namespace or it already tackles how to do it across dataset and URL namespaces?

> @elf-pavlik have you implemented anything like transfer in Vientos? If so, how does it work? If not, do you think Vientos will need it?

Currently we still have main focus on offers and requests, slowly approaching what looks to me like *Agreement Recipes* #266 

To stay hones I have impression that in different conversations here we tend to use *Transfer* to refer to different things. Once we include in conversation some specific data snippets (YAML or even pseudo data or graph diagrams) I think I will find myself in better position to understand what proposed here *Transfer* would mean and how it would work.

fosterlynn:
>If you two can come to agreement in the next week, I would say, do it.
>Otherwise, I think it would be best if Lynn implements anything she thinks will work and we get some experience with it. But table the argument in VF until we get that experience.

Well I think we slowly make progress.  But yes, sometime soon I'll just have to implement it and treat it as good enough for now.  

>Once again, I don't think we can make everything machine readable, but at the same time I don't think we should assume that nothing can get expressed in machine readable way.

All I'm trying to do is record the observation layer, that a transfer happened.  I think what you're talking about can get added any time later, do you think so @elf-pavlik ?  And I welcome proposals, especially if you think it will impact the base model.  But working with Ivan M is where I'm running into the need to make it as simple as possible.  There won't be any recording of rights and responsibilities in their use case.

I also need to make sure the observation layer transfer can work with Fulfillment and Reciprocity and the model we have talked about for being able to exchange using any event action.  So I still do need to do several concrete examples to make sure.

Before that though, here are some options I can think of to model the base data (naming needs to be worked on):
```
TransferEvent
    @id
    transferClassifiedAs (a TransferClassification)
    under (an ExchangeAgreement)
    plannedDate
    note
    resourceClassifiedAs (a ResourceClassification)
    provider (an Agent)
    receiver (an Agent)
    quantity (a QuantityValue)
    transferDate
    giveResource (an EconomicResource) (optional and only known by the provider)
    receiveResource (an EconomicResource) (optional and only known by the receiver)
```

```
Transfer
    @id
    transferClassifiedAs (a TransferClassification)
    under (an ExchangeAgreement)
    plannedDate
    note
EconomicEvent
    @id
    action: "give"
    resourceClassifiedAs (a ResourceClassification)
    provider (an Agent)
    receiver (an Agent)
    quantity (a QuantityValue)
    start
    affects (an EconomicResource) 
EconomicEvent
    @id
    action: "receive"
    resourceClassifiedAs (a ResourceClassification)
    provider (an Agent)
    receiver (an Agent)
    quantity (a QuantityValue)
    start
    affects (an EconomicResource)
```
```
TransferEvent
    @id
    transferClassifiedAs (a TransferClassification)
    under (an ExchangeAgreement)
    plannedDate
    note
    resourceClassifiedAs (a ResourceClassification)
    provider (an Agent)
    receiver (an Agent)
    quantity (a QuantityValue)
    transferDate
    affects (an EconomicResource - use when serialized or other non-stock resource)
StockFlow (use when stock resources)
    action: "give"
    affects (an EconomicResource)
StockFlow (use when stock resources)
    action: "receive"
    affects (an EconomicResource)
```

fosterlynn:
>Maybe we could discuss it around a concrete use case from that work?

One use case: Work (already a process input) exchanged for payment (in a crypto-currency).  The payments are part of a monthly distribution based on hours worked, with a few other rules about limits agreed upon.

The other use case: Purchase from producer agents of goods and services, and distribution of those between local nodes (organizations).  Possibly distribution to people who are in the community around the local nodes.  Probably payments in the crypto-currency.

And they want to do various stats on all the economic activity.

>Does this work assumes single dataset / URL namespace or it already tackles how to do it across dataset and URL namespaces?

It assumes multiple namespaces, and some existing platform apps slowly moving to more distributed data architecture and smaller apps that are pluggable.  We hasn't tackled any of this yet though, lots to be done there.  The idea of standard vocabulary is just being introduced, and there is some resistance to this.  The first introduction is to start with just EconomicEvent and slowly get some experience.  (And there will need to be something around Agent too for all of this to work.)

So far, we have worked on a new React UI using NRP as a backend.  That is graphql and basically VF, allowing the UI to use any graphql/VF backend.  We haven't tackled any linked open data.

Anyhow, that should give a flavor.

elf-pavlik:
> One use case: Work (already a process input) exchanged for payment (in a crypto-currency). The payments are part of a monthly distribution based on hours worked, with a few other rules about limits agreed upon.

I understand the 'transfer' only relates here to the crypto-currency and not work. Do you want to move crypto-tokens between accounts/addresses (each represent as resource), or you just want to keep all the tokens on the same account / address (resource itself doesn't change) and just transfer rights to some quantity of those tokens to another agent?
If you would like to decrement one resource (account) and increment other resource (account), I would see a place for a Process (non Transformation, non Transporation but some 'stock operation') with two events, one would decrement and other one would increment.


> The other use case: Purchase from producer agents of goods and services, and distribution of those between local nodes (organizations). Possibly distribution to people who are in the community around the local nodes. Probably payments in the crypto-currency.

I think we still don't have much of a clarity on services, I guess you would like to 'transfer' goods and crypto-currency again. For goods, do you want to decrement one resource an increment the other one or you just want to transfer some rights to some quantity of one resource, so resource itself doesn't get affected in any way?

>
```
TransferEvent
    @id
    transferClassifiedAs (a TransferClassification)
    under (an ExchangeAgreement)
    plannedDate
    note
    resourceClassifiedAs (a ResourceClassification)
    provider (an Agent)
    receiver (an Agent)
    quantity (a QuantityValue)
    transferDate
    giveResource (an EconomicResource) (optional and only known by the provider)
    receiveResource (an EconomicResource) (optional and only known by the receiver)
```

* I don't understand why would you use `plannedDate` if we talk about 'observation' - event which already have happened
* If you reference two resources with `giveResource` and `receiveResource`, both optional and somehow only one agent knows which one, how do you determine 'rights and responsibilities' for which resource gets transferred? It really doesn't make sense for me if we don't clearly specify *what* rights for *which* particular resource.

Looking at the two other snippets, I really have impression that you want to decrement one resource and increment another one. It doesn't seem to me like getting into assigning 'rights and responsibilities' over particular resource to various agents. I understand that in no case those resources get Transformed, when it comes to crypto-currency they probably get moved/transfered between two different accounts. Now when it comes to goods, do you see in some cases that you try to call it a transfer while in reality 'unrecorded' transport happens?

fosterlynn:
>Do you want to move crypto-tokens between accounts/addresses (each represent as resource), or you just want to keep all the tokens on the same account / address (resource itself doesn't change) and just certain transfer rights to some quantity of those tokens to another agent? 

Move between accounts/addresses (decrement one, increment another one).  Might be a fee transfer in there too, not sure.  If so I think that is a separate transfer but part of the exchange.  I wonder if we need another way to associate things like fees?  Need to do an example.

>BTW here you can only own them and transfer again, you can not use or do anything else with currency tokens

For sure!  The nature of currencies.

>If you would like to decrement one resource (account) and increment other resource (account), I would see a place for a Process (non Transformation, non Transporation but some 'stock operation') with two events, one would decrement and other one would increment.

That's what I'm trying to do with Transfer.  I just don't want to call it a Process, as I have explained.  Any reason to not just call it Transfer?  And figure out how to support both stock and non-stock transfers cleanly and consistently.  

>I guess you would like to 'transfer' goods and crypto-currency again. For goods, do you want to decrement one resource an increment the other one or you just want to transfer some rights to some quantity of one resource, so resource itself doesn't get affected in any way?

Would probably need to support both stock-like and non-stock-like resources.  I don't really know, but probably need both for different kinds of goods.  And I think safer to have a model that supports both while we are working on it.

>I don't understand why would you use plannedDate if we talk about 'observation' - event which already have happened

Good point.  At some point we will need to plan transfers though.  So that option doesn't work so well for that.  Although probably only one commitment is needed for transfers, because the actual resource is usually not known (only the resource classification).  And it if is known, it is a serialized or otherwise unique resource, like the "dodgy used bike".

>If you reference two resources with giveResource and receiveResource, both optional and somehow only one agent knows which one, how do you determine 'rights and responsibilities' for which resource gets transferred? It really doesn't make sense for me if we don't clearly specify what rights for which particular resource.

Why not?  And how would you like to do that?  (I clearly still don't understand why you think that, even though I understand that is the case.  What are the implications?)

>Now when it comes to goods, do you see in some cases that you try to call it a transfer while in reality 'unrecorded' transport happens?

Don't know.  I suspect people will do that sort of thing, and I don't have any real objection.  People record what is important for what they are doing.  Why do you ask / what implications do you see?

A thought on the answers above: if we do end up with transfers needing to support a m:m (or 1:m) set of give/receive events (like for fees), then it does need a process-like structure and can't be simplified.  That doesn't mean it needs a process-like definition though.  But it would need both complete give and receive events.  I'll do a little investigation on that.

fosterlynn:
I asked about the crypto transactions, and some wallets do it with 2 separate "transactions", some do it with one amount from the sender, which goes to 2 agents, one the receiver and the other the agent/address that gets the fee.   In either case, it looks like they are stored on the same block so at least the blockchain knows they are connected.

So it looks like a transfer could be 1:m at least.  So much for simplification!

>Just for fun, I asked our neighbor who sometimes works in a robot warehouse what events are involved in receiving some inventory and the robot putting it away.

Sure, why not? :)

fosterlynn:
@elf-pavlik a couple thoughts that may (or may not, I don't know) speak to some of your concerns:

Would it help to have Process as superclass, and Transform, Transport, Transfer as subclasses?  That would be OK with me, it would distinguish Transfer enough.

Would it help to define some other actions that imply some high level sets of rights/responsibilities?  Like lend/borrow, rentTo, rentFrom (or something like that), etc.?  I'm fine with that.

elf-pavlik:
> Would it help to have Process as superclass, and Transform, Transport, Transfer as subclasses? That would be OK with me, it would distinguish Transfer enough.

We could try that, which verbs would you use for value of *vf:action*?
I understand an event would still always reference a single resource?

> Would it help to define some other actions that imply some high level sets of rights/responsibilities? Like lend/borrow, rentTo, rentFrom (or something like that), etc.? I'm fine with that.

`rentTo` & `rentFrom` seem like duplication of `receiver` & `provider` which I still need to gives some thoughts via #276 as for *lend* & *borrow* we probably need to work with some examples and see how it works together with *use*, *consume* etc. it does sound to me like some kind of change of possession but not ownership. they would also need some kind of *return* counterpart.

fosterlynn:
>which verbs would you use for value of vf:action?

give and receive - ok for you?  (plus others for different rights if you prefer)

fosterlynn:
>I understand an event would still always reference a single resource?

yes

elf-pavlik:
> give and receive - ok for you?

Could we try very generic *increment* and *decrement*, which seem reasonable for 'stock operation' style processes? Later we could define *produce* as sub property of *increment* and *consume* as sub property of *decrement*. Since we currently have *provider* and *receiver* on events, I think the verb should state what action gets taken on the resource and not have any connotations to how agents may relate to event.

fosterlynn:
>Could we try very generic increment and decrement, which seem reasonable for 'stock operation' style processes? Later we could define produce as sub property of increment and consume as sub property of decrement. Since we currently have provider and receiver on events, I think the verb should state what action gets taken on the resource and not have any connotations to how agents may relate to event.

We already have increment and decrement, with subproperties.

I actually really disagree with a vocabulary talking just to the machine and not to the people, whether devs or users who are working with it.  And yes, we will have users working with the vocabulary.  Most of the actions we have defined have some content that is meaningful, and I think these need to too.  Otherwise, let's all just write assembler and call it a day.

elf-pavlik:
> We already have increment and decrement, with subproperties.

I understand that from computational requirements one could use *increment* and *decrement*. If system you build, would encounter a data using basic *increment* and *decrement*, would it treat it any differently than data using *give* and *recive*?

Do you also want to make *give* a subproperty of *decrement* and *recive* a subproperty of *increment*?

bhaugen:
@elf-pavlik 
> I understand that from computational requirements one could use increment and decrement. If system you build, would encounter a data using basic increment and decrement, would it treat it any differently than data using give and recive?

Pavlik, I thought you were all about semantics. In a transfer, the increment and decrement events are related. If all you have is the resource effects, you need to walk up a relationship or two and then down a relationship or two to see the story. Give and receive tell more of the story more clearly than increment and decrement.

I know that software devs can present them to users any way they want, but we find it also difficult to explain valueflows to software devs. They can use clear semantics too.

fosterlynn:
>I personally like give and take better because it's shorter and has a lot of history

I'm fine with give and take, especially if that helps us get to consensus.  And yes, that is what most of the REA guys use, so maybe a good idea.

>Do you also want to make give a subproperty of decrement and recive a subproperty of increment?

Yes.  Since transfers will involve a resource becoming part of another agent's custody, I think we can assume a different resource identifier, whether stock or not.  (Not saying serialized and lot resources won't be traceable, they of course will, using the trackingIdentifier.)  Seem right to you?

fosterlynn:
A specific note @elf-pavlik 

I got rid of giveOf and receiveOf, and just made it inputOf and outputOf, just like Process.  Couldn't think of a better name.

elf-pavlik:
Not sure if we settled on *take* or *receive* ?

fosterlynn:
>Not sure if we settled on take or receive ?

Bob liked take because it is what the REA folks use and is shorter.  I liked receive because I think it sounds less grabby (more of a push than a pull) and because it is used when receiving resources into a facility.  I left it receive because it didn't sound like a big issue.

But I am completely fine with either.  If either of you @elf-pavlik or @bhaugen wants me to change it, I'll do it happily.

docs/introduction/exchanges.md
```diff
@@ -1,10 +1,50 @@
...
+
+Another example is when a service is created as an output of a process, where that service delivery event can be considered an implied transfer, and exchanged for some other resource.
+
+### Agreements
+
+Various kinds of agreements between agents govern the rights and responsibilities for economic resources that are transferred. Agreements can reflect any economic paradigm, so make it possible for the Exchange and Transfer vocabulary to work equally well for capitalist businesses, transitional economic interactions, and next economy interactions.
+
+Agreements can be of any kind and scope, from an order to a value equation to a much larger blanket agreement.
+
+ValueFlows does not define the vocabulary for agreements, but defines optional references to an agreement when it could be useful.
```
elf-pavlik: Actually we do define `vf:Agreement` and `vf:under`. So far we mostly used agreements to wrap commitments and events into a transaction. We don't have any example of using `vf:Agreement` for 'rights and responsibilities'.

PR #276, opened 2018-04-08
change how flows reference agents

elf-pavlik:
Since we look at more sophisticated mechanism to address rights and responsibilities. I thought this change may clarify how agents relate to flows, in particular Commitment and Event. If diff dosn't make it clear, changes boil down to replacing:
* `vf:provider`
  * `vf:commitedBy` in `vf:Commitment`
  * `vf:performedBy` in `vf:EconomicEvent`
* `vf:receiver`
  * `vf:commitedBy` in `vf:Commitment`
  * `vf:requestedBy` in `vf:EconomicEvents`

I don't think currently terminology made it clear that
* *provider* makes the commitment and then performs the action carried out in event
* *receiver* holds agent who made commitment accountable and requests the action carried out in event

In cases like handling by products, the *receiver* of the by product would probably make the commitment to receive it, not the *provider*.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -265,7 +265,7 @@ vf:appreciationWith
         rdfs:range          vf:EconomicEvent ;
         rdfs:comment        "The economic event implemented in appreciation (gift economy)." .
 
-vf:commitedBy  a     owl:ObjectProperty ;
+vf:committedBy  a     owl:ObjectProperty ;
         rdfs:domain  vf:Commitment ;
         rdfs:label   "commitedBy" ;
         rdfs:range   foaf:Agent ;
@@ -277,7 +277,7 @@ vf:performedBy  a    owl:ObjectProperty ;
         rdfs:range   foaf:Agent ;
         rdfs:comment "The economic agent performs the action." .
 
-vf:commitedTo  a     owl:ObjectProperty ;
+vf:committedTo  a     owl:ObjectProperty ;
         rdfs:domain  vf:Commitment ;
         rdfs:label   "commitedTo" ;
         rdfs:range   foaf:Agent ;
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -40,7 +40,7 @@ vf:ProcessClassification  a           owl:Class ;
 vf:ExchangeClassification  a           owl:Class ;
         rdfs:label           "vf:ExchangeClassification" ;
         rdfs:comment   "An exchange category, model, taxonomy item, tag, facet value, or other kind of classification." .
-        
+
 # PLAN CLASSES
 
 #vf:Intent  a       owl:Class ;
@@ -221,7 +221,7 @@ vf:exchangeClassifiedAs
 vf:involves
         a                   owl:ObjectProperty ;
         rdfs:label         "involves" ;
-        rdfs:domain         vf:Commitment ; 
+        rdfs:domain         vf:Commitment ;
         rdfs:range          vf:EconomicResource ;
         rdfs:comment        "EconomicResource the Commitment intends or commits to affect. Could be an actual resource or a category or type of resource." .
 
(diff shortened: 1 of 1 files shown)


release-doc-in-process/all_vf.TTL
```diff
@@ -265,17 +265,29 @@ vf:appreciationWith
...
+        rdfs:label   "commitedBy" ;
+        rdfs:range   foaf:Agent ;
+        rdfs:comment "The economic agent who makes the commitment." .
+
+vf:performedBy  a    owl:ObjectProperty ;
+        rdfs:domain  vf:EconomicEvent ;
+        rdfs:label   "performedBy" ;
         rdfs:range   foaf:Agent ;
-        rdfs:comment        "The economic agent from whom the intended, committed, or actual economic event is initiated." .
+        rdfs:comment "The economic agent performs the action." .
```
fosterlynn: provides instead of performs? performs seems only like work

release-doc-in-process/all_vf.TTL
```diff
@@ -265,17 +265,29 @@ vf:appreciationWith
...
-        rdfs:domain  [ owl:unionOf (vf:EconomicEvent vf:Commitment) ] ;
-        rdfs:label   "receiver" ;
+vf:commitedTo  a     owl:ObjectProperty ;
+        rdfs:domain  vf:Commitment ;
+        rdfs:label   "commitedTo" ;
         rdfs:range   foaf:Agent ;
-        rdfs:comment        "The economic agent whom the intended, committed, or actual economic event is for." .
+        rdfs:comment "The economic agent who holds the agent who made the commitment accountable." .
+
+vf:requestedBy  a    owl:ObjectProperty ;
```
fosterlynn: receivedBy seems better, not clear there is always a request

fosterlynn: I don't assume accountability, seems putting in place governance not necessarily there, would prefer something more neutral or descriptive.

"The economic agent who plans on receiving the committed action"?

But if you care strongly, I don't care strongly enough to object.

release-doc-in-process/all_vf.TTL
```diff
@@ -265,17 +265,29 @@ vf:appreciationWith
...
-        rdfs:label   "provider" ;
+vf:commitedBy  a     owl:ObjectProperty ;
+        rdfs:domain  vf:Commitment ;
+        rdfs:label   "commitedBy" ;
+        rdfs:range   foaf:Agent ;
+        rdfs:comment "The economic agent who makes the commitment." .
+
+vf:performedBy  a    owl:ObjectProperty ;
+        rdfs:domain  vf:EconomicEvent ;
+        rdfs:label   "performedBy" ;
```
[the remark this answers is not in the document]
elf-pavlik: each event has *vf:action*, one of the available values *vf:cite*, i think this action gets performed or executed or carriet just done. i would like to move away from *provide* since for example for events with `"vf:action": "vf:consume"` having agent who carries that action as *providedBy* seems a little misleading.

release-doc-in-process/all_vf.TTL
elf-pavlik: once again, i try to move away from in some cases misleading *provide* and *receive* wording, so I don't provide consuming apples to someone :wink: 


what other alternative would you suggest to:

* Alice makes commitment to Bob
* Bob holds Alice accountable for that commitment

i think the accountability part makes something for me a commitment and not just intent

elf-pavlik: > You just provide apples. :) You don't perform apples, or perform consuming apples. Let's keep thinkning.....

I do perform / carry out / execute *vf:consume* (referenced from instance of *vf:EconomicEvent* with *vf:action*) on apple resource managed as stock. In that case referencing me from this event with *performedBy* (the *vf:consume* action) seems to me more reasonable than referencing me on that event via *provider* or *receiver*.
If I have harvested some apples and for change peform / carry out / execute *vf:produce* (action) on the same apple resource, I would still use *performedBy* to reference myself from event, rather than *provider* or *receiver*. Particular action *vf:consume* & *vf:produce* already define if the resource will get decremented or incremented, we only need to know who performed / carried out / executed action this event describes. Would you always reference the agent who does *consume* action with *receiver* and the one who does *produce* action with *provider*?

fosterlynn: It is possible that the action verbs are confusing things, because they conceptually blend into what happens in the process, not just the event.  I think "performing" happens more inside the process, as does consuming or producing in some ways.  The 'provider' is just the agent that provides the resource so it can be consumed in the process and since it was consumed, it gets decremented.

We started using provider and receiver (or recipient) because that was becoming somewhat widely used by the REA people.  In NRP, we used fromAgent and toAgent.  Maybe the from and to concepts are still useful?

This ties into the Transfer discussion too, but I'll take that to the other PR.

And @bhaugen says he would like to write up the apple cider process in some detail as a way to look at this more.

PR #286, opened 2018-10-31
Plan

fosterlynn:
First draft, for discussion.  

I'm going to start again being more consistent in moving VF forward, now that more experience has been had in the field, and there are some things needed by multiple projects.  Plan is a relatively simple one (I think!), so I'm putting it out here so people can think and discuss while I work on Intent, which is more complex.

(Note the budget.ttl file got here because I forgot what branch I was on, but I left it because it is just a start to a discussion, and I used it as a link already, and it is just in the "in progress" folder. It will apply to Intent eventually. Sorry.)

(Also note that there are a few more docs that would need updating, like the UML model and table of contents, which I would rather do after there is some agreement on this.)


elf-pavlik:
:wave: glad you keep making progress :+1: 

why this PR includes separate budget.ttl and plan.ttl?

release-doc-in-process/plan.TTL
```diff
@@ -0,0 +1,68 @@
...
+
+vf:name a owl:DatatypeProperty ;
+        rdfs:label      "name" ;
+        rdfs:domain     vf:Plan ;
+        rdfs:comment    "An informal or formal textual identifier for an object. Does not imply uniqueness." .
+
+vf:note a owl:ObjectProperty ;
+        rdfs:label      "note" ;
+        rdfs:range      skos:note ;
+        rdfs:domain     vf:Plan ;
```
elf-pavlik: this property already had definition in https://github.com/valueflows/valueflows/pull/286/files#diff-a5832671e1376b69e3b7f7fb50526009L372

defining it with domain `vf:Plan` would cause that any instance which uses that property will get additional type `vf:Plan` which doesn't seem desired since this property should stay generic and don't cause any inferences.

release-doc-in-process/all_vf.TTL
```diff
@@ -178,6 +182,18 @@ vf:committedOutputs
         rdfs:domain         vf:Process ;
         rdfs:range          vf:Commitment .
 
+vf:plannedFor
+        a                   owl:ObjectProperty ;
+        rdfs:label         "planned for" ;
+        rdfs:range          vf:Plan ;
+        rdfs:domain         [ owl:unionOf (vf:Process vf:Transfer) ] .
```
elf-pavlik: I don't think `vf:Transfer` should reference a *Plan*. I think enough for Commitments to reference it.
BTW do we even have instance of `vf:Transfer` in planning phase? I would think that transfer only gets created for actual events to reference it, in planning phase we probably only have an *Agreement* referenced by Commitments

release-doc-in-process/plan.TTL
fosterlynn: Yes, the plan.ttl file includes some things already in the main ttl.  That is just to be able to create the picture for that text section, nothing else.

fosterlynn:
:wave: good to hear from you @elf-pavlik and with such great detail !

>why this PR includes separate budget.ttl and plan.ttl?

That is just a mistake, I lost track of what branch I was in.  File budget.ttl is not meant for this PR, but it won't do any harm, and it was easier to leave it there.  I'll get it out of this PR if you think that is important.

>I don't think vf:Transfer should reference a Plan. I think enough for Commitments to reference it.

I think vf:Transfer will occupy a similar space as vf:Process in the value flows.  It is there to tie the "give" and "receive" Commitments together.  This structure with a Plan able to include Transfers, allows us to plan for producing and transferring in the same flow.  I think the same structure will work in recipes too.  The transfers and processes will be able to connect through resource flows as usual.  (I don't know why I thought it was going to be complicated to tie exchange and process together in recipes.  If you use the transfer level instead of exchange, it becomes immediately simple and obvious.  And I think Plan is the same, just down a level.)

fosterlynn:
>why this PR includes separate budget.ttl and plan.ttl?

True, they are all part of planning.  I thought I would look at getting vf:Plan pulled by itself, since it is relatively simple, and then tackle Intent, which I think will include budgeting, from the conversation we've been having with @almereyda and colleagues.  If you think they are too interconnected potentially, I can work on them at the same time.

release-doc-in-process/all_vf.TTL
```diff
@@ -47,9 +47,13 @@ vf:TransferClassification  a     owl:Class ;
...
 # PLAN CLASSES
 
+vf:Plan     a           owl:Class ;
+        rdfs:label      "vf:Plan" ;
+        rdfs:comment    "A logical collection of processes and/or transfers that constitute a body of planned work with defined deliverable(s)." .
+
 #vf:Intent  a       owl:Class ;
 #        rdfs:label       "vf:Intent" ;
-#        rdfs:comment   "A published want or offer or plan, sometimes with what is expected or desired in return." .
+#        rdfs:comment   "A request or offer or proposal." .
```
elf-pavlik: I think whatever we will come up in #217 we should call `vf:Proposal` or similar but keep `vf:Intent` in line with `vf:Commitment` and `vf:EconomicEvent` - it specifies action and references a resource or classification.

fosterlynn:
>This unit of the budget would equal to currency, or?  Is currency as subclass of unit?

Oops, I've confused another person by having that budget.ttl file included in this PR by mistake, very sorry!!

But to answer the question as I understand you @y0va is that in VF, when dealing with currency, we tend to think that currencies will be units.  Probably more like data though, rather than subclasses.  There are several repositories of standard units existing.  We have been using qudt.org, but also have had some trouble working with it - and their currency section seems to be under definition.  What currency(ies) do you think you will be using to normalize the budget in the Solawis?  I need to dig deeper into units I think.

fosterlynn:
I removed budget.ttl from this PR.

release-doc-in-process/all_vf.TTL
fosterlynn: I removed this whole definition (it was already commented) so we can include it later without distraction.  I had taken out the word "plan" because it didn't seem correct in that definition.

fosterlynn:
>If you use a Transfer just as something that both FlowEvents give and receive can reference, I don't think it makes sense that Transfer references a plan. Especially that give of one agent and receive of other agent will very likely want to reference distinct plans.

That is possible, but it is also possible for different agents to be involved in defining one plan together.  And a plan often crosses scopes, whether transfers or process flows.

fosterlynn:
>If you use a Transfer just as something that both FlowEvents give and receive can reference, I don't think it makes sense that Transfer references a plan. Especially that give of one agent and receive of other agent will very likely want to reference distinct plans.

Thinking more about this....  It does make sense that one agent might want to include the same transfer in a plan of their own.  Can we solve that in the model by supporting plans containing transfers, but not the opposite reference?  I hate to make a M:M structure for this case of agents that aren't working on the same plan, since they will not see each others' plans I would think, and their planning will be in different namespaces.  I definitely picture that the give event would fall into one agent's plan and the receive event would fall into the other agent's plan, but only through the transfer, which they would share. Does that make any sense in the land of linked-open-data?

elf-pavlik:
> Can we solve that in the model by supporting plans containing transfers, but not the opposite reference?

While every property has a specific direction. One can use it in a statement in any web resource. We can recommend which resource should include which statements but everyone can still choose to put any kind of statement in any web resource. You can see also #123 for clarification that a web resource doesn't have to have a single 'root' (subject) and can include any number of statements about any number of subjects. In JSON-LD speak it would use `@graph` with an array where each object could group statement about each subject. We have no requirement that JSON-LD representation of a resource has to have top level object (single subject). Even if we did, in JSON-LD one can use any property in opposite direction using special `@reverse` construct.
Once again, we can only have *best practice* of which statements should appear in representation of which web resource, but can't really constrain it.

fosterlynn:
>If flows (i, c, e) can reference Plan directly, why Process, Transfer or Agreement would also need that? This way one can still find processes, transfers and agreements relevant to a given plan through the flows which reference both, the plan and any of the other transaction like constructs (process, transfer, agreement)

Continuing with this topic, I talked to Bob about how that happens in "standard" business practice without data exchange.  Even then, agents manage to keep track of what is the same object, like a Transfer.  For example if one agent sends a "purchase order", the other agent will create a "customer order" from that, and reference the purchase order number.  And sometimes the providing agent will send an "advanced shipping notice" to notify the receiving agent a transfer is coming.  All of those kinds of documents will have identifiers, which are recorded by the other agent and can be referenced in communication.  If you buy something on Amazon, they control all the Transfer data for which they are the provider, and you can just see it.  In other words, mostly agents understand already what is the same Transfer.

My current conclusion, based on that discussion, is that flows (i, c, e) won't ever need to reference Plan directly, they can continue to reference Process, or in this discussion Transfer, directly.  In the case of Transfer, either agents operating in different namespaces can map identifiers, or use the first created identifier.  And most often, they can explicitly agree on an identifier when the Transfer is created, especially in closer relationships.  Could be kind of like the trackingIdentifier.  But maybe more like sharedIdentifier?  I can also picture using one IRI source for agreed upon Transfers for closely cooperating partners; or blockchain could facilitate that too.  Maybe we need a property to hold various referenceIdentifiers too, don't know.

Any of this make sense?  Can we think of examples where it won't work?

>vf:Transfer seems to me pretty much like specialization of vf:Agreement but possibly we still have to clarify how those two suppose to differ.

I guess I'd have to understand examples where that seems true.  Usually an agreement that is about transfers involves more than one transfer, some reciprocity.  So there is something that needs agreeing. 

>Just to clarify, we may very likely use FlowIntent in some way for proposals (classified ads #217) but FlowIntent on its own pretty much only comes useful in plans before actual commitments (often under agreements) get made.

Agreed.  (And I removed the section where I used the word 'proposal', it was premature.)

P.S. If you would rather, we can drop Transfer from this PR, and just do Plan for processes?  It would fit in again in Intents or Recipes, both of which are more complex PRs anyway.

elf-pavlik:
> My current conclusion, based on that discussion, is that flows (i, c, e) won't ever need to reference Plan directly, they can continue to reference Process, or in this discussion Transfer, directly.

What about cases where agent plans to produce salsa and in that plan includes a FlowItent to *receive* 300kg of tomatoes. It doesn't make sense to create a Transfer for it since that intent may lead to multiple commitments, possibly each one with different agent which will give/issue some quantity of them, so also resulting in different 'transfers'. In that case having flows (i, c, e) referencing plan directly seems more flexible. Probably it may work similar on the 'other side' if that agent produces that salsa to 'stock up', in the end it will give/issue it to multiple agents but may start with a single FlowIntent to eventually give/issue all that produced salsa.
What advantage do you see in transaction constructs (process, transfer, agreement) referencing plans instead of just having flows (i, c, e) referencing them? Once again:
> One can always find processes, transfers and agreements relevant to a given plan through the flows which reference both, that plan and any of the other transaction like constructs (process, transfer, agreement)

fosterlynn:
>What about cases where agent plans to produce salsa and in that plan includes a FlowItent to receive 300kg of tomatoes. It doesn't make sense to create a Transfer for it since that intent may lead to multiple commitments, possibly each one with different agent which will give/issue some quantity of tomatoes, so also resulting in different 'transfers'.

There can be a lot of looseness between intents and plans in both directions.  For example an agent might want to consolidate all the requirements for consuming tomatoes for the next 2 weeks into one Intent.  (Purchasing systems do this kind of thing.  And can also check for onhand or committed receipts so that can be subtracted.)

But for your example, what do you think would be best?  Would you prefer that for Intents for receiving into a Plan, the intent just references the plan directly (maybe one or more plans)?  It occurs to me that it could be even more complex because suppliers' standard sizes could mean that the tomatoes don't naturally fit into the batch sizes in your plans, so it could easily be a M:M. Maybe with this kind of situation, there is no reason to connect the Intent with any Plan, and no need for Transfers yet?  And if people want to do that, maybe it is a different relationship?  Would have to think what might be useful.  But right now I don't see a need to connect the intents to receive to the plan when the consumes are there already.  Could be missing something, of course.

elf-pavlik:
How about just adding `vf:Plan` and `vf:due` and we already have `vf:scope` which any flow (i, c, e) can use to reference anything, so also a Plan.
Then we can see if we find some limitations of this approach and if introducing new property to relate `vf:Process` (later possibly also `vf:Agreement` and `vf:Transfer`) would overcome those limitations without introducing other ones.
Since implementations will most likely have already functions to find all flows referencing something via `vf:scope` it will also work for `vf:Plan`. Having all the flows, one can easily get all the transaction construct (Process, Agreement, Transfer or whatever else we may come up with), which any of those flows references. This looks to me much more flexible and reusing existing features compared to having additional property which Process would use to reference a Plan.

bhaugen:
The reason we add a link to the plan to processes in the first place was because people in Fair Coop created processes that they wanted to be part of the same plan, but had no flows between them, so it was impossible to collect all the processes for the plan by following the flows. They wanted to log their work and know when the previous process was complete but did not want to record the output-input resources (which were intangible anyway...)

fosterlynn:
>all the transaction construct (Process, Agreement, Transfer or whatever else we may come up with),

Not completely related to this, but I would like to understand why you think of Agreement as a "transaction construct"? @elf-pavlik

elf-pavlik:
Agreement pretty much groups flows together, flows reference it with *vf:under*. In similar way Process groups flows together, flows reference it with *vf:inputOf* or *vf:outputOf* (it seems that we made Transfer a special case of Process for flows with *give* and *receive* actions).

> The reason we added a link to the plan to processes in the first place was because people in Fair Coop created processes that they wanted to be part of the same plan, but had no flows between them, so it was impossible to collect all the processes for the plan by following the flows.

Given all the flows (i, c, e) besides linking to processes and agreements also reference a Plan via *vf:scope* you should have this requirement covered as well.

We had conversations earlier that sometimes we have flows (i, c, e) which do not reference any process, for example some unexpected *work* event that doesn't really go as input to a specific process. Those still may need to get considered as part of the plan. We discussed similar nuance a long time ago before deciding that *vf:scope* will have domain of flow not a process. We also left range open so besides your initial use case to use it with agents people can use it with anything else they need, plans seem like another good fit for using *vf:scope*.

elf-pavlik:
Actually I recall discussions about concept of *Project*. Where you looked at *project* more as an *agent* and me more as a *process*. *Plan* seems a little bit related to *Project*, possibly more in direction of *process* like but more broad and 'having' more specific processes. I think both *Plan* and *Project* supposed to deliver something.
Since *plan* seems to 'have' various specific *processes*, I think to avoid nesting processes into plans and plans into plans and processes into processes. Having the fundamental construct - flows (i, c, e) referencing all the plans, processes, agreements, transactions etc. directly, should help us avoid unpredictable nesting of 'transaction constructs' (which already group certain flows (i, c, e) together).

bhaugen:
@elf-pavlik 
> Actually I recall discussions about concept of _Project_. Where you looked at _project_ more as an _agent_ and me more as a _process_.

This is a small clarification, not very important in this overall discussion, but still should be explained.

This is not about how we look at it, it is about definitions in the wild. We have run into two different concepts named "Project":  the most common one is how you look at it, as having a defined outcome and usually end date. The other one, which we first ran into in Sensorica, and have since run into in other scopes, is an ongoing team focused on something like a problem domain. Some Sensorica examples include scientific instruments, garden sensors, and televised soccer. The project team might perform many different processes with different outcomes.

We don't care which definition people use, but would prefer to use words that don't have two very different definitions in the organizations we know, or if so, to differentiate them, like Economic Event to differentiate from an event you attend, like a party, and Economic Resource to differentiate from a Web resource.

bhaugen:
Actually I think the idea of a project being more like an ongoing organized team than a time-limited one-shot deal might have arisen from open source software projects. Like I think Linux thinks of itself as a project, as do most of the distros (e.g. Debian).

elf-pavlik:
> Plans on the other hand do group flows, that is their purpose in life. As do Transfers and Processes.

Why don't we start with what seams the simplest and most flexible approach to introduce `vf:Plan` and use already existing `vf:scope` from flows to reference it? That seems to cover requirement you've encountered in FairCoop

fosterlynn:
>>* planning

>Planning: Sometimes a generic recipe will cross scope boundaries for particular agents. For example one agent could produce a resource that consumes a component made by another agent. In this case, can the first agent schedule the production of the component by the second agent? Possibly yes, if there are agreements in place for that, and the first agent has verified that inventory does not already exist. Or possibly, based again on agreements, the first agent can assume the second agent will provide the component, with the second agent taking responsibility for checking if it is onhand, and if not, scheduling it for production. Or possibly, the first agent plans only to source the component in some way.

So, the reference to planning in that scoping write-up is to say that scope can affect planning, not that planning creates a scope, although of course in a very broad sense you could say that.  But I think that is on the "too general" side of the sweet spot in conceptual definitions of the vocabulary.  I could be wrong, would be interested in actual use cases that would disprove that.

>As long as you don't have an actual use case for creating processes in a plan that have no flows at all, I don't see problem that UX can fist ask to create plan, later a process and then in the end a flow that captures purpose of that process

No, there is no use case for processes that won't eventually have flows, even though for FairCoop they often only think about the inputs.  But in the meantime, whether 30 seconds or 3 days, the process needs to know its plan (or vice versa).

>Description of your UX suggests to me that application doesn't allow adding 'loose' flows to the plan (eg. 'overhead' work) and forces creating a process even if it would have just one flow referencing it.

Yes, that is currently the case.  It will eventually have at least one input, but not necessarily an output.

fosterlynn:
>Possibly changing vf:scope to vf:inScopeOf would clarify that we focus on scoping flows and not emphasize some scope construct. e.g.

I don't mind changing vf:scope to vf:inScopeOf, but let's look at some examples.  I disagree with not emphasizing scope construct, because that was the original purpose of "context agent" which became "scope".  Other flow-scoping things can also be emphasized, so we can combine our different perspectives and experiences, which is usually useful.  How about "equal emphasis"?  :blush:

Without re-reading all those discussions - what I remember is Agents, geographies or ecosystems (Location? or not), projects that don't consider themselves to have agency as a team (which we never nailed down how to represent).  Others?  

Plan or high-level Process could fit in there, as you suggest.  As @bhaugen has pointed out, that is not a simple assignment or reference, it involves "nesting" processes in a way that the inputs and outputs translate to the higher level.  Like, internal inputs that become outputs internally won't show up as inputs or outputs of the plan or higher-level process.  You could also create an input that crosses all nested processes more indirectly, like cleaning the lab.  But still, I can see where that could fit into the concept of scope.  And I think it would work, since software can see the owl:Class of what is the vf:inScopeOf, and behave accordingly.  So I offer that compromise.

One problem I always run into with open ranges, though, is that it is easy to do that in rdf, but hard to do it in relational or OO implementations, which are of technical necessity more structured.  So in this example, I think that to reference Agents, Locations, Plans, etc. would require a separate property for each one so you can have a FK, or a collection on the other end.  So far, we have left vf:scope loose, reflecting the rdf core, but at some point it would need more definition for non-rdf implementations.

Given that, I would like to try to nail down what is the actual range of vf:scope or vf:inScopeOf.  OK with you?

Funny how what I thought would be a simple PR to get warmed up turned into a much more in-depth discussion.  All good.

elf-pavlik:
> One problem I always run into with open ranges, though, is that it is easy to do that in rdf, but hard to do it in relational or OO implementations, which are of technical necessity more structured. So in this example, I think that to reference Agents, Locations, Plans, etc. would require a separate property for each one so you can have a FK, or a collection on the other end. 

Could you please create a dedicated issue for that, I think this goes way beyond this PR

> I don't mind changing vf:scope to vf:inScopeOf, but let's look at some examples. I disagree with not emphasizing scope construct, because that was the original purpose of "context agent" which became "scope". Other flow-scoping things can also be emphasized, so we can combine our different perspectives and experiences, which is usually useful. How about "equal emphasis"?

Currently you only used instances of `vf:Agent` as objects in statements with `vf:scope`, I recommend just adding `vf:Plan` and also use instances of it as objects in statements with `vf:scope`.

> No, there is no use case for processes that won't eventually have flows, even though for FairCoop they often only think about the inputs. But in the meantime, whether 30 seconds or 3 days, the process needs to know its plan (or vice versa).

If applications has UX which has order plan than process than flows it can keep track on 'current plan' and 'current process' however it wants in the application state. Then we also have a resource state (see: https://ruben.verborgh.org/blog/2012/08/24/rest-wheres-my-state/) do you really need to actually create a process resource (publish in on the web) if it doesn't have any flows? I think it might just belong to application state and only need resource state once it has at least one flow (or even one input and one output for process)

> Yes, that is currently the case. It will eventually have at least one input, but not necessarily an output. 

I would like to understand the use case for process which doesn't have at least one input and one output. I recall using process to keep track on causality between inputs and outputs, if you don't have both inputs and outputs why do you even need that process?

fosterlynn:
>I would like to understand the use case for process which doesn't have at least one input and one output. I recall using process to keep track on causality between inputs and outputs, if you don't have both inputs and outputs why do you even need that process?

There is an output in real life, but it often doesn't get recorded.  And the users are just starting to understand the possibilities of the structure and keeping inventory.  Some of their unrecorded outputs could eventually be useful to have record of, like documents and graphics.  Some of their unrecorded outputs are more like services, which they don't see a need to record, like an assembly for example.  The processes are still useful to contain inputs that require coordination and visibility to work on the same thing.  Like prepare the assembly agenda, announce the assembly, take notes, facilitate, etc.

fosterlynn:
>If applications has UX which has order plan than process than flows it can keep track on 'current plan' and 'current process' however it wants in the application state. Then we also have a resource state (see: https://ruben.verborgh.org/blog/2012/08/24/rest-wheres-my-state/) do you really need to actually create a process resource (publish in on the web) if it doesn't have any flows? I think it might just belong to application state 

That would be possible, but as we get more and more smaller units that need to communicate with each other, I think it is helpful to have a standard vocabulary that covers what they need covered.  In this example, the UI is a React app that is intended to be able to speak with any backend that supports VF.  So it doesn't want to have to know about different app state variables if possible.  

What is the problem with having vf:Process reference a plan or a scope?  Why does it seem wrong to you?

elf-pavlik:
> Like prepare the assembly agenda, announce the assembly, take notes, facilitate, etc.

In this case they could probably use a Plan directly since they just want to group some flows together but don't need to track causality between inputs and outputs.

I guess that the specific application which Fair Coop requires that people using it create processes before creating flows and doesn't support creating flows without a process, but for example still related to specific plan?

> In this example, the UI is a React app that is intended to be able to speak with any backend that supports VF. So it doesn't want to have to know about different app state variables if possible.

Application state shouldn't matter for anyone else since it only exists for this specific application. Why do you need to publish on the web process which has no flows (at least one flow intent)? 

> What is the problem with having vf:Process reference a plan or a scope? Why does it seem wrong to you?

When one wants to query the data, currently one only has to check if flows reference something with `vf:scope`. In case where anyone can reference something used for scoping flows **either** from flows of from Process etc. now one needs to have queries that check both (or more), references from flows and references from processes and other possible references.
For the sake of simple application (closed world) it doesn't matter since people developing it have full control over it and can make assumptions based valid for their specific implementation. On the other hand in an ecosystem of applications and services (open world) one can't make such assumptions any more so in the end will have to check data for any possible way of creating relations.

fosterlynn:
>>In this example, the UI is a React app that is intended to be able to speak with any backend that supports VF. So it doesn't want to have to know about different app state variables if possible.

>Application state shouldn't matter for anyone else since it only exists for this specific application. Why do you need to publish on the web process which has no flows (at least one flow intent)?

It is not published on the web, it uses graphql.  But still it wants a standard vocabulary to communicate, otherwise you can't decouple the UI from the backend as separate apps.  In the sense of Open App Ecosystem.  This is just one of the many ways apps might need to communicate besides from node to node over the open web.  I don't want to assume an application is full stack and everything else is internal.  For example, I've seen a need for a stable generic VF backend and multiple UI's tailored to domain, different use cases within domain, all kinds of specific needs.  There is a UI app in FairCoop that does only one kind of income distribution tailored completely to only one group within FairCoop's specific requirements, for example.

fosterlynn:
>>Like prepare the assembly agenda, announce the assembly, take notes, facilitate, etc.

>In this case they could probably use a Plan directly since they just want to group some flows together but don't need to track causality between inputs and outputs.

But what if they want a plan that does more than an assembly, maybe it contains all the work done by a facilitator during the month.  So that example is a process within that plan, which contains other processes, and even process flows where it makes sense to them.  Just to say, actual user use in the field is sometimes "messy".

elf-pavlik:
> It is not published on the web, it uses graphql. But still it wants a standard vocabulary to communicate, otherwise you can't decouple the UI from the backend as separate apps.

Why do you need to store on the server a process which has no flows? I understand that you may want to create it in the app and keep it in application state to have a reference you can use in the flows. Maybe in your implementation only server creates IRIs so you need to make a request first to have IRI you can then use as a reference? I often use UUID based IRIs which client can create based on URI template. This way clients can create IRIs even while they stay offline.

Even if you only let server create IRIs, client can create plan, create process (without linking it to the plan) and then create flow(s) which reference both. So process doesn't really have to reference plan directly. Until flows get create the relation between process and plan only exists in that application (client) and it doesn't require any interoperable way of doing it.

fosterlynn:
>When one wants to query the data, currently one only has to check if flows reference something with vf:scope. In case where anyone can reference something used for scoping flows either from flows of from Process etc. now one needs to have queries that check both (or more), references from flows and references from processes and other possible references.

Can't a query just check what it is querying?  Like if it is querying processes, it gets processes for the scope; if it is querying events, it gets events for the scope.  A process query would be a lot simpler because the queries that get processes don't have to check events, and can assume that the events related to a process have the process's scope(s).  It doesn't seem like it is the job of a query to validate data.  I don't think it is even valid for events related to a process to have different scopes, can you think of an example where that would be true?  Transfers are different because they often have more than one scope because of the nature of a transfer.  (This applies to leaf-level processes.)

If we end up with nested processes however, such as plans, then non-leaf-level processes could include more than one scope.  So that would need to be part of that model.  I think in that case, the process would not reference any scope that is for only some of its nested processes.  Or maybe it is a reason to have those things be something other than processes, would need to think about the trade-offs.  Anyhow, I'll do an experiment for nested processes.

fosterlynn:
>Why do you need to store on the server a process which has no flows?

Because the server is the only place any data is stored.  Clients have no persistent data.  There is no LOD needed yet in this situation.  Soon, we hope!  But not clear if any of that will come from the client....

fosterlynn:
>Could you please share a use case where you need to store on a server a process with not even a single flow referencing it but where this process needs to reference a plan? I would like to understand purpose of creating a process without any inputs or outputs.

I am worried we are going in circles, but I will try again.
process-no-commit

Warning, test data!  I just brought up this screen and that last process has no commitments yet.  But the only way I can see it exists is because it belongs to a plan called Macao Hackaton which I clicked on. And the process/plan data persists on the backend app, which has a VF api (yay!).  There is no other way to find it in this UI.  Since I can see it here, I can now add commitments and so can other people.  This is not about static pure model in the abstract, this is about real world use of a model.  Just because it doesn't have commitments or events *yet* doesn't mean you shouldn't be able to see it in this UX.  The use case is creating a plan which consists of multiple processes and which is not created from a recipe.  During creating of this plan, processes will have 0 to N commitments.  They always start with 0.  When the plan is done, none of them have 0.

elf-pavlik:
Thank you @fosterlynn for sharing that example, I think it can help with keeping our conversation focused around your use case!

Quick observation: if you use a Plan to represent a board and Process to represent a list, just based on some property relating those two one can't keep track on the order. I guess that lists get ordered on the board based on dates (eg. startDate) but one has no way to reorder them in some other arbitrary way for example by dragging them left right.

Quick question: when someone creates *work* commitment the person automatically gets assigned as *provider* or you allow people assigning other people so when someone creates *work* commitment it has required select field to choose assigned person? By the way a `vf:Commitment` differs from `vf:Intent`, in cases where someone doesn't actually commits to *provide* right away, one should start with instance of `vf:Intent` (Flow Intent) which hopefully will lead to one or more instances of `vf:Commitment` (Flow Commitment) which in turn will hopefully lead to one or more instances of `vf:EconomicEvent` (Flow Event).

Quick question: does this application allow adding flows (intents, commitments, events) to the plan, where those flows don't appear on input or output of some process?

Now a general thought. I don't expect that people will build applications relying only on terms defined in VF ontology. As we keep the core VF very minimal, to build more complex things often focusing on some specific domain, people will need to use other web ontologies to fully represent their data.
For example one could use `kanban:Board` and `kanban:List` with other constructs like ordered list to structure their data. In similar way one aiming to gamify things could use `rpg:Campaign` and `rpg:Mission` with various relations between them. Having range of `vf:inScopeOf` open allows everyone to use whatever their can imagine to group flows around it. I would expect that people would also use schema.org vocab and use calendar events to scope flows around them. Everyone also has freedom to make statements like
```jsonld
{
  "@type": ["vf:Process", "schema:Event"]
}
```
This way one can use some generic calendaring app to manage calendar events, but also use those events in VF app to manage process aspect of them with input/output flows.

Using terms from VF ontology should have some clear implication and enable strictly defined reasoning over the data which includes classes and properties we define. I don't think we should include terms which don't have such clearly defined implications and enable some specific reasoning.

I also would like us to keep in mind how choices we make in ontology design will impact querying multiple sources of data over the network ([Latency Numbers Every Programmer Should Know](https://gist.github.com/jboner/2841832)). As well as distributing workload of query execution between clients an servers, where http://linkeddatafragments.org/ does a very systematic work on exploring available spectrum.

Coming back to the implications of using VF terms and reasoning it enables. In case of `vf:Process` it allows to reason about causality between inputs and outputs - specific set of inputs lead to specific set of outputs. If those flows reference specific resources or even just classifications (taxonomy references), it allows further reasoning about IPO chains.

> Can't a query just check what it is querying? Like if it is querying processes, it gets processes for the scope; if it is querying events, it gets events for the scope.

If we take a case where someone uses an instance 
```jsonld
{
  "@context": {
    "schema": "https://schema.org/"
   },
  "@id": "https://mycalendar.example/lynns-bday-party",
  "@type": "schema:Event"
}
```
to scope flows, and wants to query dataset for all the flows scoped by that event. When only flows use `vf:inScopeOf`, query based on simple pattern `?flow vf:inScopeOf <https://mycalendar.example/lynns-bday-party>` will return all the flows of interest. Now if both processes and flows can link to that event with `vf:inScopeOf` one can start with very similar query `?flowOrProcess vf:inScopeOf <https://mycalendar.example/lynns-bday-party>` but now for each of the `?flowOrProcess` results, one has to retrieve them check their type and for each instance of `vf:Process` query for flows that reference it either with `vf:inputOf` or `vf:outputOf`. Of course details will vary based on what kind of query interfaces given dataset will allow. Since client can not make assumption if in some dataset only flows reference that event or processes as well, now each client always have to anticipate the more complex case, which requires checking type of each result of first query and for `vf:Process` issue two more queries to get both all the inputs and all the outputs.

fosterlynn:
@elf-pavlik thanks for the thoughts!  A few quick answers first, and am thinking about your general comments.

>I guess that lists get ordered on the board based on dates (eg. startDate) but one has no way to reorder them in some other arbitrary way for example by dragging them left right.

It orders based on resource flow, which requires outputs.  If there aren't outputs, I think it does order by date.

>Quick question: when someone creates work commitment the person automatically gets assigned as provider or you allow people assigning other people so when someone creates work commitment it has required select field to choose assigned person?

The backend doesn't support Intents right now (forked from NRP).  So it creates a Commitment without a provider (which we understand is actually an Intent of course) as part of the plan.  As soon as we get Intents nailed down in VF, I can add it to the api even without that structure in the backend, but need some decisions here.  People can't assign other people, people can only commit themselves to the intent.

>Quick question: does this application allow adding flows (intents, commitments, events) to the plan, where those flows don't appear on input or output of some process?

No.  Has been some discussion to maybe do so, but no clear requirements and no decisions.

elf-pavlik:
>  A few quick answers first, and am thinking about your general comments.

Thanks, I think those general topics will keep coming back, especially once we work on more implementations.

>> Quick question: does this application allow adding flows (intents, commitments, events) to the plan, where those flows don't appear on input or output of some process?

> No. Has been some discussion to maybe do so, but no clear requirements and no decisions.

> Warning, test data!

Actually *Sdas* process doesn't come that helpful. Possibly in real scenario one would start with an instance of `vf:Intent` and only create an instance of `vf:Process` when need of capturing causality between inputs and outputs arises. Also having possibility to use anything for scoping flows around it, one would create something else than `vf:Process` unless one really has to track causality between inputs and outputs. 

> @elf-pavlik do you think it is worth pursuing the idea of nested processes or not really? I think I'd like to give it a try.

Sounds worth finding use cases that would need it. I would also keep in mind what we have explored so far with using `vf:service` action for flows that connect two processes directly (so no Process - Flow - Resource - Flow - Process but Process - Flow - Process)

Looking at your kanban board application I would see it important to see what does matter for ValueFlows and what doesn't. If it tries to capture something that doesn't directly affect value flows related reasoning, it can use any other vocab for that. We should clearly define what one can reason when
* flow references a plan with `vf:inScopeOf`

If we also recommend referencing plan from processes
* process references a plan with `vf:inScopeOf`
* how does it differ from flows referencing plan directly

Then if we introduce some kind of hierarchies between processes and possibly between plans
* what does it imply when flow references something with `vf:inScopeOf` with respect to how this flow relates to other entities in this hierarchy

Some things we could even consider formalizing with [OWL Axioms](https://www.w3.org/TR/2012/REC-owl2-quick-reference-20121211/#Axioms) but then we either require reasoners or mandate to always [Materialize Inferences](http://patterns.dataincubator.org/book/materialize-inferences.html)

fosterlynn:
>Possibly in real scenario one would start with an instance of vf:Intent and only create an instance of vf:Process when need of capturing causality between inputs and outputs arises.

That is certainly one scenario.  But the one I documented is a quite real scenario, being used in production by several groups.  Intents would be created just like what you see, after creation of the process.  It is all part of creating a plan when you don't have a recipe.  The Intents on the plan could then be published to see if people want to commit.  People will want to think about plan creation differently, and we need to support how they do.  Personally, I would find it harder to think about individual input intents without having first thought through the overview of the process/resource flow.

>unless one really has to track causality between inputs and outputs.

Mostly they do.  It is a resource flow kanban. This is a better example:
process-flow

elf-pavlik:
> Mostly they do. It is a resource flow kanban. This is a better example from production

In that example starting with *produce revision | proofread | translated post* intent seems very reasonable. Those seem a point / purpose / goal of running that process. 

Anyways, I just thought of another possibility. We could actually make domain of `vf:inScopeOf` open as well, this way anything could stay in scope of anything. At the same time we would not define `vf:inScopeOf` as [transitive property](https://www.w3.org/TR/2012/REC-owl2-primer-20121211/#a_TransitiveObjectProperty). This way having flow in scope of process and process in scope of a plan would not imply that this flow also stays in scope of this plan. One would have to make an additional explicit statement that this flow stays in scope of this plan as well. In addition plans could get scoped by other plans, processes scoped by other processes etc.
Given above lack of transitivity, one could still only query for flows that stay in scope of the plan since having process in scope of a plan would not imply that flows coming as that process' inputs and outputs also stay in scope of that plan.

elf-pavlik:
> having flow in scope of process

defining `vf:inputOf` and `vf:outputOf` as sub properties of `vf:inScopeOf` would imply that all the inputs and outputs get scope by the process

fosterlynn:
>We could actually make domain of vf:inScopeOf open as well, this way anything could stay in scope of anything. 

That would be interesting.  What is your experience in the LOD world when people leave range and/or domain open?  To people do crazy things that don't make any sense?  Or do people stick to what makes sense, even though the vocab says they are allowed to do crazy things?

elf-pavlik:
If you look at https://www.w3.org/TR/rdf-schema/ you can see that it doesn't put any restrictions. It only causes reasoner to infer additional statements from existing statements which include RDFS constructs.
To provide something which puts restrictions on data and enables automatic validation of data we can take a look at Data Shapes
* Shapes Constraint Language (SHACL) https://www.w3.org/TR/shacl/
* ShEx - Shape Expressions https://shex.io/

With very generic term like `vf:inScopeOf` I don't see case where someone would use it in 'crazy` way. Stating that something stays in scope of itself doesn't seem helpful in anything. We should still start some kind of *common pitfalls* section and possibly warn about creating circular scopes which could cause problem with recursive querying.

We definitely should also explain that one MUST NOT consider `vf:inScopeOf` as transitive. Exemplifying it with a process scoped by a plan NOT implying that input & output flows of that process also get scoped by that plan. To have both process and its input scoped by the plan one has to have explicit `vf:inScopeOf` statements for the process and each flow.

Particular implementation can choose to automatically add `vf:inScopeOf` to newly created flows if they reference a process which already has `vf:inScopeOf` statements and hide that detail in UX.
This way to get all the flows scoped by the plan (eg `<:specific-plan>`) one can simply query for pattern `?flow vf:inScopeOf <:specific-plan>` and not need to check any scoping hierarchy (via processes etc.) since `vf:inScopeOf` doesn't work as transitive property and requires explicit direct scoping.

elf-pavlik:
At current stage of conversation I would look for those changes to this PR:
* [ ] remove `plan.TTL` which sooner or later will start getting out of sync with `all_vf.TTL`
* [ ] remove proposed `vf:plannedFor`
* [ ] remove proposed `vf:deliverableFor`
* [ ] change domain of `vf:scope` to `owl:Thing`

I can make separate PR to rename `vf:scope` to `vf:inScopeOf` and clarify that one MUST NOT consider it as transitive property.

fosterlynn:
>At current stage of conversation I would look for those changes to this PR:

Thanks for the summary!  I'm actually still toying with removing vf:Plan and just using vf:Process for that and other cases where process flows are nested into a higher-level process.  Or creating a new class for the higher-level processes, which might be plans, functions, the thing they want to use for budgeting, or whatever else comes up.  Started doing some ttl but didn't get far enough to have anything coherent.  Thought I might do an alternative PR for this view.

And thinking about the scope thing, although leaning towards it.

fosterlynn:
>remove plan.TTL which sooner or later will start getting out of sync with all_vf.TTL

A note on all the random ttl's that we have - they are just for the webvowl pics in each text section.  And yes, some of them do need updating.  But, they aren't part of any public deliverable.  Maybe we should move them to a different subdirectory?  I think it is nice to have the small webvowl pictures, that big one is not comprehensible.

fosterlynn:
>* change domain of vf:scope to owl:Thing

>I can make separate PR to rename vf:scope to vf:inScopeOf and clarify that one MUST NOT consider it as transitive property.

You could make this PR if you like, and I'll use it for the Process idea, and we can try it out also for working on Intent, Recipe, etc.

I think I would then want to put several specific uses of vf:inScopeOf into the UML diagram so it makes sense in that paradigm.  I can do that after your PR, or as part of your PR.

P.S. I'll be traveling Thursday - Tuesday, with some time online but not a lot.  And be fairly busy tomorrow.

elf-pavlik:
would you like to close this one in favor of #293 ?

fosterlynn:
Closing, will use nested processes instead.

PR #288, opened 2018-11-06
deconflating intent and proposal

elf-pavlik:
In this PR I try to de-conflate `vf:Intent` (earliest phase of a flow) from `vf:Proposal`. It will need more iterating over it but I think we need to take this first step already. #217 should yield initial definition of `vf:Proposal`. Here I'll just add another commit which defines `vf:Intent` as flow similar to how we define `vf:Commitment` and `vf:EconomicEvent`.


fosterlynn:
Couple of typos: "borad", "boradest" in the proposals file.

Also, I would like to not talk about "request for proposal' until we think it is applicable to what we are doing and understand how it fits in.

In general, I'm otherwise OK with your PR, pending what we figure out about Intents.  @bhaugen had an idea that proposals might be a 3rd type of Intent (after offer and request).  But that might be awhile, so if you want to separate them in the meantime, it is OK with me.

bhaugen:
> @bhaugen had an idea that proposals might be a 3rd type of Intent (after offer and request). But that might be awhile, so if you want to separate them in the meantime, it is OK with me.

This PR is ok with me, assuming I read this part correctly:
> Proposals refer to published requests or offers, sometimes with what is expected in return. Agreements can evolve from proposals, or be entered into without proposals. They are specifically committed to by agents.

I read that to mean that proposals could refer to requests or offers. In other words, requests and offers are not the same as proposals, but proposals could refer to them, and start a conversation for action where a request could be fulfilled or an offer accepted and completed. 

@elf-pavlik - is that what you meant? If so, +1

elf-pavlik:
This PR only focuses on clearly definition of `vf:Intent` as flow in planning phase just like `vf:Commitment` but before involved parties (provider & receiver) actually commit (agree).

#217 should lead us to sorting out how to model Proposal, in many cases it will need to provide a way to capture expected reciprocity.

elf-pavlik:
This PR pretty much addresses issue discussed here

https://github.com/valueflows/valueflows/pull/286#issuecomment-435901416

>> Quick question: when someone creates work commitment the person automatically gets assigned as provider or you allow people assigning other people so when someone creates work commitment it has required select field to choose assigned person?

> The backend doesn't support Intents right now (forked from NRP). So it creates a Commitment without a provider (which we understand is actually an Intent of course) as part of the plan. As soon as we get Intents nailed down in VF, I can add it to the api even without that structure in the backend, but need some decisions here. People can't assign other people, people can only commit themselves to the intent.

Issue #152, opened 2016-08-22
Proposal for representing units of time and their relationships


bhaugen:
This issue came up when discussing time-based resources, but I am putting it here in the main VF repo because it may have broader implications.

I suggest to represent units of time and their relationships, we use [owl-time](https://www.w3.org/TR/owl-time/) instead of [qudt time units](http://qudt.org/vocab/unit/units-space-and-time.html), although we are using qudt for physical units.

Qudt focuses on scientific uses of time, for astronomy etc, with a lot of emphasis on velocities.

Owl-time is based on temporal relations, which is what we will need. As they say on the owl-time doc, "The basic structure of the ontology is based on an algebra of binary relations on intervals (e.g., meets, overlaps) developed by Allen and Ferguson for representing qualitative temporal information, and to address the problem of reasoning about such information."

Allen and Ferguson is the source for every good temporal relations library I have ever seen.

To represent availability of time-based resources, I think we will want to use [time:DateTimeInterval](https://www.w3.org/TR/owl-time/#time:DateTimeInterval)~~, which has a beginning and ending [time:DateTimeDescription](https://www.w3.org/TR/owl-time/#time:DateTimeDescription) and a [time:Duration](https://www.w3.org/TR/owl-time/#time:Duration)~~. 
_[edit] See @dr-shorthair 's comment below and this diagram:_
![owl-time classes](https://www.w3.org/TR/owl-time/images/TemporalDuration.png)

_TemporalUnit is based on TemporalDuration, so my statement that I crossed out was incorrect._

The availability of time-based resources can be represented by a calendar, a collection of vf:Commitments where the unit is a time:DateTimeInterval. The intervals can have any required granularity (days, hours, minutes, whatever) and can have gaps in between them where the time-based resource is not available.

In the recently cited use cases where some equipment is being repaired, the interval of the expected repair would be removed from the availability calendar.

bhaugen:
> In the recently cited use cases where some equipment is being repaired, the interval of the expected repair would be removed from the availability calendar.

One way to "remove" those intervals semi-automatically would be for the repair process to reserve them. Then they would no longer be available for other uses, plus other prospective users could see what is happening on the resource calendar.

bhaugen:
Example of a resource calendar: https://help.xmatters.com/ondemand/groups/groups-calendar.htm

bhaugen:
See also https://plus.google.com/u/0/+SimonCox1959/posts/XiFNP59VU1m
[...]

PR #291, opened 2018-11-16
flow granularity

fosterlynn:
This is a proposal to document that intents, commitments, and economic events can be used in VF for any level of granularity.  This means that for analytical or higher level planning, budgeting, etc., we can use the same flow classes as for operational needs.

If using as higher level flows, it could be possible that there are lower level flows that roll up into the higher level flows.  I'm not including any data structures to express that, thinking that can be inferred from the situations and if needed, from the dates, resource classifications, etc.

To make this happen, I changed the date/time related properties to use time:TemporalEntity, which is subclassed into Instant and Interval, per this spec https://www.w3.org/TR/owl-time/#time:Duration.  This was originally suggested by @bhaugen in this issue: https://github.com/valueflows/valueflows/issues/152.  The dates/durations of events and commitments needed to be cleaned up anyhow, vf:start was not working in practice.  I assume that specifying the superclass infers that people can use any subclass - please confirm @elf-pavlik .


fosterlynn:
Note: This is the picture that should appear in the flows.md file, hopefully it does.
i-c-e

fosterlynn:
As a general note to think about, I'm starting to think it is worth creating a superclass of I-C-E and a superclass of Process-Transfer.  Something like EconomicInteraction and EconomicFlow?  What do you think @elf-pavlik ?

elf-pavlik:
Sometimes I use 
* Flow
  * Flow Intent
  * Flow Commitment
  * Flow Event

Since one can use `vf:intpuOf` and `vf:outputOf` on Process and Transfer, in a way it makes Transfer more specific Process.
* Process
  * Transfer - give/issue & receive
  * Transportation - load & unload ...
  * Transformation - produce & consume ...

We should still take a look at how Transfer and Transportation relate. Transfer seems a bit like 'Teleportation'. In case where agents want to track transportation they could have just *load* (no *issue* / *give*) and *unload* (no *receive*). Making `vf:load rdfs:subPropertyOf vf:issue` would imply that if you load from a resource you also issue from it. In similar way `vf:unload rdfs:subPropertyOf vf:receive` would imply that if you unload to a resource you also receive to it.

Could we focus on flows (i, c, e) here and deal with Process, Transfer and Transportation separately?

fosterlynn:
>Could we focus on flows (i, c, e) here and deal with Process, Transfer and Transportation separately?

Yes, sorry about that.

elf-pavlik:
Would you see that we can have an instance of a flow in scope of instance of another flow? For example agent may have intent to *produce* 300kg of salsa in next month. Then they plan smaller processes and one day they have intent to produce 30kg, other day intent to produce 40kg etc. In a way those fractional intents stay in scope of this monthly intent.

This ties a lot to #217 where we discussed that a Proposal would possibly reference an intent to get 500kg of tomatoes and not necessary each fractional amount for each process.
It comes even more useful if we have intents for *work*, *service* and *use*. Here we don't have comfort of resource which acts as a buffer and have to managed everything with flows. To produce 300kg of salsa it would require 20h of work. Each planned process would have *work* intent for fraction of that, but agent could put a proposal for that 20h of *work* and later each intent in planned process could reference that 'aggregated' intent.

fosterlynn:
>Would you see that we can have an instance of a flow in scope of instance of another flow? For example agent may have intent to produce 300kg of salsa in next month. Then they plan smaller processes and one day they have intent to produce 30kg, other day intent to produce 40kg etc. In a way those fractional intents stay in scope of this monthly intent.

This is a lot like the budgeting question or the locecon question.  I was hoping we don't have to scope flows in other flows.  That could be handled by processes being scoped in other processes (and/or plans, etc.), the flows would just follow along.  Or I guess inScopeOf would handle it also if people find a need.  It also could be handled on the reporting side by matching dates and resource classifications, probably, if there are not actual data references.  But, I feel out of touch with a real use case, so it is hard to think about concretely, a lot of my feeling about not doing it is just gut.

What are your thoughts?

elf-pavlik:
> That could be handled by processes being scoped in other processes (and/or plans, etc.), the flows would just follow along.

I see two reasons why not to rely on relations between processes imply something about flows
* we don't require that every flow references some process, so we should handle 'loose' flows
* as we discussed in #286 `vf:inScopeOf` does NOT work as transitive property - so if we consider both `vf:intpuOf` and `vf:outputOf` as subproperties of `vf:inScopeOf`, which would imply that all input and output flows stay in scope of the process. Having that process in scope of another process or plan, DOES NOT imply that those input/output flows also stay in scope of that another process or plan. To imply that we would have to define `vf:inScopeOf` as *transitive property* and I really don't want to do it because it would require all queries to recurse.

fosterlynn:
>we don't require that every flow references some process, so we should handle 'loose' flows

Good point

>vf:inScopeOf does NOT work as transitive property - so if we consider both vf:intpuOf and vf:outputOf as subproperties of vf:inScopeOf, which would imply that all input and output flows stay in scope of the process.

I guess I need to understand all of that better.  Making vf:inputOf and vf:outputOf subproperties of vf:inScopeOf wasn't part of your PR.  And I worry about the scoping concept being so diluted it just becomes "this thing is related to that thing".  Why do we want to make those subproperties instead of just leaving them as they are?

fosterlynn:
>I see two reasons why not to rely on relations between processes imply something about flows

Do you disagree that flows should be able to be used at different levels of the model using the same naming?  If so, let's look at how that would work.

Or is this more that you agree, but want to work out the details more to see how it would work?

fosterlynn:
>Having that process in scope of another process or plan, DOES NOT imply that those input/output flows also stay in scope of that another process or plan. To imply that we would have to define vf:inScopeOf as transitive property and I really don't want to do it because it would require all queries to recurse.

Well, I think I have finally understood your transitive vs intransitive properties concern!  I think the essence of scoping implies recursion.  Otherwise it doesn't really mean "scope".  (Which also is part of why we should not over-use the term.)  And recursion is really useful in many parts of VF, like parent-child organizations.  We do it all the time in NRP, and of course it is more work, but it also does what it needs to do.  Are you mostly concerned because of possibly needing to do the recursion over the web from different sources?

elf-pavlik:
> Do you disagree that flows should be able to be used at different levels of the model using the same naming? If so, let's look at how that would work.

I think we can merge this flows related part. What do you think about removing the *Economic Interaction* part of the picture and for now just leave the *Flow* part? This would actually match change to prose better since you haven't change anything in sections talking about Processes and Transfers.

fosterlynn:
>Note: This is the picture that should appear in the flows.md file, hopefully it does.

I removed the process/transfer part of the picture.

docs/introduction/flows.md
```diff
@@ -17,4 +17,10 @@ Commitments describe potential future event which involved agents already agreed
...
 
-Economic event describes past event, something already observed, never some potential future event.
\ No newline at end of file
+Economic event describes past event, something already observed, never some potential future event.
+
+## Granularity
+
+Intents, Commitments, and Economic Events can occur at any granularity that is needed or for which data can be obtained.  So they primarily are used for all operational needs, but can also be used at higher levels for analytical and high level planning needs for communities or regions, for example.
+
+![Intent-Commitment-Event](https://rawgit.com/valueflows/valueflows/master/release-doc-in-process/i-c-e.png)
```
elf-pavlik: This link still has old picture, as well as https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/i-c-e.png

PR #295, opened 2018-11-23
simpified classifications

elf-pavlik:
based on conversation in #293

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -30,22 +30,6 @@ vf:AgentRelationshipRole  a  owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
 
-vf:ResourceClassification  a    owl:Class ;
-        rdfs:label           "vf:ResourceClassification" ;
-        rdfs:comment   "A resource category, model, taxonomy item, tag, facet value, or other kind of classification." .
-
-vf:ProcessClassification  a           owl:Class ;
-        rdfs:label           "vf:ProcessClassification" ;
-        rdfs:comment   "A process category, model, taxonomy item, tag, facet value, or other kind of classification." .
-
-vf:ExchangeClassification  a           owl:Class ;
-        rdfs:label           "vf:ExchangeClassification" ;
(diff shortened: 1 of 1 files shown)


release-doc-in-process/all_vf.TTL
```diff
@@ -179,33 +153,12 @@ vf:committedOutputs
...
+vf:classifiedAs
         a                   owl:ObjectProperty ;
-        rdfs:label          "exchange classified as" ;
-        rdfs:domain         vf:Transfer ;
-        rdfs:range          vf:TransferClassification ;
-        rdfs:comment        "The primary classification that describes a transfer, at any level of granularity." .
+        rdfs:label          "classified as" ;
+        rdfs:domain         owl:Thing ;
+        rdfs:range          owl:Thing ;
+        rdfs:comment        "References concept in a common taxony for purpose of categorizing." .
```
fosterlynn: Suggestion on definition:  References a concept in a common taxonomy or other classification scheme for purposes of categorization.

release-doc-in-process/all_vf.TTL
```diff
@@ -30,22 +30,6 @@ vf:AgentRelationshipRole  a  owl:Class ;
...
-        rdfs:comment   "A process category, model, taxonomy item, tag, facet value, or other kind of classification." .
-
-vf:ExchangeClassification  a           owl:Class ;
-        rdfs:label           "vf:ExchangeClassification" ;
-        rdfs:comment   "An exchange category, model, taxonomy item, tag, facet value, or other kind of classification." .
-
-vf:TransferClassification  a     owl:Class ;
-        rdfs:label               "vf:TransferClassification" ;
-        rdfs:comment   "A transfer category, model, taxonomy item, tag, facet value, or other kind of classification." .
-
```
fosterlynn: Can we come up with new names and keep these around to build off of for the knowledge layer?  And also the reference properties with new names?

fosterlynn:
Looks good.  I want to make sure before merging this that taking away the classes in the "top layer" does not mean we can't get them back with new names very soon though.  Are you good with a new top layer that has recipes and other configurations needed to drive the other layers?  Including classes that rename the ResourceClassification, ProcessClassification, etc.?  We'll want to fill that back in as quickly as we can manage it.

Issue #290, opened 2018-11-15
How to model upper-level processes?


fosterlynn:
This is to discuss if and/or how we want to support process nesting.  I'll do a PR when we have answered these few questions.

There are several requirements that have come up that might fit in here:
* Processes to handle activities that cover a lot of operational processes, often "overhead" processes, like monthly bookkeeping, cleaning the space, etc.  If we support nested processes, we can support this kind of thing, still on an operational level.  The nesting process can have its own inputs and outputs, and also include the lower-level processes.  This was discussed earlier as part of thinking about projects that are not agents.
* Plan.  This nests one or more series of processes that flow into one or more final deliverables (outputs) for the plan.  It is the same as Work Order in NRP.  This is useful to scope to processes that will be coordinated from one context, and are related in some way that makes sense to the users.  Like a campaign to inform about a milestone or do fundraising.
* Function.  This is from http://locecon.org.  It is a higher level process used for analysis of a community, region, etc. to see what can be connected, what gaps can be filled, etc. to make a more resilient economic network.
* Activity.  This is from discussion with @almereyda and @y0va around a budgeting app they are designing.  This is very similar to Function, and will not be called Activity in any case, but I'm listing it separately because it is not exactly locecon Function. #284 

These are processes in that they transform or transport inputs into outputs.  They might reference all or some of Intent, Commitment, EconomicEvent, depending on the use case.  I expect we will come across other examples and needs.

Nesting:  See the picture here https://github.com/valueflows/valueflows/issues/284#issuecomment-433746338.  Basically, higher level processes are not not driven by a simple taxonomy, they contain processes in value flows.  Their inputs and outputs are only the ones that are not created and then consumed within themselves. 

Here are the options I can think of.
![process-options](https://user-images.githubusercontent.com/3776081/48576634-adba9180-e8da-11e8-83c8-b4aaccdf84f6.png)

1. Processes can be in scope of any number of higher level processes, as well as agents, etc.  The behavior (purpose, type, role, ?) is defined as an explicit property.  Like "plan", "function", etc.

2. Same as 1. but use a more specific relationship for nested processes.

3. Same as 1. or 2. but use just a string for the behavior.

4. Explicitly define the higher level processes as classes, Plan, Function, etc. instead of having higher level processes.  Although we still might want that for the first requirement above.

In summary: Here are the questions:
* Do we want to use nesting Processes for anything that basically follows the I-P-O definition we are using? Or do we want to define all of those cases explicitly as different classes?
[...]

PR #296, opened 2018-11-25
Recipes

fosterlynn:
DO NOT MERGE, this is a first draft to trigger some discussion.  Besides general feedback, I'm looking for help on the names.

### DIFF
--- a/docs/introduction/recipes.md
+++ b/docs/introduction/recipes.md
@@ -5,7 +5,7 @@ Recipes are for:
 * generating plans for people to do it together,
 * providing signals for coordinating their work.
 
-Recipes contain all the info required to create a resource.  In ERP (Enterprise Resource Planning) terms, it’s a combination of bills of material and routings and suppliers.
+Recipes contain all the info required to create a resource or provide a service.  In ERP (Enterprise Resource Planning) terms, it’s a combination of bills of material and routings and suppliers.
 
 ### Recipe patterns
 
--- a/docs/introduction/recipes.md
+++ b/docs/introduction/recipes.md
@@ -1,0 +1,38 @@
+# Recipes
+
+Recipes are for:
+* documenting how to do something,
+* generating plans for people to do it together,
+* providing signals for coordinating their work.
+
+Recipes contain all the info required to create a resource.  In ERP (Enterprise Resource Planning) terms, it’s a combination of bills of material and routings and suppliers.
+
+### Recipe patterns
+
+Recipe patterns can be used alone, or mixed and matched in a recipe:
+* <b>Manufacturing pattern</b>: assemble or transform input resources into different output resources. This combines bills of material, processing instructions (routings), other inputs like labor and equipment requirements, and possible suppliers for the inputs.  They are structured like an upside down tree. The manufacturing pattern creates recursive structures. In other words, if an input component has a recipe itself, that recipe will be incorporated into its parent, so you can view a processing tree from parents through children unto many generations.  For example:
+    * Assemble a robot from metal, wires, computer chips, software, etc.
+    * Bake bread from flour, yeast, water, etc., using an oven.
+* <b>Workflow pattern</b>: change the same resource into a different stage of the same resource. This describes a sequence of processes used to complete work on one resource.  They create a series of stages that one resource will go through until it is finished. For example:
+    * Translate a source document, edit the translation, format for publication, and publish.
+    * Harvest, dry and garble (refine) batches of herbs.
+    * Repair a bike.
+    * Do quality testing on something that was created using the manufacturing pattern.
+
+The examples above have to do with producing specific goods or services.  Recipes can also be used to document more general business processes that an organization uses to produce more generally defined kinds of outputs.  For example, an R&D process might be hard to document in detail, but it does have general steps like researching existing technology, concept generation, experimentation, concept selection, refinement, testing, documenting the design. 
+
+
+### Generation of plans
+
+<b>Back-scheduling a plan from a Recipe</b>:
+Start with end items and a due date, generate the plan from the end item to its inputs, to the outputs leading the inputs, to their inputs, etc.
+
+<b>Forward-scheduling from a Recipe</b>:
+Start with the inputs with no predecessors and a start date, generate the plan from the inputs to their outputs, to the inputs that want the outputs, etc.
+
+<b>Forward-scheduling from a Resource</b>:
+Start with a Resource and generate the plan based on its recipe.  Examples: 
+    Translation: start with a source document
+    Auto repair: start with an auto that needs repair.
+
+
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -36,15 +36,15 @@ vf:AgentRelationshipRole  a  owl:Class ;
 
 vf:ResourceSpecification a owl:Class ;
         rdfs:label         "vf:ResourceSpecification" ;
-        rdfs:comment       "." .
+        rdfs:comment       "Specification of a kind of resource. Could define a material item, service, digital item, currency account, etc." .
 
 vf:RecipeProcess a          owl:Class ;
         rdfs:label         "vf:RecipeProcess" ;
-        rdfs:comment       "." .
+        rdfs:comment       "Specifies the process part of a recipe for use in planning from recipe." .
 
 vf:RecipeFlow a            owl:Class ;
         rdfs:label         "vf:RecipeFlow" ;
-        rdfs:comment       "The linkage between resource types, process types, and actions that structure a recipe." .
+        rdfs:comment       "The linkage between a resource type, process type, and action that create the structure of a recipe." .
 
 
 # PLAN CLASSES
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -420,6 +420,20 @@ vf:hasDuration a        owl:ObjectProperty ;
         rdfs:range      time:TemporalDuration ;
         rdfs:comment    "The calendar duration defined for this type of process." .
 
+vf:resourceTypedAs
+        a                   owl:ObjectProperty ;
+        rdfs:label          "resource typed as" ;
+        rdfs:domain         [ owl:unionOf (vf:Commitment vf:EconomicResource) ] ;
+        rdfs:range          vf:ResourceType ;
+        rdfs:comment        "A resource's primary classification, part of describing that resource." .
+
+vf:processTypedAs
+        a                   owl:ObjectProperty ;
+        rdfs:label          "process typed as" ;
+        rdfs:domain         vf:Process ;
+        rdfs:range          vf:ProcessClassification ;
+        rdfs:comment        "A process's primary classification, as part of describing that process." .
+
 
 # ################################ agent relationship verbs
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -390,23 +390,17 @@ vf:definedQuantity a    owl:ObjectProperty ;
 #        rdfs:domain     vf:ResourceSpecification ;
 #        rdfs:comment    "Defines the category of behavior expected for the type of resource." .
 
-vf:variantOf a          owl:ObjectProperty ;
-        rdfs:label      "variant of" ;
-        rdfs:domain     vf:Recipe ;
-        rdfs:range      vf:Recipe ;
-        rdfs:comment    "This recipe is a variation of another recipe." .
+#vf:variantOf a          owl:ObjectProperty ;
+#        rdfs:label      "variant of" ;
+#        rdfs:domain     vf:Recipe ;
+#        rdfs:range      vf:Recipe ;
+#        rdfs:comment    "This recipe is a variation of another recipe." .
 
-vf:successorOf a        owl:ObjectProperty ;
-        rdfs:label      "successor of" ;
-        rdfs:domain     vf:Recipe ;
-        rdfs:range      vf:Recipe ;
-        rdfs:comment    "This recipe is an updated version of another recipe." .
-
-#vf:affected a           owl:ObjectProperty ;
-#        rdfs:label      "affected" ;
-#        rdfs:domain     vf:RecipeFlow ;
-#        rdfs:range      vf:RecipeResource ;
-#        rdfs:comment    "Defines the type of resource this recipe flow affects." .
+#vf:successorOf a        owl:ObjectProperty ;
+#        rdfs:label      "successor of" ;
+#        rdfs:domain     vf:Recipe ;
+#        rdfs:range      vf:Recipe ;
+#        rdfs:comment    "This recipe is an updated version of another recipe." .
 
 vf:duration a           owl:ObjectProperty ;
         rdfs:label      "duration" ;
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -30,9 +30,9 @@ vf:AgentRelationshipRole  a  owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
 
-vf:Recipe a                owl:Class ;
-        rdfs:label         "vf:Recipe" ;
-        rdfs:comment       "Recipes contain all the info required to create or improve or move a resource." .
+#vf:Recipe a                owl:Class ;
+#        rdfs:label         "vf:Recipe" ;
+#        rdfs:comment       "Recipes contain all the info required to create or improve or move a resource." .
 
 vf:ResourceSpecification a owl:Class ;
         rdfs:label         "vf:ResourceSpecification" ;
@@ -143,7 +143,7 @@ vf:inputOf
 vf:outputOf
         a                   owl:ObjectProperty ;
         rdfs:label         "output of" ;
-        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow vf:Recipe) ] ;
+        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow) ] ; # vf:Recipe
         rdfs:range          [ owl:unionOf (vf:Process vf:Transfer vf:RecipeProcess) ] .
 
 vf:inputs
@@ -176,8 +176,6 @@ vf:classifiedAs
         rdfs:domain         owl:Thing ;
         rdfs:range          owl:Thing ;
         rdfs:comment        "References a concept in a common taxonomy or other classification scheme for purposes of categorization." .
-
-# waiting to do the recipe model
 
 vf:involves
         a                   owl:ObjectProperty ;
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -44,7 +44,7 @@ vf:RecipeProcess a          owl:Class ;
 
 vf:RecipeFlow a            owl:Class ;
         rdfs:label         "vf:RecipeFlow" ;
-        rdfs:comment       "The linkage between a resource type, process type, and action that create the structure of a recipe." .
+        rdfs:comment       "The linkage between a process recipe, an action that structures a recipe, and a resource specification." .
 
 
 # PLAN CLASSES
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -408,32 +408,45 @@ vf:successorOf a        owl:ObjectProperty ;
         rdfs:range      vf:Recipe ;
         rdfs:comment    "This recipe is an updated version of another recipe." .
 
-vf:actsOn a             owl:ObjectProperty ;
-        rdfs:label      "acts on" ;
+vf:affected a           owl:ObjectProperty ;
+        rdfs:label      "affected" ;
         rdfs:domain     vf:RecipeFlow ;
-        rdfs:range      vf:ResourceType ;
-        rdfs:comment    "Defines the type of resource this recipe flow acts on." .
+        rdfs:range      vf:RecipeResource ;
+        rdfs:comment    "Defines the type of resource this recipe flow affects." .
 
 vf:hasDuration a        owl:ObjectProperty ;
         rdfs:label      "duration" ;
-        rdfs:domain     vf:ProcessType ;
+        rdfs:domain     vf:RecipeProcess ;
         rdfs:range      time:TemporalDuration ;
         rdfs:comment    "The calendar duration defined for this type of process." .
 
-vf:resourceTypedAs
+vf:durationMultiplier 
+        a               owl:DataTypeProperty ;
+        rdfs:label      "multiplier" ;
+        rdfs:domain     vf:RecipeProcess ;
+        rdfs:range      ???? ; #TODO
+        rdfs:comment    "The calendar duration defined for this type of process." .
+
+vf:resourceBasedOn
         a                   owl:ObjectProperty ;
         rdfs:label          "resource typed as" ;
         rdfs:domain         [ owl:unionOf (vf:Commitment vf:EconomicResource) ] ;
-        rdfs:range          vf:ResourceType ;
+        rdfs:range          vf:RecipeResource ;
         rdfs:comment        "A resource's primary classification, part of describing that resource." .
 
-vf:processTypedAs
+vf:processBasedOn
         a                   owl:ObjectProperty ;
         rdfs:label          "process typed as" ;
         rdfs:domain         vf:Process ;
-        rdfs:range          vf:ProcessClassification ;
+        rdfs:range          vf:RecipeProcess ;
         rdfs:comment        "A process's primary classification, as part of describing that process." .
 
+vf:deliverable
+        a                   owl:ObjectProperty ;
+        rdfs:label          "deliverable" ;
+        rdfs:domain         vf:RecipeFlow ;
+        rdfs:range          vf:Recipe ;
+        rdfs:comment        "The main deliverable(s) of a recipe." .
 
 # ################################ agent relationship verbs
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -30,19 +30,19 @@ vf:AgentRelationshipRole  a  owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
 
-vf:Recipe a               owl:Class ;
+vf:Recipe a                owl:Class ;
         rdfs:label         "vf:Recipe" ;
         rdfs:comment       "Recipes contain all the info required to create or improve or move a resource." .
 
-vf:ResourceType a         owl:Class ;
-        rdfs:label         "vf:ResourceType" ;
+vf:ResourceSpecification a owl:Class ;
+        rdfs:label         "vf:ResourceSpecification" ;
         rdfs:comment       "." .
 
-vf:ProcessType a          owl:Class ;
-        rdfs:label         "vf:ProcessType" ;
+vf:RecipeProcess a          owl:Class ;
+        rdfs:label         "vf:RecipeProcess" ;
         rdfs:comment       "." .
 
-vf:RecipeFlow a           owl:Class ;
+vf:RecipeFlow a            owl:Class ;
         rdfs:label         "vf:RecipeFlow" ;
         rdfs:comment       "The linkage between resource types, process types, and actions that structure a recipe." .
 
@@ -138,13 +138,13 @@ vf:inputOf
         a                   owl:ObjectProperty ;
         rdfs:label         "input of" ;
         rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow) ] ;
-        rdfs:range          [ owl:unionOf (vf:Process vf:Transfer vf:ProcessType) ] .
+        rdfs:range          [ owl:unionOf (vf:Process vf:Transfer vf:RecipeProcess) ] .
 
 vf:outputOf
         a                   owl:ObjectProperty ;
         rdfs:label         "output of" ;
-        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow) ] ;
-        rdfs:range          [ owl:unionOf (vf:Process vf:Transfer vf:ProcessType) ] .
+        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow vf:Recipe) ] ;
+        rdfs:range          [ owl:unionOf (vf:Process vf:Transfer vf:RecipeProcess) ] .
 
 vf:inputs
         a                   owl:ObjectProperty ;
@@ -368,13 +368,13 @@ vf:fulfills     a owl:ObjectProperty ;
 
 vf:isSubstitutable a    owl:DatatypeProperty ;
         rdfs:label      "is substitutable" ;
-        rdfs:domain     vf:ResourceType ;
+        rdfs:domain     vf:ResourceSpecification ;
         rdfs:range      xsd:boolean ;
         rdfs:comment    "Defines if any resource of that type can be freely substituted for any other resource of that type when used, consumed, traded, etc." .
 
 #vf:unit a               owl:ObjectProperty ;
 #        rdfs:label      "unit" ;
-#        rdfs:domain     vf:ResourceType ;
+#        rdfs:domain     vf:ResourceSpecification ;
 #        rdfs:range      qudt:unit ;
 #        rdfs:comment    "The unit expected for this resource type as a default." .
 
@@ -384,10 +384,10 @@ vf:definedQuantity a    owl:ObjectProperty ;
         rdfs:range      qudt:QuantityValue ;
         rdfs:comment    "The amount and unit defined for this recipe flow." .
 
-# TODO: need to define this, something like currency, inventory, work, service; may be other aspects too.
+# TODO: need to define this, something like currency, inventory, skill, service; may be other aspects too.
 #vf:resourceCategory a    owl:ObjectProperty ;
 #        rdfs:label      "category" ;
-#        rdfs:domain     vf:ResourceType ;
+#        rdfs:domain     vf:ResourceSpecification ;
 #        rdfs:comment    "Defines the category of behavior expected for the type of resource." .
 
 vf:developedBy a        owl:ObjectProperty ;
@@ -408,13 +408,13 @@ vf:successorOf a        owl:ObjectProperty ;
         rdfs:range      vf:Recipe ;
         rdfs:comment    "This recipe is an updated version of another recipe." .
 
-vf:affected a           owl:ObjectProperty ;
-        rdfs:label      "affected" ;
-        rdfs:domain     vf:RecipeFlow ;
-        rdfs:range      vf:RecipeResource ;
-        rdfs:comment    "Defines the type of resource this recipe flow affects." .
+#vf:affected a           owl:ObjectProperty ;
+#        rdfs:label      "affected" ;
+#        rdfs:domain     vf:RecipeFlow ;
+#        rdfs:range      vf:RecipeResource ;
+#        rdfs:comment    "Defines the type of resource this recipe flow affects." .
 
-vf:hasDuration a        owl:ObjectProperty ;
+vf:duration a           owl:ObjectProperty ;
         rdfs:label      "duration" ;
         rdfs:domain     vf:RecipeProcess ;
         rdfs:range      time:TemporalDuration ;
@@ -422,31 +422,26 @@ vf:hasDuration a        owl:ObjectProperty ;
 
 vf:durationMultiplier 
         a               owl:DataTypeProperty ;
-        rdfs:label      "multiplier" ;
+        rdfs:label      "duration multiplier" ;
         rdfs:domain     vf:RecipeProcess ;
-        rdfs:range      ???? ; #TODO
-        rdfs:comment    "The calendar duration defined for this type of process." .
+        rdfs:range      xsd:float ;
+        rdfs:comment    "The multiplier for duration based on capacities for the recipe." .
 
-vf:resourceBasedOn
-        a                   owl:ObjectProperty ;
-        rdfs:label          "resource typed as" ;
-        rdfs:domain         [ owl:unionOf (vf:Commitment vf:EconomicResource) ] ;
-        rdfs:range          vf:RecipeResource ;
-        rdfs:comment        "A resource's primary classification, part of describing that resource." .
+vf:basedOn
+        a               owl:ObjectProperty ;
+        rdfs:label      "based on" ;
+        rdfs:domain     vf:Process ;
+        rdfs:range      vf:RecipeProcess ;
+        rdfs:comment    "The recipe definition or specification for a process." .
 
-vf:processBasedOn
-        a                   owl:ObjectProperty ;
-        rdfs:label          "process typed as" ;
-        rdfs:domain         vf:Process ;
-        rdfs:range          vf:RecipeProcess ;
-        rdfs:comment        "A process's primary classification, as part of describing that process." .
+vf:resourceConformsTo
+        a               owl:ObjectProperty ;
+        rdfs:label      "conforms to" ;
+        rdfs:domain     [ owl:unionOf (vf:Commitment vf:EconomicResource vf:RecipeFlow ) ] ;
+        rdfs:range      vf:ResourceSpecification ;
+        rdfs:comment    "The primary resource knowledge specification or definition of an existing or potential resource." .
 
-vf:deliverable
-        a                   owl:ObjectProperty ;
-        rdfs:label          "deliverable" ;
-        rdfs:domain         vf:RecipeFlow ;
-        rdfs:range          vf:Recipe ;
-        rdfs:comment        "The main deliverable(s) of a recipe." .
+
 
 # ################################ agent relationship verbs
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -390,12 +390,6 @@ vf:definedQuantity a    owl:ObjectProperty ;
 #        rdfs:domain     vf:ResourceSpecification ;
 #        rdfs:comment    "Defines the category of behavior expected for the type of resource." .
 
-vf:developedBy a        owl:ObjectProperty ;
-        rdfs:label      "developed by" ;
-        rdfs:domain     vf:Recipe ;
-        rdfs:range      foaf:Agent ;
-        rdfs:comment    "The agent who developed this recipe." .
-
 vf:variantOf a          owl:ObjectProperty ;
         rdfs:label      "variant of" ;
         rdfs:domain     vf:Recipe ;
@@ -437,7 +431,7 @@ vf:basedOn
 vf:resourceConformsTo
         a               owl:ObjectProperty ;
         rdfs:label      "conforms to" ;
-        rdfs:domain     [ owl:unionOf (vf:Commitment vf:EconomicResource vf:RecipeFlow ) ] ;
+        rdfs:domain     [ owl:unionOf (vf:Commitment vf:EconomicResource vf:Intent vf:RecipeFlow ) ] ;
         rdfs:range      vf:ResourceSpecification ;
         rdfs:comment    "The primary resource knowledge specification or definition of an existing or potential resource." .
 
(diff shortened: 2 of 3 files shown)


docs/introduction/recipes.md
```diff
@@ -0,0 +1,38 @@
+# Recipes
+
+Recipes are for:
+* documenting how to do something,
+* generating plans for people to do it together,
+* providing signals for coordinating their work.
+
+Recipes contain all the info required to create a resource.  In ERP (Enterprise Resource Planning) terms, it’s a combination of bills of material and routings and suppliers.
```
elf-pavlik: "Create a resource or provide a service"
not all recopies have to result in *produce* flow, some can provide a service and make resource available for usage

release-doc-in-process/all_vf.TTL
```diff
@@ -30,6 +30,23 @@ vf:AgentRelationshipRole  a  owl:Class ;
...
 
+vf:Recipe a               owl:Class ;
+        rdfs:label         "vf:Recipe" ;
+        rdfs:comment       "Recipes contain all the info required to create or improve or move a resource." .
+
+vf:ResourceType a         owl:Class ;
+        rdfs:label         "vf:ResourceType" ;
+        rdfs:comment       "." .
+
+vf:ProcessType a          owl:Class ;
```
elf-pavlik: Since you chose `vf:RecipeFlow` lets go with `vf:RecipeProcess` to represent a process on Recipe layer?

release-doc-in-process/all_vf.TTL
```diff
@@ -30,6 +30,23 @@ vf:AgentRelationshipRole  a  owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
 
+vf:Recipe a               owl:Class ;
+        rdfs:label         "vf:Recipe" ;
+        rdfs:comment       "Recipes contain all the info required to create or improve or move a resource." .
+
+vf:ResourceType a         owl:Class ;
```
elf-pavlik: Why would we need it? We already have `vf:classifiedAs` to position resources in taxonomy. I think `vf:RecipeFlow` can also reference the same taxonomy item, just as intents and commitments will often just use `vf:classifiedAs` rather than reference concrete resource.

release-doc-in-process/all_vf.TTL
```diff
@@ -30,6 +30,23 @@ vf:AgentRelationshipRole  a  owl:Class ;
         rdfs:subClassOf   rdf:Property ;
         rdfs:comment   "A relationship role defining the kind of association one agent can have with another." .
 
+vf:Recipe a               owl:Class ;
```
elf-pavlik: Since we go with nested / overlapping processes to represent plans. We probably can in similar way use `vf:RecipeProcess` with similar nesting / overlapping. I understood that from Recipe after scaling quantities one would get a Plan.

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
 #        rdfs:range      xsd:boolean ;
 #        rdfs:comment    "The commitment or process is complete or not." .
 
-# isSubstitutable "Defines if any resources of that type can be freely substituted for any other resource of that type when used, consumed, traded, etc."
+vf:isSubstitutable a    owl:DatatypeProperty ;
```
elf-pavlik: This seems like the only reason for `vf:ResourceType`. I think we have an issue where we discuss substitutable https://github.com/valueflows/resource/issues/25

Does this stay specific to specific inventory or any would always stay the same for resource using the same taxonomy reference?

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:range      vf:Recipe ;
+        rdfs:comment    "This recipe is a variation of another recipe." .
+
+vf:successorOf a        owl:ObjectProperty ;
+        rdfs:label      "successor of" ;
+        rdfs:domain     vf:Recipe ;
+        rdfs:range      vf:Recipe ;
+        rdfs:comment    "This recipe is an updated version of another recipe." .
+
+vf:actsOn a             owl:ObjectProperty ;
```
elf-pavlik: Can we simply use something `vf:affectedClassifiedAs`. Somehow we don't have such property to reference taxonomy of a resource affected by flow, we need something like that for Intent and Commitment and I think we can just use it also with RecipeFlow.

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:range      vf:ResourceType ;
+        rdfs:comment    "Defines the type of resource this recipe flow acts on." .
+
+vf:hasDuration a        owl:ObjectProperty ;
+        rdfs:label      "duration" ;
+        rdfs:domain     vf:ProcessType ;
+        rdfs:range      time:TemporalDuration ;
+        rdfs:comment    "The calendar duration defined for this type of process." .
+
+vf:resourceTypedAs
```
elf-pavlik: Why can't we just rely on `vf:classifiedAs` ?

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        a                   owl:ObjectProperty ;
+        rdfs:label          "resource typed as" ;
+        rdfs:domain         [ owl:unionOf (vf:Commitment vf:EconomicResource) ] ;
+        rdfs:range          vf:ResourceType ;
+        rdfs:comment        "A resource's primary classification, part of describing that resource." .
+
+vf:processTypedAs
```
elf-pavlik: I assume that process can use `vf:classifiedAs` to reference taxonomy and this only would reference a `vf:RecipeProcess` from whjich it was created. I would prefer something more specific as `vf:basedOn`. We may also need to use it to relate flows to `vf:RecipeFlow` from which they were created.

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        a                   owl:ObjectProperty ;
+        rdfs:label          "process typed as" ;
+        rdfs:domain         vf:Process ;
+        rdfs:range          vf:ProcessClassification ;
```
elf-pavlik: `vf:RecipeProcess`

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,74 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:range      vf:Recipe ;
+        rdfs:comment    "This recipe is an updated version of another recipe." .
+
+vf:actsOn a             owl:ObjectProperty ;
+        rdfs:label      "acts on" ;
+        rdfs:domain     vf:RecipeFlow ;
+        rdfs:range      vf:ResourceType ;
+        rdfs:comment    "Defines the type of resource this recipe flow acts on." .
+
+vf:hasDuration a        owl:ObjectProperty ;
```
elf-pavlik: I think sometimes duration could depend on quantities and sometimes not. To assemble 10 bikes it will take around 10x time than 1 bike. On the other hand baking 10 loafs of bread doesn't have to take 2x time of baking 5 loafs.

release-doc-in-process/all_vf.TTL
fosterlynn: Because we have some properties that are useful there as part of a recipe.  The ResourceType or RecipeResource can reference one or more classifications itself to cover that need.

docs/introduction/recipes.md
fosterlynn: How about "to create a resource or provide a service"?  I don't understand about making a resource available for usage, that is just creating the resource, right?  If it is available, that could be an intent, but doesn't need a recipe?

elf-pavlik: We rent bungalow from organization running an orphanage. They actually need to execute a small process before each rental. clean the bungalow, issue keys, read water meters, install new tank of gas, sometimes bring fridge and stove etc. This mostly requires some work, for example scheduling work of people doing the maintenance.

release-doc-in-process/all_vf.TTL
fosterlynn: Are you thinking we don't need a Recipe class?  I put it there as a simple way to handle forks and versions.  And just different ways of producing the same thing. I think a data structure that doesn't include everything for a recipe will be just hard to deal with, especially with distributed information across lots of recipe creators.

release-doc-in-process/all_vf.TTL
elf-pavlik: Looking at this PR I only see `vf:isSubstitutable`

release-doc-in-process/all_vf.TTL
fosterlynn: It is  helpful for planning from a recipe to know if the resource type is substitutable, because if it is not (like say documents), then you can't just find another one in inventory and use that one.  Every resource of that type will "inherit" that, it is a property of the type.

release-doc-in-process/all_vf.TTL
elf-pavlik: I think we should make it consistent wit how we model Plan
if from Recipe we get Plan
and we represent Plan as Process
we should represent Recipe as RecipeProcess
It just needs to map clearly to how we do things on Planning layer.

release-doc-in-process/all_vf.TTL
fosterlynn: I guess that depends on the resolution of if we need a RecipeResource thing rather than just a classification.  I like "affects" better than "acts on" -- but we already used that.... and again might depend on if we have a RecipeResource thing.

release-doc-in-process/all_vf.TTL
fosterlynn: basedOn is good, thanks!

release-doc-in-process/all_vf.TTL
fosterlynn: Good point, talked to Bob a while about the complexity of capacity planning, which he has some experience in.  I'm thinking adding a multiplier in RecipeProcess, seems useful.  We might think of other things with some research; or can just leave it as extensible.  For example equipment has a capacity, so the bread multiplier might depend on that.  But the capacity of the oven is different for bread and for cookies.

docs/introduction/recipes.md
fosterlynn: Got it.  (And missed your initial suggestion somehow, sorry.)

release-doc-in-process/all_vf.TTL
fosterlynn: isSubstitutable is enough imo... but also note, name, image go there... and I guess I didn't get some other configuration things done, it is useful to know if it is work, currency, etc. for behavior but I don't have that well defined and am not sure it should be in VF, although they would be good extensions.... also thinking about some of the standard identifiers that people could extend it with like GTIN, etc. that identify very specific resource types (could use gr vocabulary for those)

And we need something that a resource only has one of, which is not the case for classification.

elf-pavlik:
Can we start in small steps and first only introduce
* `vf:ProcesRecipe`
* `vf:FlowRecipe`
* `vf:basedOn`

This would give us something like here (without agreement part)

Later we can deal with all the other issues one by one.

release-doc-in-process/all_vf.TTL
fosterlynn: Interesting thought.  Recipe and Plan are not always the same granularity.  Often plans will include more than one deliverable, which often implies more than one recipe - when the deliverables are for one purpose or goal, but produced separately.  Even if a recipe has more than one output, it would never include value flows that don't intersect, it is all connected processes.

But even with that, your point is still valid, still thinking about the idea.  Recipe does have some different properties too than you would find on a process type, but if we called out the purposes of processes and purposes of  process types, say with subclassing, then we could do that easily.

fosterlynn:
>Can we start in small steps and first only introduce   vf:ProcesRecipe    vf:FlowRecipe   vf:basedOn

I'm good with starting in smaller steps, but I think some of the naming will depend on other parts of the model.  

Also, I made a mistake (thanks @bhaugen ! ) with how I hooked up the new Recipe class, it needs to hook in at the RecipeFlow.  That is where it has a M:1 relationship with a recipe.  For example, "this is a recipe for 20 widgets", not "this is a recipe for widget".   The RecipeFlow will belong to only 1 Recipe.  The only reason you would have multiple RecipeFlows for 1 Recipe is if you have co-products or similar.

release-doc-in-process/all_vf.TTL
fosterlynn: Still struggling with this naming.  I need to make sure that this isn't needed outside of recipes - for example for high level processes.  Changed in the meantime, but still open question for me.

release-doc-in-process/all_vf.TTL
fosterlynn: Changed name to affected.

fosterlynn:
>Recipe class, it needs to hook in at the RecipeFlow.

Fixed this, and made some changes related to the review from @elf-pavlik

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,87 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:comment    "The calendar duration defined for this type of process." .
+
+vf:durationMultiplier 
+        a               owl:DataTypeProperty ;
+        rdfs:label      "multiplier" ;
+        rdfs:domain     vf:RecipeProcess ;
+        rdfs:range      ???? ; #TODO
+        rdfs:comment    "The calendar duration defined for this type of process." .
+
+vf:resourceBasedOn
```
elf-pavlik: I would not see a resource itself based on recipe. Just the flow that will *produce* that resource would stay *basedOn* a `FlowRecipe`. And again for now I would like to start with that `FlowRecipe` simply referencing a taxonomy with `vf:affectedClassifiedAs`. I don't think we need `RecipeResource` at all, or at least not to get started.

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,87 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        a                   owl:ObjectProperty ;
+        rdfs:label          "resource typed as" ;
+        rdfs:domain         [ owl:unionOf (vf:Commitment vf:EconomicResource) ] ;
+        rdfs:range          vf:RecipeResource ;
+        rdfs:comment        "A resource's primary classification, part of describing that resource." .
 
-# something for models for GTIN, etc. identifiers
+vf:processBasedOn
```
elf-pavlik: I think we can use same `vf:basedOn` between
* `vf:Process` and `vf:ProcessRecipe`
*  Flow (`vf:Intent` & `vf:Commitment`) and `vf:FlowRecipe`

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,87 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        a                   owl:ObjectProperty ;
+        rdfs:label          "process typed as" ;
+        rdfs:domain         vf:Process ;
+        rdfs:range          vf:RecipeProcess ;
+        rdfs:comment        "A process's primary classification, as part of describing that process." .
+
+vf:deliverable
```
elf-pavlik: This needs to match what we do with Plan. To get started we can simply use `vf:outputOf`

release-doc-in-process/all_vf.TTL
elf-pavlik: To clarify, resource with same classification could get produced from various alternative recipes.

release-doc-in-process/all_vf.TTL
elf-pavlik: I think we need to clarify how Plan and Recipe relate. I would expect that for any Plan we can create Recipe that would automate crating similar (isomorphic?) plans.

fosterlynn: In our experience: You can use one or more recipes to create a plan.  There's a bunch of logic that goes in between those though, not a data relationship that seems useful to me.  And often there are tweaks to the plan after generating from a recipe.

For any plan you could create a recipe for each deliverable, yes.  Which would be valid if the plan didn't rely on inventory, etc.  We have had requests to be able to create a recipe from a flow of actual processes and economic events, like "hey this worked pretty well, let's create a recipe".

release-doc-in-process/all_vf.TTL
fosterlynn: >We may also need to use it to relate flows to vf:RecipeFlow from which they were created.

Not that I know of. Let's wait and see if that is needed for anything.

release-doc-in-process/all_vf.TTL
elf-pavlik: Maybe let's first figure out how we model plans and nested/overlapping processes on Planning layer and then we try to see what we need on Recipe layer to automate crating constructs on Planning layer? I think we can already start with `vf:ProcessRecipe` and `vf:FlowRecipe` and see how far this gets us.

release-doc-in-process/all_vf.TTL
fosterlynn: >To clarify, resource with same classification could get produced from various alternative recipes.

Right, that is why the Recipe relates to the RecipeFlow.  And EconomicResource relates to resource type-thing. Yes, the resource type-thing could be the same but produced in different ways.  Maybe I misunderstand?

>I don't think we need RecipeResource at all, or at least not to get started.

How would you possibly use a recipe without it?  I see no need to put out a partial model that doesn't make sense, it will just confuse people.  Better to have none at all.  If we do a partial model, I would like it to make sense as it is.

release-doc-in-process/all_vf.TTL
elf-pavlik: :-1: to similar to `vf:affects` and if it references classification of the resource `vf:affectedClassifiedAs` seems much more clear. Actually we shouldn't handle it in Recipes PR since we also seem to need at `vf:Intent` and `vf:Commitment` (*produce*, *consume*, *use*) to reference classification of a resource affected by that flow.

release-doc-in-process/all_vf.TTL
fosterlynn: I guess that is the same as we did the other day with includedIn.  Are you sure?  The UML model will help because it will detail it properly.  But there will be no machine readable definition that is correct.  Possibly a larger topic, and it is in my mind because of the Protege issue.

release-doc-in-process/all_vf.TTL
fosterlynn: Ouch, that is another one of those convoluted range and domain definitions.  Plan does not have its own outputOf though.  Although it could in the sense of defined deliverables which may get changed based on the lower level processes in the plan, I was thinking of adding deliverable there, but am still thinking about differences between plans and operational processes.

I guess in any case it could be vf:outpufOf.  I just hate to add it to that already confusing and fuzzily defined property that has to be clarified in text somewhere that people may or may not see. I wonder if there is some way in rdf to make that more precise and not have to use different vocabulary? :pray:

release-doc-in-process/all_vf.TTL
fosterlynn: >-1 to similar to vf:affects and if it references classification of the resource vf:affectedClassifiedAs seems much more clear.

Well right now it doesn't reference a classification, it references something on the knowledge layer, and we have finally gotten clear on the difference.  I realize you don't agree that should exist. :)

>Actually we shouldn't handle it in Recipes PR since we also seem to need at least vf:Intent and vf:Commitment (produce, consume, use) to reference classification of a resource affected by that flow.

But the recipe won't hang together without it.  We might need vf:Intent and vf:Commitment to reference this knowledge layer class, good point - but right now they can reference the classification.  We can add that before doing this PR if you like.  But we'll still need the name.

release-doc-in-process/all_vf.TTL
fosterlynn: Can we resolve this one?  It is the same as the question of if we have something like a RecipeResource or ResourceType in the knowledge layer.  We'll need to clean up all naming based on that resolution.

fosterlynn:
One thing I'm wondering about is if we can always use vf:classifiedAs references to external taxonomies etc. for intents, or do we want also the knowledge layer thing?  Like a provider's product catalog kind of thing, or any offer that specifies the classification very exactly including units, etc.  Like "Haralson apples #2 organic bushel" and "Haralson apples #2 organic pound".  Or even "Haralson apples #2 organic bushel from Turkey Ridge Orchard".  (Maybe Turkey Ridge Orchard produced higher quality apples.)

In any case, something that can be offered that the offering agent did not produce.  Maybe there is a recipe somewhere and maybe there isn't.

To me, this affects whether we name the knowledge layer thing as RecipeResource.

release-doc-in-process/all_vf.TTL
elf-pavlik: Ok, let's say we have `vf:ResourceSpecification`. I think we'll need to have two distinct properties that flows can use:
* a Flow - `vf:resourceClassifiedAs` -> a taxonomy item
* a Flow - `vf:resourceConformsTo` -> a Resource Specification

While data would in most cases also have *a Resource Specification - `vf:classifiedAs` -> a taxonomy  item. I would prefer that to get the classification we don't have to sometimes get it directly from the flow and sometimes go through the resource specification. Systems which rely on Resource Specification can always just copy over the taxonomy reference to every flow referencing that specification.

elf-pavlik:
> Like a provider's product catalog kind of thing, or any offer that specifies the classification very exactly including units, etc. Like "Haralson apples #2 organic bushel" and "Haralson apples #2 organic pound".

I think units get referenced from flows, most likely proposal would reference an intent that would provide the unit. Last comment in https://github.com/valueflows/resource/issues/25 also references concept of  *Stock Keeping Unit*.

When it comes to manufactures, wikidata for example includes reference to manufacturer and brand of a specific car model https://www.wikidata.org/wiki/Q603079

release-doc-in-process/all_vf.TTL
fosterlynn: > I would prefer that to get the classification we don't have to sometimes get it directly from the flow and sometimes go through the resource specification. Systems which rely on Resource Specification can always just copy over the taxonomy reference to every flow referencing that specification.

Makes sense to me.  But we can't define that, right?  It has totally open range and domain.  We can consider tightening that up if you like.

Also, does that mean we want to require a resource to have a resource specification?

fosterlynn:
@elf-pavlik I just pushed several name changes based on our discussions.  There might still be some holes, particularly in relationship property names - and there are a couple open questions in the review too.  But I think we are getting close.  

I think I will put this into UML so I can see it better.  And the ttl could use a detailed code review too.

How are you feeling about simplifying the PR at this point?  I'd like to get it all in if we can.

fosterlynn:
This is what I have now.  Might not be exactly right on the properties.  (Won't be part of this PR, I'd like to catch up the UML after this connected batch of PRs.)

recipe-uml

release-doc-in-process/all_vf.TTL
elf-pavlik: I don't think we can use `vf:classifiedAs` directly from the Flow. We can do it from a Resource, Process and possibly other constructs. In a Flow we don't reference taxonomy that categorizes the flow itself but the resource that this flow affects. For service flows using `vf:classifiedAs` might make sense since the flow itself seems to represent service. Though sometimes the process that outputs a service would have some `vf:classifiedAs` and not necessary the flow itself.

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,82 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:range      qudt:QuantityValue ;
+        rdfs:comment    "The amount and unit defined for this recipe flow." .
+
+# TODO: need to define this, something like currency, inventory, skill, service; may be other aspects too.
+#vf:resourceCategory a    owl:ObjectProperty ;
+#        rdfs:label      "category" ;
+#        rdfs:domain     vf:ResourceSpecification ;
+#        rdfs:comment    "Defines the category of behavior expected for the type of resource." .
+
+vf:developedBy a        owl:ObjectProperty ;
```
elf-pavlik: once again, I don't think we need to create property to represent authorship, recommending `dcterms:creator` seems sufficient. http://www.dublincore.org/documents/dces/#creator

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,82 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:label      "developed by" ;
+        rdfs:domain     vf:Recipe ;
+        rdfs:range      foaf:Agent ;
+        rdfs:comment    "The agent who developed this recipe." .
+
+vf:variantOf a          owl:ObjectProperty ;
```
elf-pavlik: I don't think this comes as fundamental and possibly we can take advantage of PROV https://www.w3.org/TR/prov-primer/

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,82 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        rdfs:label      "variant of" ;
+        rdfs:domain     vf:Recipe ;
+        rdfs:range      vf:Recipe ;
+        rdfs:comment    "This recipe is a variation of another recipe." .
+
+vf:successorOf a        owl:ObjectProperty ;
```
elf-pavlik: I don't think this comes as fundamental and possibly we can take advantage of PROV https://www.w3.org/TR/prov-primer/

elf-pavlik:
Yes @fosterlynn I think we make progress and will just need to capture some issues that arouse here. I left few comments inline.

I still would like to follow up on https://github.com/valueflows/valueflows/pull/296#discussion_r236074718

Could you please give an example where you can not use `vf:RecipeProcess` (an 'upper' one) to represent a Recipe? Based on `vf:RecipeProcess` we get a `vf:Process` on Planning layer, what do you get on Planning layer from a Recipe?

release-doc-in-process/all_vf.TTL
```diff
@@ -346,9 +366,82 @@ vf:fulfills     a owl:ObjectProperty ;
...
+        a               owl:ObjectProperty ;
+        rdfs:label      "based on" ;
+        rdfs:domain     vf:Process ;
+        rdfs:range      vf:RecipeProcess ;
+        rdfs:comment    "The recipe definition or specification for a process." .
+
+vf:resourceConformsTo
+        a               owl:ObjectProperty ;
+        rdfs:label      "conforms to" ;
+        rdfs:domain     [ owl:unionOf (vf:Commitment vf:EconomicResource vf:RecipeFlow ) ] ;
```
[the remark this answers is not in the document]
fosterlynn: Done.

release-doc-in-process/all_vf.TTL
fosterlynn: Done

release-doc-in-process/all_vf.TTL
fosterlynn: I borrowed variantOf and successorOf from GR.  Although now I notice they have isVariantOf.  How would we use prov?

I think forking and versioning recipes is not fundamental to REA, but is an important part of supporting open hardware specifications.  I don't care too much how we do it.

fosterlynn:
>I still would like to follow up on #296 (comment)

>Could you please give an example where you can not use vf:RecipeProcess (an 'upper' one) to represent a Recipe? Based on vf:RecipeProcess we get a vf:Process on Planning layer, what do you get on Planning layer from a Recipe?

I am fine with generally keeping that pattern to match the process nesting.  I am fine with making Recipe a subclass of RecipeProcess if you like.  I do think it is important to call out the name, otherwise it will be confusing, and now the model is quite clear, so I don't want to obscure the Recipe class.

I was thinking we could deal with that when we finish process nesting, so that we get everything working the same on both layers.  I am not sure how close we are on that. 

>what do you get on Planning layer from a Recipe?

I hope a Plan that is a process, thinking about subclassing there too.  (Noting that a Plan can use more than one Recipe.)  But obviously not done yet, and don't know where we are at on that one.

elf-pavlik:
Can we start with `vf:ProcessRecipe` and once we move forward with modeling Plan on Planning layer we will adjust Recipe layer accordingly? If you prefer we can just name it `vf:Recipe` and don't prefix it with *Process*.

>> what do you get on Planning layer from a Recipe?

> I hope a Plan that is a process, thinking about subclassing there too. (Noting that a Plan can use more than one Recipe.) But obviously not done yet, and don't know where we are at on that one.

I guess we might need to look at composing plans into a bigger plan. Possibly each Recipe yields a Plan and those get composed into a 'upper' plan. I think once we write some example data we'll get a better grip of what we really need.

release-doc-in-process/all_vf.TTL
elf-pavlik: I've created dedicated issue for it https://github.com/valueflows/valueflows/issues/298

I would prefer to start with less and proposed `vf:variantOf` and `vf:successorOf` do not seem to me fundamental enough to try to get them in this first pass.

fosterlynn:
>Can we start with vf:ProcessRecipe and once we move forward with modeling Plan on Planning layer we will adjust Recipe layer accordingly? If you prefer we can just name it vf:Recipe and don't prefix it with Process.

I don't think so, either one.  People need to know the differences as well as the similarities.  And we need the functional meanings out there clearly imo.  Can you draw a model that has a clear structure without those two entities?

>I guess we might need to look at composing plans into a bigger plan. Possibly each Recipe yields a Plan and those get composed into a 'upper' plan.

People could certainly do that, but I expect our model will support both, depending on how closely we tie things together between the recipe and planning.  This is the reason to work on the nesting on both layers at the same time, unless you think the 2 names on the top layer is a show stopper.

> I think once we write some example data we'll get a better grip of what we really need.

We have running code in NRP/OCP for creating plans from recipes, used by several groups in production - except for the Recipe class which will support forking recipes.  But examples are good of course, especially to document it for people.

release-doc-in-process/all_vf.TTL
fosterlynn: OK, I will drop them from this PR.  Probably good to take a more comprehensive look.

fosterlynn: Done.

elf-pavlik:
> We have running code in NRP/OCP for creating plans from recipes, used by several groups in production - except for the Recipe class which will support forking recipes.

I understood that NRP only has equivalent of proposed `vf:RecipeProcess`. Could you please quickly scribble an example of `vf:Recipe` with at least two `vf:RecipeProcess` and couple of `vf:RecipeFlow` ?

fosterlynn:
>Could you please quickly scribble an example of vf:Recipe with at least two vf:RecipeProcess and couple of vf:RecipeFlow ?

Sure, hopefully tonight, but if not, tomorrow morning.  (I thought you were talking about recipes into plans.)

elf-pavlik:
Once we have that example we can see what we get out of it. Maybe you could choose an example of Recipe which used with another Recipe gives a Plan. Let's start with just one of them and in next step we can introduce the second one. I just want to see better how you see `vf:Recipe` and `vf:RecipePlan` differ and work together.

fosterlynn:
>Could you please quickly scribble an example of vf:Recipe with at least two vf:RecipeProcess and couple of vf:RecipeFlow ?

ResourceSpecification
--id: 'rs1-post'
--name: 'Post'

ResourceSpecification
--id: 'rs2-image'
--name: 'Image'

ResourceSpecification
--id: 'rs3-text'
--name: 'Text'

Recipe
-- id: 'recipe1'
-- name: 'Create Post'

RecipeProcess
-- id: 'recipeprocess1'
-- name: 'Write Post'

RecipeProcess
-- id: 'recipeprocess2'
-- name: 'Create Image'

RecipeProcess
-- id: 'recipeprocess3'
-- name: 'Format Post'

RecipeFlow
--id: 'recipeflow1'
--action: produce
--resourceConformsTo: rs1-post
--outputOf: recipe1, recipeprocess3  [or you could create two flows, but let's try this first]

RecipeFlow
--id: 'recipeflow2'
--action: produce
--resourceConformsTo: rs2-image
--outputOf: recipeprocess2

RecipeFlow
--id: 'recipeflow3'
--action: cite
--resourceConformsTo: rs2-image
--inputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow4'
--action: produce
--resourceConformsTo: rs3-text
--outputOf: recipeprocess1

RecipeFlow
--id: 'recipeflow5'
--action: cite
--resourceConformsTo: rs3-text
--inputOf: recipeprocess3

elf-pavlik:
Thank you! Why you can't use

RecipeProcess
-- id: 'recipe1'
-- name: 'Create Post'

?

elf-pavlik:
Also for posts that just use existing image. One would simply not use those

RecipeProcess
-- id: 'recipeprocess2'
-- name: 'Create Image'

RecipeFlow
--id: 'recipeflow2'
--action: produce
--resourceConformsTo: rs2-image
--outputOf: recipeprocess2

Also instead of ResourceSpecification `rs2-image` one could just use taxonomy category eg.
* http://www.wikidata.org/entity/Q182270  (raster graphics)
* http://www.wikidata.org/entity/Q170130  (vector graphics)

RecipeFlow
--id: 'recipeflow3'
--action: cite
--resourceClassifiedAs: http://www.wikidata.org/entity/Q182270
--inputOf: recipeprocess3

fosterlynn:
>--outputOf: recipe1, recipeprocess3 [or you could create two flows, but let's try this first]

Was thinking about this, and I think it is a mistake.  That is why I started with vf:deliverableOf or something like that.  I think it could make sense (if we even have process nesting, which I'm starting to wonder about) to have this:

--deliverableOf: recipe1 (since it isn't the output of recipe1 itself.  If recipe1 has an output of its own, it would be outputOf.)

Always good to run some data! :)

fosterlynn:
>Thank you! Why you can't you use
>RecipeProcess
>-- id: 'recipe1'
>-- name: 'Create Post'

I think we should think about that while thinking about nested processes in general.  NRP kind of works that way itself, since it doesn't have Recipe.  In a longer value flow, say you are making a component and then assembling it into something, there is an implied recipe for the whole thing, and also an implied recipe for making the component.

At this point, I think I should just remove Recipe from this PR and make that part of deciding if / how we want to do process nesting.  Then we'll have the minimum model that we know works, and some naming we can use.  And we can do more detailed data testing of the stuff we don't know if it works.

fosterlynn:
>Also for posts that just use existing image. One would simply not use those

Yeah, our experience with recipes is there is a sweet spot of generalization.  Recipes are for convenience as well as for sharing.  So you might create a whole recipe for a post, and when the plan is created, if you are using only existing images, you could just delete those from the plan.  If the whole thing was complicated, you might create another recipe that works only with existing images so you don't have to mess around so much with the plan.

elf-pavlik:
Could you please explain why you don't see Recipe having an output? We can also take example with process of making a lemon pie #290  which has two processes
* preparing unbaked pie
* baking pie

The main recipe (upper process) would have the baked pie as output and all ingredients as inputs, it would include two process recipes (lower) as listed above.
lemon pie process

When it comes to finding a recipe. One would most likely look for 
RecipeFlow
--id: 'recipeflow342423'
--action: produce
--resourceClassifiedAs: http://www.wikidata.org/entity/Q1550301 (Lemon meringue pie)
--outputOf: recipeprocess223423424

Since one aims for starting with intent: *I want to **produce** a **Lemon meringue pie***
Classification of recipies (recipie processes) themselves don't seem that important. The flow that you would like to use with `deliverableOf` seems the most important.

I can also imagine that economy can become more 'liquid' and agents will micro delegate processes in more dynamic way. One bakery having a big oven but not enough people available to work, could delegate part of *preparing unbaked pie* to another bakery. This way aiming for 5000 pies, they would prepare 2000 themselves, delegate preparing 3000 to another bakery and then bake all 5000 themselves. In that case the unbaked pie would make the *deliverable* for other bakery and they would only use that recipe process and don't care about the *bake pie* recipe process.

Just as we need a way to compose processes in very flexible way. I think we also need to handle recipes accordingly and avoid making them brittle (I do NOT say here that your proposals may lead to it!)

fosterlynn:
>Also instead of ResourceSpecification rs2-image one could just use taxonomy category 

Sure, some resource specifications won't need that much of a spec, and will be available on the web; or there may be a complete spec available in an existing taxonomy.  But I think we need something that provides one and only one identifiable kind for a resource, so economic activity around the resource can be coordinated.  If resourceClassifiedAs allows multiples, and they could be any kinds of classification that could be used to help people match an intent say, then it doesn't work for actually making sure you have the right thing. 

Maybe resourceConformsTo could refer to http://www.wikidata.org/entity/Q182270 as its ResourceSpecification.  And resourceClassifiedAs would be used for any and all classifications that are useful for whatever.  Like resourceConformsTo is what the resource or potential resource "is", and resourceClassifiedAs is how it can be "classified".  In other words, the different meanings are embodied in the properties, not where the reference resides.  Will that work?

fosterlynn:
>When it comes to finding a recipe. One would most likely look for RecipeFlow

Yes, the RecipeFlow for the output is the way to find a specific recipe.  It would usually be 1:1 with Recipe, except it is m:1 if there are co-products of the same recipe.

(Maybe in the UI, people would look for a ResourceSpecification (NRP works this way), or starting more with classifications if their data sources were more broad.  For example "I want to produce a Lemon meringue pie" might match to a lot of recipes.  Even an intent to produce a very exactly specified lemon meringue pie with exactly the same ingredients might have more than one way to create it based on available equipment, best work flow for the facility, etc.)

>Could you please explain why you don't see Recipe having an output?

I'm seeing the meaning of vf:outputOf as meaning it has its own output.  It could derive its output(s) from the nested processes as needed.  I can see some possible examples of nesting processes that would have their own output.  But.... I"m straying into https://github.com/valueflows/valueflows/pull/293.... and I still have a bunch of questions about that to work through.

>I can also imagine that economy can become more 'liquid' and agents will micro delegate processes in more dynamic way.

:+1: :heart:

fosterlynn:
I removed Recipe from this PR, let's see what that gives us in terms of remaining work to do for this.  I'll also start an issue to complete recipe.

release-doc-in-process/all_vf.TTL
```diff
@@ -30,6 +30,23 @@ vf:AgentRelationshipRole  a  owl:Class ;
...
+        rdfs:label         "vf:ResourceSpecification" ;
+        rdfs:comment       "." .
+
+vf:RecipeProcess a          owl:Class ;
+        rdfs:label         "vf:RecipeProcess" ;
+        rdfs:comment       "." .
+
+vf:RecipeFlow a            owl:Class ;
+        rdfs:label         "vf:RecipeFlow" ;
+        rdfs:comment       "The linkage between resource types, process types, and actions that structure a recipe." .
```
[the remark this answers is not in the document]
fosterlynn: Thanks, will fix.  But I think that resource specifications could include types of work and services, whether they specify a material resource or not.  All potential economic resources that could be imagined in a recipe must be described by a resource specification I think.  Maybe the resource specification only has a reference to one resource classification, that could work.

Changed to: "The linkage between a process recipe, an action that structures a recipe, and a resource specification."  What do you think?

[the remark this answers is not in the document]
elf-pavlik: Let's leave it like that for now. I would not use `ResourceSpecification` since we don't reference any resource from *work* flows and this would just make everything more confusing.
Only flows which at event stage will reference an actual resource with *affects* should use `vf:resourceConformsTo` and reference specification of a resource which event will reference using `vf:affects`. I'll create another issue about categorizing ans specifying work and services.

fosterlynn:
>Only flows which at event stage will reference an actual resource with affects should use vf:resourceConformsTo and reference specification of a resource which event will reference using vf:affects. I'll create another issue about categorizing ans specifying work and services.

Moving this discussion out to the main line, since it is more fundamental.  I disagree, every flow in a recipe should have vf:resourceConformsTo, or it basically won't work.  So that has to include work, services, whatever kind of things will be referenced from a RecipeFlow in the resource area.  I looked at your new issue, which doesn't seem to affect this question, and seems like it can be considered separately.  Seems to affect more which actions are used.

Anyhow, I think everything needs to be "specified", and the pattern has to remain the same, that is the beauty of this model.  Just like we don't care if people want to trade for currency or work/services or material items, for example.  Also some services (or types of work/skills) are quite specific, for example.  But everything needs to "specified" (not so much classified, although that can be done too) on the level that makes sense, even if quite general. 

Also, from a long time ago, we decided to always have an EconomicResource on the observation layer, even if it is one of those non-material things, and only references the resourceConformsTo.  If there is a problem with non-material resources, it lies there.  I'm fine to go back to giving the event the capability to reference an EconomicResource or a ResourceSpecification.

release-doc-in-process/all_vf.TTL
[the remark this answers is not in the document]
fosterlynn: I put it back and changed it to be singular.  Also noticed a couple non-definitions, and added those.  Please check them because you may disagree.  I'm trying to clarify based on this discussion.

release-doc-in-process/all_vf.TTL
```diff
@@ -30,6 +30,23 @@ vf:AgentRelationshipRole  a  owl:Class ;
...
+        rdfs:label         "vf:ResourceSpecification" ;
+        rdfs:comment       "Specification of a kind of resource. Could define a material item, service, digital item, currency account, etc." .
+
+vf:RecipeProcess a          owl:Class ;
+        rdfs:label         "vf:RecipeProcess" ;
+        rdfs:comment       "Specifies the process part of a recipe for use in planning from recipe." .
+
+vf:RecipeFlow a            owl:Class ;
+        rdfs:label         "vf:RecipeFlow" ;
+        rdfs:comment       "The linkage between a resource type, process type, and action that create the structure of a recipe." .
```
elf-pavlik: once again, we don't have *resource type* but *resource specification*, similar *recipe process* not *process type*

elf-pavlik:
Could you please amend that fully automated post creation example to add some *work* flows?
We can also consider variant where formatting gets delegated to another agent who will provide it asa  service, and use specific software tool in that process and particular laptop (with serial number).

release-doc-in-process/all_vf.TTL
fosterlynn: I thought you wanted me to put it back, per the comment above.  Tell me the description you would like there please, I'm confused.

elf-pavlik: Sorry, I meant to leave it the way you had it changed:
*"The linkage between a process recipe, an action that structures a recipe, and a resource specification."*

for now we will not say anything about some flow recipes possibly will not have reference to a resource specification.

fosterlynn: Thanks, done.

fosterlynn:
>Could you please amend that fully automated post creation example to add some work flows?

ResourceSpecification
--id: 'rs1-post'
--name: 'Post'

ResourceSpecification
--id: 'rs2-image'
--name: 'Image'

ResourceSpecification
--id: 'rs3-text'
--name: 'Text'

ResourceSpecification
--id: 'rs4-designwork'
--name: 'Graphic Design'

ResourceSpecification
--id: 'rs5-writingwork'
--name: 'Writing (EN)'

RecipeProcess
-- id: 'recipeprocess1'
-- name: 'Write Post'

RecipeProcess
-- id: 'recipeprocess2'
-- name: 'Create Image'

RecipeProcess
-- id: 'recipeprocess3'
-- name: 'Format Post'

RecipeFlow
--id: 'recipeflow1'
--action: produce
--resourceConformsTo: rs1-post
--outputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow2'
--action: produce
--resourceConformsTo: rs2-image
--outputOf: recipeprocess2

RecipeFlow
--id: 'recipeflow3'
--action: cite
--resourceConformsTo: rs2-image
--inputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow4'
--action: produce
--resourceConformsTo: rs3-text
--outputOf: recipeprocess1

RecipeFlow
--id: 'recipeflow5'
--action: cite
--resourceConformsTo: rs3-text
--inputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow6'
--action: work
--resourceConformsTo: rs4-designwork
--inputOf: recipeprocess2

RecipeFlow
--id: 'recipeflow7'
--action: work
--resourceConformsTo: rs5-writingwork
--inputOf: recipeprocess1

fosterlynn:
>We can also consider variant where formatting gets delegated to another agent who will provide it as a service, and use specific software tool in that process and particular laptop (with serial number).

Yes, for sure, any action should be able to be in a recipe.  

However, particular laptop with serial number isn't something I see happening in a recipe.  For sure not useful in a recipe published openly using VF.  Only ResourceSpecifications, not EconomicResources.  Seem right to you?

There are a lot of details involved in creating a plan from recipe, that I think @bhaugen will be writing up at some point.  Like when can you plan the whole way back on the value chain, and when would you just delegate/source the item from another agent, who would use that part of the recipe to create a plan.  Part of this is about crossing scopes, and previous agreements between scope agents.

PR #301, opened 2018-12-01
exchange rates

elf-pavlik:
This PR proposes `vf:ExchangeRate` which allows referencing two flows together with values to calculate exchange rate. Both units come from flows. For example:

```yaml
`@context': ...
`@graph`:

  - `@id`: ex:an-agreement
    `@type`: Agreement

  - `@id`: ex:work-event
    `@type`: EconomicEvent
    under: ex:an-agreement
    action: work
    quantity:
      'qudt:numericValue': 3
      `qudt:unit`: unit:Hour

  - `@id`: ex:use-event
    `@type`: EconomicEvent
    under: ex:an-agreement
    action: use
    affects: ex:some-tool
    quantity:
      'qudt:numericValue': 2
      `qudt:unit`: unit:Hour

  - `@id`: ex:issue-event
    `@type`: EconomicEvent
    under: ex:an-agreement
    action: issue
    affectedClassifiedAs: http://www.wikidata.org/entity/Q618355 # apple juice
    quantity:
      'qudt:numericValue': 20
      `qudt:unit`: unit:Liter 

  - `@id`: ex:an-exchange-rate
    `@type`: ExchangeRate
    under: ex:an-agreement
    numerator: 5
    denominator: 1
    numeratorFlow: ex:issue-event
    denominatorFlow: ex:work-event

  - `@id`: ex:another-exchange-rate
    `@type`: ExchangeRate
    under: ex:an-agreement
    numerator: 2.5
    denominator: 1
    numeratorFlow: ex:issue-event
    denominatorFlow: ex:use-event
```

One can use them with `vf:Commitment` in the same way. I'll try to add an example of a *Proposal* which uses exchange rate together with two intents.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -71,6 +71,10 @@ vf:DistributionAgreement  a       owl:Class ;
         rdfs:subClassOf   vf:Agreement ;
         rdfs:comment      "An agreement to distribute some resource in exchange for inputs or outputs created by other agents." .
 
+vf:ExchangeRate a owl:Class ;
+        rdfs:label        "vf:ExchangeAgreement" ;
+        rdfs:comment      "Used with exchange agreements to represante an exchange rate between two reciprocal flows" .
+
 # OBSERVATION CLASSES
 
 
@@ -259,6 +263,30 @@ vf:under  a          owl:ObjectProperty ;
         rdfs:label   "under" ;
         rdfs:range   vf:Agreement ;
         rdfs:comment        "Reference an agreement between agents which specifies the rules or policies which govern this event." .
+
+vf:numeratorFlow a   owl:ObjectProperty ;
+        rdfs:label   "vf:numeratorFlow" ;
+        rdfs:domain  vf:ExchangeRate ;
+        rdfs:range   [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent ) ] ;
+        rdfs:comment "References flow related to the numerator value" .
+
+vf:denominatorFlow a   owl:ObjectProperty ;
+        rdfs:label   "vf:denominatorFlow" ;
+        rdfs:domain  vf:ExchangeRate ;
+        rdfs:range   [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent ) ] ;
+        rdfs:comment "References flow related to the denominator value" .
+
+vf:numerator a   owl:DataProperty ;
+        rdfs:label   "vf:numerator" ;
+        rdfs:domain  vf:ExchangeRate ;
+        rdfs:range  xsd:decimal ;
+        rdfs:comment "Numerator value" .
+
+vf:denominator a   owl:DataProperty ;
+        rdfs:label   "vf:denominator" ;
+        rdfs:domain  vf:ExchangeRate ;
+        rdfs:range  xsd:decimal ;
+        rdfs:comment "Denominator value" .
 
 vf:createdBy  a        owl:ObjectProperty ;
         rdfs:label   "created by" ;
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ npm-debug.log
 
 .build
 *~
+package-lock.json
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "mdast": "^2.1.0",
     "mdast-autolink-headings": "^1.0.0",
     "mdast-html": "^1.2.1",
-    "mdast-slug": "^2.0.0"
+    "mdast-slug": "^2.0.0",
+    "turtle-validator": "^1.0.2"
   }
 }


release-doc-in-process/all_vf.TTL
```diff
@@ -71,6 +71,10 @@ vf:DistributionAgreement  a       owl:Class ;
         rdfs:subClassOf   vf:Agreement ;
         rdfs:comment      "An agreement to distribute some resource in exchange for inputs or outputs created by other agents." .
 
+vf:ExchangeRate a owl:Class ;
+        rdfs:label        "vf:ExchangeAgreement" ;
```
fosterlynn: label copy/paste?

release-doc-in-process/all_vf.TTL
```diff
@@ -260,6 +264,30 @@ vf:under  a          owl:ObjectProperty ;
...
+        rdfs:domain  xsd:decimal ;
+        rdfs:range   [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent ) ] ;
+        rdfs:comment "Numerator value" .
+
+vf:denominator a   owl:DataProperty ;
+        rdfs:label   "vf:denominator" ;
+        rdfs:domain  xsd:decimal ;
+        rdfs:range   [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent ) ] ;
+        rdfs:comment "Denominator value" .
+
```
fosterlynn: Not sure I understand the model.  Are the domain and range what you intend?

fosterlynn:
Thanks @elf-pavlik !

Can you give a couple scenarios where this would be used?  And what its primary purpose is?

release-doc-in-process/all_vf.TTL
elf-pavlik: I think we did it because WebVOWL uses those labels on diagrams

fosterlynn: but do you want to use ExchangeAgreement for ExchangeRate?

release-doc-in-process/all_vf.TTL
elf-pavlik: Good catch! I've just fixed it...

elf-pavlik:
In the example above, without exchange rates one would not see exactly how reciprocity breaks down across all the events. If someone wants to model nowadays common receipt, one would have many cart items on one side and single currency payment on the other. Without exchange rates I don't think we can capture how much someone paid for given quantity of each cart item.

I also think this will come very useful for proposals. Agent can have a proposal with a pair of intents, which would result in pair of reciprocal events. Using exchange rates we can express how expected quantities should line up. The intents could stay 'upper' intents as we have in #293 so more granular intents in planning would reference them with `vf:includedIn`. So the pair intents referenced from proposals would have arbitrary quantities which have nothing to do with exchange rate expected in proposed exchange. Proposal can include an ExchangeRate to provide explicit information with ratio of expected quantities of commitments and later events.

elf-pavlik:
```yaml
`@context': ...
`@graph`:

  - `@id`: ex:alice
    `@type`: foaf:Person

  - `@id`: ex:work-intent
    `@type`: Intent
    provider: ex:alice
    action: work
    quantity:
      'qudt:numericValue': 100
      `qudt:unit`: unit:Hour

  - `@id`: ex:use-intent
    `@type`: Intent
    provider: ex:alice
    action: use
    affects: ex:some-tool
    quantity:
      'qudt:numericValue': 200
      `qudt:unit`: unit:Hour

  - `@id`: ex:issue-intent
    `@type`: Intent
    recipient: ex:alice 
    action: issue
    affectedClassifiedAs: http://www.wikidata.org/entity/Q618355 # apple juice
    quantity:
      'qudt:numericValue': 30
      `qudt:unit`: unit:Liter 

  - `@id`: ex:an-exchange-rate
    `@type`: ExchangeRate
    numerator: 5
    denominator: 1
    numeratorFlow: ex:issue-intent
    denominatorFlow: ex:work-intent

  - `@id`: ex:another-exchange-rate
    `@type`: ExchangeRate
    numerator: 2.5
    denominator: 1
    numeratorFlow: ex:issue-intent
    denominatorFlow: ex:use-intent

  - `@id`: ex:a-proposal
    `@type`: Proposal
    eligibleIntent:
      - ex:work-intent
      - ex:issue-intent
    applicableRate: ex:an-exchange-rate

  - `@id`: ex:another-proposal
    `@type`: Proposal
    eligibleIntent:
      - ex:use-intent
      - ex:issue-intent
    applicableRate: ex:another-exchange-rate
```

actually here we may want to use `vf:RecipeFlow` besides `vf:Intent` for more reusable exchange rates.

```yaml
`@context': ...
`@graph`:

  - `@id`: ex:alice
    `@type`: foaf:Person

  - `@id`: ex:work-intent
    `@type`: Intent
    provider: ex:alice
    action: work
    quantity:
      'qudt:numericValue': 100
      `qudt:unit`: unit:Hour

 - `@id`: ex:work-recipe
    `@type`: RecipeFlow
    action: work
    quantity:
      'qudt:numericValue': 1
      `qudt:unit`: unit:Hour

  - `@id`: ex:use-intent
    `@type`: Intent
    provider: ex:alice
    action: use
    affects: ex:some-tool
    quantity:
      'qudt:numericValue': 200
      `qudt:unit`: unit:Hour

  - `@id`: ex:use-recipe
    `@type`: RecipeFlow
    action: use
    affects: ex:some-tool
    quantity:
      'qudt:numericValue': 200
      `qudt:unit`: unit:Hour

  - `@id`: ex:issue-intent
    `@type`: Intent
    action: issue
    recipient: ex:alice
    affectedClassifiedAs: http://www.wikidata.org/entity/Q618355 # apple juice
    quantity:
      'qudt:numericValue': 30
      `qudt:unit`: unit:Liter 

  - `@id`: ex:issue-recipe
    `@type`: RecipeFlow
    action: issue
    affectedClassifiedAs: http://www.wikidata.org/entity/Q618355 # apple juice
    quantity:
      'qudt:numericValue': 1
      `qudt:unit`: unit:Liter 

  - `@id`: ex:an-exchange-rate
    `@type`: ExchangeRate
    numerator: 5
    denominator: 1
    numeratorFlow: ex:issue-recipe
    denominatorFlow: ex:work-recipe

  - `@id`: ex:another-exchange-rate
    `@type`: ExchangeRate
    numerator: 2.5
    denominator: 1
    numeratorFlow: ex:issue-recipe
    denominatorFlow: ex:use-recipe

  - `@id`: ex:a-proposal
    `@type`: Proposal
    eligibleIntent:
      - ex:work-intent
      - ex:issue-intent
    applicableRate: ex:an-exchange-rate

  - `@id`: ex:another-proposal
    `@type`: Proposal
    eligibleIntent:
      - ex:use-intent
      - ex:issue-intent
    applicableRate: ex:another-exchange-rate
```

I think for Proposal we still need to consider #266 and how Agreements get created based on proposals. Since exchange rates just needs classifications/specifications and units from flows, they should work with any phase of flow as well as flow recipe.

fosterlynn:
In terms of exchange rates, wouldn't it be better to use quantities with units instead of ratios?  I can't think of anything where an exchange rate isn't expressed in terms of some base quantities and units, often tied to packaging, but often just a standard measurement - the ratio would always be 1:1.  And if not tied to specific intents, then it can be used for lots of intents, including for different agents.

elf-pavlik:
What do you think about this change to naming?
```yaml
  - `@id`: ex:an-exchange-rate
    `@type`: ExchangeRate
    reciprocalQuantity: 5
    primaryQuantity: 1
    reciprocal:Flow: ex:issue-event
    primaryFlow: ex:work-event
```

This also relates to #307 and #302

fosterlynn:
I understand more how you are seeing it.  Your intents in a proposal are not already proportional, while I assumed they were.  I would be interested in how this model would support simpler use cases, where the agents' overall estimates for themselves are not known (or not public), like maybe:
* proposal to trade my old bike for a bushel of apples
* proposal to sell 25# case of carrots for $25 (assumes will have plenty of carrots for the expected requests - this is basically the exchange rate, per [this price list](https://github.com/valueflows/valueflows/pull/302#issuecomment-447595739)

I'll do the same for my [latest model](https://github.com/valueflows/valueflows/pull/302#issuecomment-447604351), because I'm not at all convinced it works for everything.

fosterlynn:
>I would consider 'plenty' as very subjective, someone planning food for event with 1000 participants would see it differently than a family planning food for week.

>Still, one could still probably make vary vague Proposal which only includes 'exchange rates' / 'reciprocity ratios'. 

It isn't really subjective, because this is a wholesale situation, they sell only to known agents, with fairly consistent orders (co-ops, buying clubs, local restaurants, etc.).  I expect there is conversation if something comes up that is unusual.

fosterlynn:
>This way we only have flow patterns and reciprocity ratios. 

Not proposing anything yet, just documenting what it would look like to have a self contained exchange rate / ratio.  In contrast to breaking it apart as above.

```yaml
`@id`: ex:work-veg-reciprocity-ratio
`@type`: ReciprocityRatio
primaryQuantity {
     numericValue: 1
     unit: Each
}
primaryAction: give
primaryResourceClassifiedBy: Mixed veggie box
reciprocalQuantity: {
     numericValue: 3
     unit: Hour
}
reciprocalAction: work
reciprocalResourceClassifiedBy: Harvesting
```

Or, taking it farther into an agreement:
```yaml
`@id`: ex:SFF-work-veg-reciprocity-ratio-agreement
`@type`: ReciprocityRatioAgreement
primaryQuantity {
     numericValue: 1
     unit: Each
}
primaryAction: give
primaryResourceClassifiedBy: Mixed veggie box
provider: Small Family Farm
reciprocalQuantity: {
     numericValue: 3
     unit: Hour
}
reciprocalAction: work
reciprocalResourceClassifiedBy: Harvesting
provider: Alice
```

elf-pavlik:
I find this smoothing, quite awkward! Let's compare it with templates + patterns based approach discussed in #266 
BTW you also would need *primaryProvider* and *reciprocalProvider* instead of *provider* in your snippets above

```yaml
`@graph`:
  - `@id`: ex:work-veg-agreement-template
    `@type`: AgreementTemplate

  - `@id`: ex:veg-flow-template
    `@type`: FlowTemplate
    underTemplate: ex:work-veg-agreement-template
    provider: Small Family Farm
    action: give
    resourceClassifiedAs: Mixed veggie box
    quantity:
     numericValue: 1
     unit: Each

  - `@id`: ex:veg-flow-template
    `@type`: FlowTemplate
    underTemplate: ex:work-veg-agreement-template
    recipient: Small Family Farm
    action: work
    workClassifiedAs: Harvesting
    quantity:
      numericValue: 3
      unit: Hour
```

It has advantage of not having those duplicate *primary* & *reciprocal* version for *quantity*, *unit*, *resourceClassifiedAs*, *action*, *provider* and *recipient*.

It also doens't have limitation of a pair (primary & reciprocal) but can work with triple etc. as in:
* receive work
* provide meals
* provide accommodation

https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#work-exchange

fosterlynn:
>BTW you also would need primaryProvider and reciprocalProvider instead of provider in your snippets above

thanks, fixed

elf-pavlik:
I think we can already eliminate this approach with single construct with represents a rate / ratio between a pair. Approach that uses combinations of constructs where some of them look like flows (eg. Flow Templates) seems much more flexible and align around flows.

PR #302, opened 2018-12-09
Proposal

fosterlynn:
Basic Intent, with relationships to Proposal.


fosterlynn:
Note I just added a couple things to this PR.  I'm happy to remove anything that needs more discussion, since the intent of the PR is to get the basic minimum here for Intents.

release-doc-in-process/all_vf.TTL
```diff
@@ -427,6 +441,47 @@ vf:resourceConformsTo
         rdfs:range      vf:ResourceSpecification ;
         rdfs:comment    "The primary resource knowledge specification or definition of an existing or potential resource." .
 
+vf:intents
```
elf-pavlik: I think we should name property based on meaning of the relationship, not the type of the object. Also using plural stays consider as anti pattern. Any property can take a single value or an array (speaking bit JSON-LD specific way)
Some options
* *eligibleIntent*
* *applicableTo*

release-doc-in-process/all_vf.TTL
```diff
@@ -427,6 +441,47 @@ vf:resourceConformsTo
...
+        a                   owl:ObjectProperty ;
+        rdfs:label         "intents" ;
+        rdfs:domain         vf:Proposal ;
+        rdfs:range          vf:Intent ;
+        rdfs:comment        "The primary offers, requests, estimates, or other intents proposed." .
+
+vf:reciprocalIntents
```
elf-pavlik: In what cases we can not rely on *provider* and *recipient* relations from intents that proposal applies to? For proposals that should lead to exchange, one intent would specify agent with *provider* and another one with *recipient*. Practically in such proposals an agent 'offers' and 'requests' something at the same time.

release-doc-in-process/all_vf.TTL
```diff
@@ -427,6 +441,47 @@ vf:resourceConformsTo
...
+        a                   owl:ObjectProperty ;
+        rdfs:label         "reciprocal intents" ;
+        rdfs:domain         vf:Proposal ;
+        rdfs:range          vf:Intent ;
+        rdfs:comment        "The reciprocal offers, requests, estimates, or other intents proposed." .
+
+vf:intentOf
```
elf-pavlik: inverse properties just complicate queries, one has to always query for both directions to make sure all results get returned. if we let any inverse property slip into the vocab we should remove it as part of next release cleanup.
on can always adjust position of object and subject in the statement without need to have inverse property, in JSON-LD `@reverse` keyword provides some convenience.

release-doc-in-process/all_vf.TTL
```diff
@@ -427,6 +441,47 @@ vf:resourceConformsTo
...
+        a                   owl:ObjectProperty ;
+        rdfs:label         "intent of" ;
+        rdfs:domain         vf:Intent ;
+        rdfs:range          vf:Proposal ;
+        rdfs:comment        "The proposal which this intent applies to." .
+
+vf:reciprocalIntentOf
```
elf-pavlik: same as above comment above inverse properties

release-doc-in-process/all_vf.TTL
fosterlynn: I don't think provider and receiver have anything to do with which intent is reciprocal (if any).  Intent software I have seen (as well as @matslats proposed layout) has a choice for is this proposal an offer or a request.  Since you did not want that, you can only tell based on what is reciprocal.  I don't know the percentages, but I think a fairly large percentage of people will not want their reciprocal intent to imply they want that intent to be matched.

fosterlynn: If an agent does want to offer and request something at the same time, then neither on is "reciprocal" and they don't use that.

elf-pavlik: Can you provide an example of what you would consider a Proposal
* that references two Intents
* aims at providing a path to an Exchange Agreement
* out of those two referenced intents, one specifiers *provider* and the other one DOES NOT specify *recipient*

Just to make sure that we don't have misunderstanding here. Agent creates all the Intents first during initial phase of planning. Later on Agent creates various Proposals, where each Intent can get referenced from any number of different Proposals. So we don't have Intents specific to a Proposal, Intents just express initial plan of given agent, which later will lead to commitments and events.

release-doc-in-process/all_vf.TTL
fosterlynn: removed this one anyhow - but I'll try! (seems there is a sweet spot between just repeating the class and something that has to be looked up to see what it is)

release-doc-in-process/all_vf.TTL
fosterlynn: >out of those two referenced intents, one specifiers provider and the other one DOES NOT specify recipient

I don't understand the implications of this.  If they are intended to be reciprocal, then one would specify provider and the other recipient.  But that doesn't tell you which one is reciprocal.

>Just to make sure that we don't have misunderstanding here. Agent creates all the Intents first during initial phase of planning. Later on Agent creates various Proposals, where each Intent can get referenced from any number of different Proposals. So we don't have Intents specific to a Proposal, Intents just express initial plan of given agent, which later will lead to commitments and events.

I don't think there is a misunderstanding, I agree that is a valid scenario.  There is also a scenario where people just want to propose exchanges from inventory.  For example, distributors, aggregators, just individual people with something to get rid of or that they want.

elf-pavlik: > If they are intended to be reciprocal, then one would specify provider and the other recipient. But that doesn't tell you which one is reciprocal.

I think Commitments made based on those Intens, will stay reciprocal to one another. I would see *reciprocity* as symmetric relation. Having an Exchange Agreement with two Commitments
* provider: Alice, recipient: Bob, work: 2h, under: an-agreement
* provider: Bob, recipient: Alice, use: 8h, under: an-agreement

I wouldn't see only one of them as reciprocal to the other one. Once again Commitments in Exchange Agreement seem to me mutually reciprocal.

> There is also a scenario where people just want to propose exchanges from inventory. For example, distributors, aggregators, just individual people with something to get rid of or that they want.

In that case they can create Intent to *issue* that something which references them as *provider*. Then they can reference that Intent in any number of Proposals, which if the aim at Exchange Agreement, would each also reference another Intent which references that agent as *recipient*.

elf-pavlik:
Since many Proposals will reference the same intent, I think it makes more sense that we set the direction from Proposal to Intent. I would propose those amendments to this PR
1. remove `intentOf` and `reciprocalInentOf`
2. find fitting naming for proposed in this PR `intents` (singular one!) eg. `vf:eligibleIntent` 
3. leave reciprocal/primary/main issue for separate PR, it may affect Commitment and/or Agreement as well - if you want it to affect how UI presents your transactions, 'orders' already made will have commitments and not Intents any more, just having something on a proposal may not provide a best way to affect how transaction gets presented later on. 

As we discussed in a call, matching applications should consider all *eligible Intents* as **and** and response to proposal should lead to an Exchange Agreement which has Commitment satisfying every Intent according to 'exchange rate' / 'agreement recipe'. It may make sense to capture it as dedicated issue, and take into account when we document matching.

elf-pavlik:
Quick observation which doesn't affect this PR

I haven't seen in snippets we have shared so far something that seems a common example. A single instance of `vf:Commitment` will often satisfy **a pair** of `vf:Intent` instances.
* one where Alice wants to act as *provider*
* one where Bob wants to act as *recipient*

Those *matching* intents will often differ in quantities but `vf:Satisfaction` already takes care of that. So commonly for each instance `vf:Commitment` we will have two instances `vf:Satisfaction`, each for one of the matching intents. 

This also relates to https://github.com/valueflows/intent/issues/7
BTW in vientos, when someone finds a Proposal with Intent(s) that one would like to save as 'match' but not start conversation yet. One will either select the matching Proposal with Intent(s) or have chance to semi-automatically have such matching proposal created. Later when one finds another 'match', one will have it also referencing that existing or newly created proposal, it aims to have clear way to show for each Intent all the Proposals with discovered 'matches'.

elf-pavlik:
Replying to https://github.com/valueflows/valueflows/issues/217#issuecomment-447584464

>> Once it gets fully satisfied, all those 10 or more proposals should get considered 'inactive' since agent already made all the Commitments on that Intent they expected.

> I guess that would depend on the situation, for example people crowdfunding often are happy when they get more than they asked. :) I'd still like to have vf:isFinished optional property for Intents, as well as Commitments and Processes. An app could choose to handle it either way, based on its requirements.

Possibly in separate PR I would like to propose `vf:minimalQuantity` and `vf:maximalQuantity` in addition to `vf:intendedQuantity` which means *optimal*.

At the same time since the beginning I stay rather :-1: on `vf:isFinished` in favor of updating specific attributes of Intent based on changes in planning. Having *minimum* and *maximum* one would always have *satisfied* somewhere between and still have the *optimal* (just `vf:intendedQuantity`) as a reference.

fosterlynn:
>Possibly in separate PR I would like to propose vf:minimumQuantity and vf:maximumQuantity in addition to vf:intendedQuantity which means optimal.

>At the same time since the beginning I stay rather -1 on vf:isFinished in favor of updating specific attributes of Intent based on changes in planning. Having minimum and maximum one would always have satisfied somewhere between and still have the optimal (just vf:intendedQuantity) as a reference.


vf:minimumQuantity and vf:maximumQuantity could be useful, I'm good with them.  But I don't see that they help with knowing when something is finished.  People shouldn't really update plans to match reality, then they don't know how well they planned, and there is no reason to have both.  There is no algorithm I can think of that can determine when an Intent or Commitment is finished.  When they are for work, that is especially true, nobody can estimate that well, unless people are just punching a clock and it is irrelevant to what is produced.  Processes are a bit easier, but still not fool-proof.

elf-pavlik:
>  I don't think a Satisfaction would be created until a Commitment or EconomicEvent is created.

:+1: 

> You could find matches and just keep track of them some way not involving VF if you want them saved in the app in some way for later display?

I think eventually VF should provide a way to model matches. Sometimes one may not rely purely on taxonomical relations between classifications and may need a way to represent verified matching. Still this can wait.

> Another option would be to just have a property on Intent something like vf:match (boolean) - or pleaseMatch :) - or something better.... or better yet, vf:doNotMatch... or.... ?

During our call you brought up two different requirements
* how UI presents something
* how matching applications should and should not behave

I would prefer that we first document those requirements and support them with some use cases. Based on that we can evaluate possible approaches to addressing them. It really makes sense to me not to try doing it in this PR

> People shouldn't really update plans to match reality, then they don't know how well they planned, and there is no reason to have both. There is no algorithm I can think of that can determine when an Intent or Commitment is finished.

Makes sense, I think we can also document that requirement and back it with some use cases.

> Good point, this is now a M:M. I'll need to introduce something in the middle, at least for the UML people. Do we need any data there? Maybe not in this case, will have to think. If not, maybe we don't need that for rdf. Hmmm.

We could possibly qualify that relation between Proposal and Intent and use it for that 'exchange rate' information. Let's think if we see if we find any cases where we would want to reuse same exchange rate among multiple proposals. It might happen when Intents reference specific resource and not just classification, here same exchange rate would apply to different Proposals. We also might find different way to model that and have 'upper' Intent that intent for each specific resource would reference with `vf:includedIn` and then one Proposal would just reference that 'upper' Intent and we would not need multiple Proposals 'reusing' an exchange rate.

Let's explore qualifying that relation (between Proposal and Intent) as another alternative to 'exchange rates' and 'agreement recipes'? We would still need to check that use case with apple juice for work and usage, both #301 and #266 have variant of it which also shows next step - Agreement with Commitments. I think we need a way that Agreement includes information about 'exchange rates' / 'agreement recipes' used. So when you see an agreement with commitments and later recorded events, you can see breakdown
* how much juice for how much work
* how much juice for how much usage
Having something like 'exchange agreement' or 'agreement recipe' could allow just referencing that from agreement (possibly versioned snapshot). Qualified relations between Proposal and Intent would also need a way that would propagate into agreement with commitments and later events.

fosterlynn:
Here's a draft of what the M:M between Proposal and Intent might look like.  Need to do some scenarios with data, though.
prop-int

And vf:isFinished on Intent too.  (But don't worry if you see things that we haven't agreed to, I'm just thinking ahead a bit on the model.)

(moved this from here, getting tooooooo long #217 )

fosterlynn:
>Here's a draft of what the M:M between Proposal and Intent might look like.

Hmmm, might imply a M:M between Intent and Process too.  But that is dependent on having more kinds of processes (estimates, plans, etc.), which we don't have yet.

elf-pavlik:
:+1: proposed Intent qualified relation, I think we need something that expresses 'exchange rate' that we can also reuse in an Agreement with commitments and/or Events. #307 might clarify that if I understood it correctly.

> [edit] I think we need something for this PR, otherwise there is a big hole. And it might affect the way it fits together in a core model kind of way.

Would you agree that this 'primary' / 'secondary' emphasis will need to stay available for Agreements as well? Since proposals can change, we may need a way that Agreement has its own 'copy' of that information, just as it will need its own 'copy' of whatever we use to represent exchange rates.

fosterlynn:
>Would you agree that this 'primary' / 'secondary' emphasis will need to stay available for Agreements as well? 

Not as critical because then it is agreed to.  But still useful.  (And I'm thinking of ExchangeAgreements, probably not applicable to a lot of other kinds of Agreements.)

fosterlynn:
Looking at what is the same and what is different in proposed use cases, some contrasts:
* estimates for planning purposes vs. scheduled or one-off with exact quantities
* exchange rate / flow template / price list vs. actually intended quantities
* recurring vs. one-time
* multi-agent coordination vs simple proposal for 2 agents (at a time)

The basic data at the flow level:
-- action
-- provider or recipient
-- resource or resource classification(s) or resource specification
-- quantity (numericValue, unit) - this has some different permutations
-- time - this has some different permutations

This could be a:
* flow template
    * quantity is the exchange rate quantity
    * time not relevant (except for how long it is valid for, which can be in the object that contains the flow templates)
* estimate (budget, estimated plan for multi-agent coordination, etc.)
    * quantity is the estimated quantity for the whole time period, not by individual future occurrence
    * time is the overall time period for the estimate (could also be carried at the object that contains this estimate if all will be the same, like a yearly budget)
* specific offer, request
    * quantity is the quantity offered or requested
    * time is the scheduled time (if applicable, such as for a scheduled process); sometimes frequency might be part of this?

What else am I missing?

elf-pavlik:
> specific offer, request

I understand by itself it has no notion of exchange or reciprocity in general.
As for the meanings
* offer meaning "to provide"
* request meaning "to receive"

> flow template
>* quantity is the exchange rate quantity
>* time not relevant (except for how long it is valid for, which can be in the object that contains the flow templates)

I think *the object that contains the flow templates* needs to stay responsible for any validity period since we consider them all with **and**

> estimate (budget, estimated plan for multi-agent coordination, etc.)

I think here we would have `vf:includedIn` relations from the flows in what you described as *specific offer, request*

> multi-agent coordination vs simple proposal for 2 agents (at a time)

In case where Alice publishes an exchange proposal:
* to receive apple juice
* to provide work

I don't see need to make assumption that in the agreement the same agent will both
* provide apple juice
* receive work

Commitments in agreement where
* Alice receives juice from Bob
* Alice provides work to Claudia
* Claudia provides usage of something to Bob

Would still satisfy Alice's intents to *provide* and *receive* from her proposal. At the same time would also satisfy *provide* and *receive* pairs from two other proposals published by Bob and Claudia.

This means that agents may find multi party agreement even if they all just started with expressing possible pair of flow templates in their exchange proposals. From what I remember [Makerfox](http://wiki.p2pfoundation.net/Makerfox) forced agents to always have the same currency in one of those flow templates (precisely whatever equivalent its model used, most likely just *price*). I see possibilities for more general algorithms which would only rely on the ratios between those pairs and resource classifications.

fosterlynn:
>This means that agents may find multi party agreement even if they all just started with expressing possible pair of flow templates in their exchange proposals.

True.  Maybe multi-party isn't the right way to express it. I said "estimated plan for multi-agent coordination" - but still your point is well taken.  Picturing something like LocEcon or people figuring out a barter chain, possibly ongoing.  Looking for what is different enough to perhaps name differently, or call out in some way. Or deciding they can all be "proposals" if there is more alike than different.

>specific offer, request

Yes, sorry wan't looking for new names, just didn't want to say "intent" since that term is up for discussion.

elf-pavlik:
>>   specific offer, request

> Yes, sorry wan't looking for new names, just didn't want to say "intent" since that term is up for discussion.

Ok, I would like to once again emphasis, that agent can have existing Intents in a plan and make proposals for different *to provide* and *to receive* pairs. In many cases they can stay symmetric and this agent would not consider any of them as primary or secondary. Such proposals don't acts as just 'offer' or just 'request' but both at the same time. Agent offers *to provide* something and requests *to receive* something at the same time.
Intents themselves one could consider as offer *to provide* and request *to receive*, but I think that any reciprocity / exchange information needs to stay responsibility of Proposal (sometimes called listing) and that proposal itself needs to get considered as both offer & request at the same time.
UI would most likely show lists of Intents *to provide* as 'offers' and *to receive* as requests. Then based on information from number of Proposals, would show all the options for reciprocity available for any given intent.
Application implemented with pluralism in mind. Would also allow answering question 'what I can get in exchange of currency X' and one of the UX approaches could come from that agent having intent *to provide* that currency and application would show 'recommended matches' for that intent with available exchange opportunities. In that case it could simply match intents *to provide* and *to receive* 'currency X' just as it can match on any other *resourceClassifiedAs*.
Matching algorithm would of course need to take all the proposals into account, since without it it could only show 'yes those people want *to receive* currency X'. To know what they stay open to exchange it for, one would need to incorporate information from proposals.

elf-pavlik:
price list

In pluralism embracing application this looks pretty much like an example of 'what I can get in exchange of currency X'. Such application would allow views like that for unbound number of currencies, and if certain farmer created 3 different proposals for 3 different  currencies for all their intents *to provide* (offers), then their products will show in all those 3 answers for
* currency X
* currency Y
* currency Z

Asking about some edible *classification*, another view could snow such list but make 10 columns for 10 different currencies and number of non currency based proposals available as some kind of 'expand row' section.

Apps aiming at pluralism will need to invent new UX, while they still can provide views familiar to old non plural UX.

fosterlynn:
Choices I can see for the low level:
* All the flow related things are Intents, differentiated by property names (principally quantity)
* We name the different flow related things different @types (like Intent, Estimate, FlowTemplate)

In terms of the container level, some of that may depend on what we decide with higher-level processes.  It could be either a Proposal or a Process.  Proposal would imply seeking another agent(s) to create actual commitments. Process could be a budget, estimate, etc., more about planning at a higher level than a schedule - possibly a set longer time period, possibly more general classifications.  Probably not directly seeking commitments, but for analytical or general planning purposes.

I am leaning towards every flow thing being an Intent (or a better name), but I think I am fine with either way.  And defining Proposal, while waiting for other container classes for the process nesting PR.  So that would look something like:

```
Intent:
-- action
-- provider 
-- recipient
-- involves (EconomicResource)
-- resourceClassifiedAs
-- resourceConformsTo
-- intendedTime (a time:TemporalEntity)
-- frequency?
-- unitRateQuantity (this is the template or exchange rate quantity)
-- fullQuantity
-- estimatedQuantity
(all quantities include numericValue and unit)
-- isFinished
```

OR

```
FlowTemplate (or ExchangeRate or UnitRate or ?? - maybe UnitRateIntent):
-- action
-- provider 
-- recipient
-- involves (EconomicResource)
-- resourceClassifiedAs
-- resourceConformsTo
-- unitRateQuantity (numericValue and unit)
-- isFinished
```
```
SpecifiedIntent (or ???):
-- action
-- provider 
-- recipient
-- involves (EconomicResource)
-- resourceClassifiedAs
-- resourceConformsTo
-- intendedTime (a time:TemporalEntity)
-- fullQuantity (numericValue and unit)
-- frequency? 
-- isFinished
```
```
EstimatedIntent:
-- action
-- provider 
-- recipient
-- resourceClassifiedAs
-- resourceConformsTo
-- intendedTime (a time:TemporalEntity)
-- estimatedQuantity (numericValue and unit)
```

Would anyone ever want to have both the template/exchange rate info and the full info for any reason?  Nothing comes to mind, but.....

Opinions on options?

Better names?

fosterlynn:
And here's an attempt at Proposal, understanding that there will be other container classes for Intents.

Proposal:
-- validFrom
-- validThrough
-- isFinished
-- visibility (or audience?) (could be multiple agents)

ProposedIntent:
-- isReciprocal?? (not the best name.....?)

>Would you agree that this 'primary' / 'secondary' emphasis will need to stay available for Agreements as well? Since proposals can change, we may need a way that Agreement has its own 'copy' of that information

If we include the concept also in ExchangeAgreements, then some term with "match" doesn't make sense any more for both situations.  It could also be 2 relationships instead of a flag, especially if that is the only property in ProposedIntent.  Although the M:M is more difficult in UML, not so hard in RDF.

elf-pavlik:
> Intent:
> -- unitRateQuantity (this is the template or exchange rate quantity)

I still think any reciprocity / exchange responsibility has to stay with Proposal. Intent should directly only have responsibility relevant for its *action*.

Let's each of us propose 3 clear use cases tomorrow and then we'll go over them from Intents and Proposals to Agreements with Commitments and Events.

I see you already have at least 3 in https://github.com/valueflows/valueflows/issues/217#issuecomment-447376972 and we could use them and add the followup after responding to proposals.

fosterlynn:
>I still think any reciprocity / exchange responsibility has to stay with Proposal. Intent should directly only have responsibility relevant for its action.

I understand where you are coming from, I think, from a strict modeling perspective.  But I see the unit rate information as essentially an intent, just expressed slightly differently than an intent for a full quantity.  (Similar situation to having an intent or commitment that references a resource classification instead of an actual resource.)  It is not about the exchange rate, it is about this is what I have to offer, how ever many you want in this unit.  Most exchange apps I know about mix offers/requests that are expressed in a full amount with offers/requests that are expressed as unit rates, considering them as conceptually the same thing.  Which they also are from the user's perspective, the user is just looking for something or wanting to get rid of something.

elf-pavlik:
> But I see the unit rate information as essentially an intent, just expressed slightly differently than an intent for a full quantity.

I understand that this unit has nothing to do with proposals / exchanges. If you have such information on *produce* or *consume* intents, you would also rely on it when planning processes?


> Most exchange apps I know about mix offers/requests that are expressed in a full amount with offers/requests that are expressed as unit rates, considering them as conceptually the same thing.

I thought about it and I think we could consider those 'full amount' (atomic - commitments has to fully satisfy the intent) intent a special case of those proposals that scale (one can make many commitment which partially satisfy that intent). 


To emphasize once again, if we use informal terminology in our discussions:
* offer - an Intent *to provide*
* request - an Intent *to receive*

Non of them on has any exchange / reciprocity related responsibility. If someone wants to advertise a way to reach an Exchange Agreement. One has to use additional constructs to that take responsibility of expressing all the reciprocity related aspects. 

I think we will avoid confusion if we use more precise terms

* Intent *to provide*
* Intent *to receive*
* exchange Proposal
etc.

And as we consider additional constructs that we see needed to build Proposals, we can pick working names for them eg. Flow Template just to refer to them in unambiguous way.

elf-pavlik:
Responding to https://gitter.im/valueflows?at=5c1c28ba4303512ab478bde6

The same data covers both ways a human could formulate it (as discussed on chat)
> I need help for 5 hours moving, I will pay $20/hr 
> I need help for 5 hours moving, i will pay $100

I still need to write down YAML for an Agreement based on this Proposal with all the Commitments and Events.

It relies that pattern and intent it applies to match at least on *action* and *classification*. I think this part definitely needs more work.

```yaml
'@context':
  '@vocab': https://w3id.org/valueflows#
  xsd: http://www.w3.org/2001/XMLSchema#
  time: http://www.w3.org/2006/time#
  time:inXSDDateTimeStamp:
    '@type': xsd:dateTimeStamp
  alice: https://alice.example/


'@graph':

  - '@id': alice:c7dafbe5-42a7-444c-b5a3-e1b100129801
    '@type': Intent
    recipient: alice
    action: work
    workClassifiedAs: http://www.wikidata.org/entity/Q2918584 # relocation (moving)
    intendedQuantity:
      qudt:unit: unit:Hour
      qudt:value: 5
    atomicQuantity:
      qudt:unit: unit:Hour
      qudt:value: 5
    skos:note: Should take between 4 and 6 hours, I'll have cookies and tea or coffee
    intendedTime:
      '@type': time:ProperInterval
      time:intervalStarts:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2019-01-12T12:00:00-5:00
      time:intervalEnds:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2019-01-19T18:00:00-5:00

  - '@id': alice:df34e812-cd6f-49b7-9da3-478857f7dda3
    '@type': Intent
    provider: alice
    action: give
    resourceClassifiedAs: http://www.wikidata.org/entity/Q4917 # USD
    intendedQuantity:
      qudt:unit: unit:Unitless
      qudt:value: 400
    atomicQuantity:
      qudt:unit: unit:Unitless
      qudt:value: 0.01 # pennies OK
    intendedTime:
      '@type': time:ProperInterval
      time:intervalStarts:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2019-01-01T00:00:00-5:00
      time:intervalEnds:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2019-02-01T00:00:00-5:00

  - '@id': alice:3f3a4126-491d-41f3-bd5c-03dbbfa040a6
    '@type': Proposal
    forIntent:
      - alice:c7dafbe5-42a7-444c-b5a3-e1b100129801
      - alice:df34e812-cd6f-49b7-9da3-478857f7dda3
    followingPattern:
     - '@id': alice:3b0e355a-5620-44f8-bb4e-93d1c2fc456d
        action: work
        recipient: alice
        workClassifiedAs: http://www.wikidata.org/entity/Q2918584 # relocation (moving)
        quantityVector:
          qudt:unit: unit:Hour
          qudt:value: 1
      - '@id': alice:b6180863-49e7-4d2c-bf8e-22a38fa51ce8
        provider: alice
        resourceClassifiedAs: http://www.wikidata.org/entity/Q4917 # USD
        quantityVector:
          qudt:unit: unit:Hour
          qudt:value: 20
```

fosterlynn:
Closing, started a new one here https://github.com/valueflows/valueflows/pull/503.  (This has useful discussion in it though, so if it needs to be re-opened to be easily accessible as we continue to discuss, let's do that.)

Issue #336, opened 2016-08-23
start & stop events for usage & work (similar to load & unload) in transportation


elf-pavlik:
It looks like currently we can't track in real time which tools and people stay occupied (blocked) by which processes. With only _use_ and _work_ events we can't track the start and stop of work and usage. Many processes will also allow interruption so single usage and work will have multiple _start_ and _stop_ events referencing them.

fosterlynn:
A perspective on this issue: I don't think that start and stop work is actually the same as load/unload.  Stop work won't be an output of a process.  I think the way to think about it is just different methods of recording work.  In the end you get one or more observed work events.

elf-pavlik:
I think this ties to #230 and this diagram:

![flows](https://raw.githubusercontent.com/valueflows/valueflows/0d24fe91a85b02d7f6ec5b64033138cc40ea296c/assets/flows.png)

It only addresses future and past but we can have events which we started observing but they haven't finished yet. I think this might apply to any action, even *consume* on a grid resource (water, gas, electricity) which has a meter will have weeks long *consume* events.
We could possibly rely on `vf:finished` to tell difference between present events and past events. I think also `vf:Fulfillment` would get created only after the event finishes.

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

> https://lab.allmende.io/valueflows/valueflows/-/issues/336.

If you have not done so, you are very welcome to register at https://lab.allmende.io and join the ValueFlows organization there.

PR #383, opened 2018-12-31
example: bungalow rental and usage

elf-pavlik:
### example good practices
* it includes full JSON-LD `@context` and doesn't rely on any external context
* it includes all the data under `@graph` and doesn't try to frame it as a tree with a single root #123 

### VF impact

* example proposes new action `vf:lend` to represent lend/borrow flows (instead of discussed on chat *possess*)
* for time based flows, it uses quantity to represent quantity of the involved/affected resource and uses Time to represent duration
* it uses proposed in #293 `vf:includedIn` to relate short usage during working session with longer usage between checking-in and checking-out of rented bungalow

### open issues

* we can observe start of 1 year lend event when it starts but we can't observe the whole event yet, related to #336 
* we can record fulfilled quantity of a resource for 'ongoing event' but can record the fulfilled time yet

### DIFF
--- a/examples/bungalow-rental-and-usage.yaml
+++ b/examples/bungalow-rental-and-usage.yaml
@@ -1,12 +1,8 @@
 '@context':
-  '@vocab': https://w3id.org/valueflows#
-  xsd: http://www.w3.org/2001/XMLSchema#
-  time: http://www.w3.org/2006/time#
-  time:inXSDDateTimeStamp:
-    '@type': xsd:dateTimeStamp
-  elf: https://elf-pavlik.hackers4peace.net/
-  tashirat: https://tashirat.com/
-  notary: https://notary.example
+  - https://git.io/vf-examples-jsonld-context
+  - elf: https://elf-pavlik.hackers4peace.net/
+    tashirat: https://tashirat.com/
+    notary: https://notary.example
 
 '@graph':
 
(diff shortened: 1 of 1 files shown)


fosterlynn:
>example good practices

:+1: great, we will have to start gathering useful examples for the doc

>example proposes new action vf:lend to represent lend/borrow flows

>we can observe start of 1 year lend event when it starts but we can't observe the whole event yet

I think `lend` is a positive idea for an action.  But I would like to think about it as part of a chain, not part of a hierarchy.  So there would be a vf:lend and later a vf:return or something like that.  Each would be a point in time, some rights would have been passed along and be held until another event signifies they are over.  (Like in a library, people don't usually actually return the book on the due date.  Or in this example, if they don't pay the rent, the landlord might kick them out.)  Thinking of it as a chain I think resolves the open issues you posed.  You now have some rights, and you can log usage etc.

examples/bungalow-rental-and-usage.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+  - '@id': tashirat:953b40f2-4e47-4486-afb2-402512f5269c
+    '@type': Commitment
+    under: notary:40bb78fe-28df-41a6-b698-525fa9d411d6
+    provider: tashirat
+    recipient: elf
+    action: lend
+    involves: tashirat:2402b57c-1c45-423a-acb2-f50961c97d65
+    commitedQuantity:
+      qudt:unit: unit:Unitless
```
fosterlynn: I wonder if we should investigate using something besides qudt?  I heard of one from Europe I think, that might be simpler, but I'd have to track it down.  Also, I think I found something in qudt for 'each' - maybe Count or Number or something?

examples/bungalow-rental-and-usage.yaml
```diff
@@ -0,0 +1,164 @@
...
+      qudt:value: 1
```
fosterlynn: qudt is numericValue I believe

examples/bungalow-rental-and-usage.yaml
```diff
@@ -0,0 +1,164 @@
...
+    commitedTime:
```
fosterlynn: committedTime (two t's)

examples/bungalow-rental-and-usage.yaml
```diff
@@ -0,0 +1,164 @@
...
+    '@type': EconomicResource
+    rdfs:label: Bungalow 6
+
+  - '@id': notary:40bb78fe-28df-41a6-b698-525fa9d411d6
+    '@type': EchangeAgreement
+    skos:note: One year rental agreement of Bungalow 6 for 3000MXN a month
+
+  - '@id': tashirat:953b40f2-4e47-4486-afb2-402512f5269c
+    '@type': Commitment
+    under: notary:40bb78fe-28df-41a6-b698-525fa9d411d6
```
fosterlynn: I'd like to consider something besides `under` for exchange agreements, where the commitments are a big part of the agreement, as opposed to  the agreement explaining why the commitment happened.

examples/bungalow-rental-and-usage.yaml
```diff
@@ -0,0 +1,164 @@
...
+    '@type': EconomicEvent
+    skos:note: usage of bungalow as workspace durring one working session
+    action: use
+    affects: tashirat:2402b57c-1c45-423a-acb2-f50961c97d65
+    inputOf: elf:ee652454-ac08-4f5d-ae84-6b1753887086
+    includedIn: elf:f12fc3b7-1033-4da0-85fd-a378b4cfb36f
+    observedQuantity:
+      qudt:unit: unit:Unitless
+      qudt:value: 1
+    observedTime:
```
fosterlynn: Are you proposing changing affectedQuantity and affectedTime to observed...?  I think I might like observed better also.  But it means some changes to the handful of VF projects we have going now.  I think 2 of those will be finishing a major re-write in January, I'd like to do as much of this kind of naming cleanup as possible by say mid-January, because it will only get more painful.

fosterlynn:
>for time based flows, it uses quantity to represent quantity of the involved/affected resource and uses Time to represent duration

I think this adds unnecessary complexity to queries and processing - events with time based resources will be using a different quantity than others, also depending on the action.  I think the vf:underlyingResource keeps it simpler and cleaner.

examples/bungalow-rental-and-usage.yaml
fosterlynn: Oops, it is already observedTime

elf-pavlik:
I have fixed those two typos with *commited* and *qudt:value*

> I think this adds unnecessary complexity to queries and processing - events with time based resources will be using a different quantity than others, also depending on the action. I think the vf:underlyingResource keeps it simpler and cleaner.

Could you please write a snippet with EconomicEvent *observation* of 5 hours use of 3 electric drill managed as stock with many drills in it - we have just one resource representing all those electric drills with quantity 20 or whatever and we use 3 out of them for 5 hours (no serial numbers tracked on those drills).

examples/bungalow-rental-and-usage.yaml
elf-pavlik: While agreement may have some textual notes, from machine processing requirements flows that reference it with `vf:under` define it. At least as far as VF terms allow to reason.

elf-pavlik: Let's deal with that as part of *agreements* milestone

elf-pavlik:
I have updated YAML in this PR, I removed *lend* and just went with *use* everywhere. I don't know if we may need to tackle *composition* milestone first. YAML here includes 3 'levels' of nested events. In practice I think that even on days where we don't stay at the bungalow but we still have bunch of stuff here we still *use* it. In addition when we *use* it by keeping our stuff in it, but we don't stay here ourselves, someone else could stay over and *use* it as well. I wonder if we could just rely on classifications on flows to deal with those differences. On the other hand I don't think we should rely on classifications to resolve availability.

elf-pavlik:
> But, everything until the last `use` is really about transfers of rights, not really about use, no? You have the ability and rights to use the apartment for a year, but it is just that you can use it if you like.

I still think that making a commitment has similar effect as 'transferring very specific rights'. I think #387 might give use easier start and later we can see if we can model this one in similar way.

fosterlynn:
I think we should move this one to the Transfer milestone.  Although it involves composition too, but I don't thin it can be resolved without transfers.

fosterlynn:

Issue #307, opened 2018-12-16
Exchange without commitments


fosterlynn:
This issue was triggered by a demo implementation we are doing that doesn't yet have Commitment.  I didn't see a way that we could record reciprocal Transfers without a Commitment or Agreement of some sort in our current model.  Seems like we need to be able to do that, irrespective of a limited demo.

The most obvious use case is retail.  But also the appreciation use case fits in here too.  And any scenario where an exchange is made on the spot without prior commitment.

Here is a current view and a proposed view.  Naming could probably use some tweaking.  Note that Reciprocity in the proposed view would cover appreciation, just probably not need the quantity.  Also note that in the retail scenario, there could be a Proposal with an exchange rate attached that would drive the quantities.  The M:M supports retail scenario of purchasing a bunch of stuff and paying with one currency amount.  Not sure how we would support trading one bunch of stuff for another bunch of stuff.... might need another way to bundle it.

![exchange-no-commit](https://user-images.githubusercontent.com/3776081/50055103-b7b8f580-010f-11e9-8d9a-b0330ccd4710.png)

fosterlynn:
> Not sure how we would support trading one bunch of stuff for another bunch of stuff.... might need another way to bundle it.

Actually, our current model for the use case of an EconomicEvent creating a Commitment doesn't really support trading many things for many things.  It supports many EconomicEvents creating a Commitment, and one EconomicEvent creating many Commitments, but not many EconomicEvents creating many Commitments, as far as I can see.

fosterlynn:
>>Not sure how we would support trading one bunch of stuff for another bunch of stuff.... might need another way to bundle it.

>Actually, our current model for the use case of an EconomicEvent creating a Commitment doesn't really support trading many things for many things.

A possible solution would be to create a new EconomicResource which vf:contains the bundle of resources.  Then there would be one transfer for each bundle.

elf-pavlik:
Could you please write a snippet to clarify your use case? I think that we might simply reuse however we model exchange rate and have a way to save it (make a copy) and reference from an agreement. Let's discuss it around specific snippet.

elf-pavlik:
I think the only difference in representing 'exchange rate' / 'reciprocity ratio' between Proposal and Agreement, comes down that in Commitments and Events we have quantities specific for that Agreement, while quantities in Intents don't stay specific for any specific Proposal.

[...]

PR #384, opened 2018-12-31
Support for reciprocal events without reciprocal commitments.

fosterlynn:
This will support retail situations or other situations without prior commitments.  Used Reciprocity name for this situation, replacing Appreciation, which it also will support.  Renamed old Reciprocity to Claim, which ties in with standard REA naming better anyhow.

Retail could also be supported by creating a quick ExchangeAgreement and tying the events to that.  Might require renaming, but that would be similar to other situations under discussion, so better to do then.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -95,13 +95,17 @@ vf:Fulfillment  a            owl:Class ;
         rdfs:label       "vf:Fulfillment" ;
         rdfs:comment   "The quantity that the vf:EconomicEvent fulfilled towards the vf:Commitment." .
 
+vf:Claim  a              owl:Class ;
+        rdfs:label       "vf:Claim" ;
+        rdfs:comment     "The quantity that the vf:EconomicEvent claims towards creation of the reciprocal vf:Commitment." .
+
 vf:Reciprocity  a        owl:Class ;
         rdfs:label       "vf:Reciprocity" ;
-        rdfs:comment     "The quantity that the vf:EconomicEvent implies towards creation of the reciprocal vf:Commitment." .
+        rdfs:comment     "The quantity that the vf:EconomicEvent implies related to the reciprocal vf:EconomicEvent.  Or, an appreciation for a prior event, without quantity, as in the gift economy." .
 
-vf:Appreciation  a            owl:Class ;
-        rdfs:label       "vf:Appreciation" ;
-        rdfs:comment     "A way to tie an economic event that is given in loose fulfilment for another economic event, without commitments or expectations. Supports the gift economy." .
+#vf:Appreciation  a            owl:Class ;
+#        rdfs:label       "vf:Appreciation" ;
+#        rdfs:comment     "A way to tie an economic event that is given in loose fulfilment for another economic event, without commitments or expectations. Supports the gift economy." .
 
 vf:Location  a           owl:Class ;
         rdfs:label       "vf:Location" .
@@ -240,22 +244,34 @@ vf:inScopeOf  a      owl:ObjectProperty ;
         rdfs:comment "Grouping around something to create a boundary or context, used for documenting, accounting, planning." .
 
 vf:under  a          owl:ObjectProperty ;
-        rdfs:domain  [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Reciprocity vf:Fulfillment ) ] ; 
+        rdfs:domain  [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Claim vf:Fulfillment ) ] ; 
         rdfs:label   "under" ;
         rdfs:range   vf:Agreement ;
         rdfs:comment        "Reference an agreement between agents which specifies the rules or policies which govern this event." .
 
-vf:createdBy  a        owl:ObjectProperty ;
-        rdfs:label   "created by" ;
-        rdfs:domain  vf:Reciprocity ;
-        rdfs:range   vf:EconomicEvent ;
+vf:claimedBy  a             owl:ObjectProperty ;
+        rdfs:label          "claimed by" ;
+        rdfs:domain         vf:Claim ;
+        rdfs:range          vf:EconomicEvent ;
         rdfs:comment        "References the economic event that fully or partially created the commitment, often based on a prior agreement." .
 
-vf:creates  a        owl:ObjectProperty ;
-        rdfs:label   "creates" ;
-        rdfs:domain  vf:Reciprocity ;
-        rdfs:range   vf:Commitment ;
+vf:claims  a                owl:ObjectProperty ;
+        rdfs:label          "claims" ;
+        rdfs:domain         vf:Claim ;
+        rdfs:range          vf:Commitment ;
         rdfs:comment        "References the commitment that was fully or partially created because of the economic event, often based on a prior agreement." .
+
+vf:reciprocalFrom  a         owl:ObjectProperty ;
+        rdfs:label          "reciprocal from" ;
+        rdfs:domain         vf:Reciprocity ;
+        rdfs:range          vf:EconomicEvent ;
+        rdfs:comment        "References the economic event that the economic event is reciprocal from." .
+
+vf:reciprocalTo  a            owl:ObjectProperty ;
+        rdfs:label          "reciprocal to" ;
+        rdfs:domain         vf:Reciprocity ;
+        rdfs:range          vf:EconomicEvent ;
+        rdfs:comment        "References the economic event that the economic event is reciprocal to." .
 
 vf:contains  a       owl:ObjectProperty ;
         rdfs:label   "contains" ;


fosterlynn:
Reference #307

elf-pavlik:
@fosterlynn could you please add at least one `.yaml` file that shows exactly how you imagine using those constructs? I think we should take it as general practice that we always include snippets with examples whenever we add something to the vocab.

fosterlynn:
>@fosterlynn could you please add at least one .yaml file that shows exactly how you imagine using those constructs?

Added one.  I suppose there will be different use cases, but see what you think of this one.

fosterlynn:
Added one involving work.

elf-pavlik:
Thank you @fosterlynn looking at example I see that you look for something similar to `ReciprocityRatio` we have discussed in PR *exchange rates* https://github.com/valueflows/valueflows/pull/301#issuecomment-448004837

As you can see in my last comment in that PR I've found alternative based on 'flow patterns' as more promising. 

With the approach you propose here how do you express exchange: 100h of work for 4 weeks of accommodation + meals (2 a day) + 10kg of chocolate + 10kg of coffee with reciprocity ratios
* 20h of work for 1 week of accommodation + meals (2 a day)
* 1h of work for 1kg of chocolate
* 1h of work for 1kg of coffee

What you propose could very likely handle last two but not the first one which relies on 3 flows not a pair.

Could you also ad examples that use proposed `vf:Claim` ?

BTW files should have `.yaml` extension

examples/reciprocal-events-timebank
```diff
@@ -0,0 +1,72 @@
...
+    observedTime:
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-12-30T00:00:00-5:00
+
+  - '@id': timebank:7dd207e3-8f11-4fe3-95f8-d02f6c180be3
+    '@type': EconomicEvent
+    provider: bob
+    recipient: alice
```
elf-pavlik: If *give* event already have provider and recipient why do you also have it repeated on *receive* event?
I would find it clearer if either *give* or *receive* have provider and recipient but not both. Otherwise it appears like Bob provided 5 credits to Alice twice.

fosterlynn: Because `give` event is only in the scope of the provider and `receive` event is only in the scope of the receiver, so they wouldn't see the other one - not always, but often.

fosterlynn: I think there are several ways this can be modelled, and none of them are perfectly clean, they all have trade-offs.  At least, I haven't found the perfect one yet.

fosterlynn:
>Could you also ad examples that use proposed vf:Claim ?

That is just a re-name of the old vf:Reciprocity, because I wanted to use that name for the new one.  And it was a chance to use the word Claim without messing with your desire to unify claims and commitments.  (Claim is the standard term in this case I believe - where an event means you have a claim on someone else to do something in return.)

fosterlynn:
I added .yaml extension, thanks for catching that.

fosterlynn:
>Thank you @fosterlynn looking at example I see that you look for something similar to ReciprocityRatio

It is similar but not the same.  ReciprocityRatio is like "unit price" as I understand it.  This is more like "line item price", already calculated.

>20h of work for 1 week of accommodation + meals (2 a day)

In that example I would bundle the accommodation + meals into one resource specification, since that is how it is "priced".  (There is no information about accommodation and meals separated.)

>With the approach you propose here how do you express exchange: 100h of work for 4 weeks of accommodation + meals (2 a day) + 10kg of chocolate + 10kg of coffee with reciprocity ratios

This could be expressed as:
* 4 weeks accomodation + meals for 80 hours of work
* 50kg of chocolate for 5 hours of work
* 150kg of coffee for 15 hours of work

elf-pavlik:
> It is similar but not the same. ReciprocityRatio is like "unit price" as I understand it. This is more like "line item price", already calculated.

Where "line item price" = "unit price" * quantity ?

> In that example I would bundle the accommodation + meals into one resource specification, since that is how it is "priced".

I have impression that we may get more robust approach if we compose around flows and not as much around resources. I believe YAML examples can help with clarifying it.

When you plan production of meals and do scheduling on accommodation spaces. How do you incorporate commitments based on meals + accommodation? 

> (There is no information about accommodation and meals separated.)

Lets consider scenario where agent provides meals and accommodation separately as well, but has a special proposal that includes both and resulting agreements based on that proposal will have commitments that add up differently than if agents would have made commitments separately.

elf-pavlik:
I'd like to explore similar direction to #293 (Process nesting) and have Agreements nesting. An agreement could have 'lower' agreement for: *20h of work for 1 week of accommodation + meals (2 a day)*. This nested/lower sub-agreement would represent equivalent of "line item price", which would allow composing commitments with the same resource classification using `vf:includedIn`. 

> 100h of work for 4 weeks of accommodation + meals (2 a day) + 10kg of chocolate + 10kg of coffee
>* 20h of work for 1 week of accommodation + meals (2 a day)
>* 1h of work for 1kg of chocolate
>* 1h of work for 1kg of coffee

We would have an agreement with 3 sub agreements

1.
    * commitment for provide 80h of work included in commitment for 100h of work
    * commitment for receive 4 weeks of accommodation
    * commitment for receive 2 meals a day in those for weeks (maybe even a commitment per meal)
2.
    * commitment for provide 10h of work included in commitment for 100h of work
    * commitment for receive 10kg of chocolate
3. 
    * commitment for provide 10h of work included in commitment for 100h of work
    * commitment for receive 10kg of coffee

I find this approach more aligned with a spirit that exchanges can involve multiple flows and we don't always assume to have just pair of flows (which only applies to special case where all exchanges use certain currency).

Another real world use case for exchange based on 3 flows
* 500g bar of chocolate 100%
* 80 MXN
* 20 Olines (local LETS currency)

Here proposal and resulting agreement exchanges a bar of chocolate for some amount of mexican peso and some amount of local LETs credit.

 > In that example I would bundle the accommodation + meals into one resource specification, since that is how it is "priced".

I don't think such 'bundling' attempt would make sense for that ISO4217 + LETS use case

fosterlynn:
To clarify the original use case, I'd like to make sure we can support an observation layer without anything else.  Same as wanting to support a planning layer without a knowledge layer if needed.

In the case of 1:M or M:M resource classifications, I think that people could use either vf:contains to bundle resources, or divide up by "line item price" (for lack of a better term), or even just create the Reciprocity instances without quantities.  Seems like it depends on the case.

Unless there is a better way to model this requirement?

>Another real world use case for exchange based on 3 flows
>    500g bar of chocolate 100%
 >   80 MXN
  >  20 Olines (local LETS currency)

For this, if quantities are desired, the user would have to know how much of the 500g bar is paie for by MXN and how much by Olines.  Or just leave the quantities in Reciprocity, and get all related Reciprocities, which should give the info of the exchange.

elf-pavlik:
> Unless there is a better way to model this requirement?

I don't think we should expect that we will always have mutually reciprocal pair of events.

>> Another real world use case for exchange based on 3 flows
500g bar of chocolate 100%
80 MXN
20 Olines (local LETS currency)

> For this, if quantities are desired, the user would have to know how much of the 500g bar is paie for by MXN and how much by Olines.

I see it as trying to fit something in a brittle model which expect only reciprocal pairs. 

> To clarify the original use case, I'd like to make sure we can support an observation layer without anything else. Same as wanting to support a planning layer without a knowledge layer if needed.

In case of exchanging 500g of chocolate for 80 MXN + 20 Olines we don't have any problem with just relying on the exchange agreement with 3 events. At the same time I see that if we add to the same agreement 1kg of coffee for 100 MXN + 30 Olines and we only have actual events for 180MXN and 50 Olines we don't want to add additional events just to create two separate sub agreements.

Having already construct of Commitment and Fulfillment, we can just create separate commitments for those sub agreements and then single event will fulfill them both, or we can have 'upper' commitment referenced from lower commitments using `vf:includedIn`. It seems to me that we may face choice
* not to draw 'hard lines' between layers and just use existing Commitment and Fulfillment in nested agreements to model non-dual (only pairs) reciprocity.
* draw 'hard lines' between layers and come up with equivalents to Commitment and Fulfillment on 'observation' layer to model non-dual (only pairs) reciprocity.

I think we already have couple of use cases showing different scenarios in this issue, I'll write YAML which tries to simply reuse Commitment and Fulfillment in sub agreements.

fosterlynn:
When I'm thinking about wanting to support an observation layer that doesn't need a planning layer, it is for things that don't actually have any planning or commitments.  (Not trying to duplicate planning on the observation layer.)  So I'll be interested in what you come up with.

elf-pavlik:
> When I'm thinking about wanting to support an observation layer that doesn't need a planning layer, it is for things that don't actually have any planning or commitments.

I think sometimes we don't state it clear if:
* no agreement prior to events happened in actual situation
* agents haven't recorded any details of the agreement before the events

In actual situation I would see agents to agree for both
* exchanging 500g of chocolate for 80 MXN + 20 Olines
* agreement 1kg of coffee for 100 MXN + 30 Olines

And 10s later decide to handle it with single *receive* event of 180MXN and *receive* event of 50 Olines.

If they don't want to just record:
* exchanging 500g of chocolate and 1kg of coffee for 180MXN + 50 Olines

But want to record it broken down into distinct:
* exchanging 500g of chocolate for 80 MXN + 20 Olines
* agreement 1kg of coffee for 100 MXN + 30 Olines

It looks to me that they want a record details of an agreement based on which those events occurred. 

Those
> "line item price" (for lack of a better term)

represent for me agreed exchanges with specific commitments, where some of them (with same resource classification) later get fulfilled by a single event.

elf-pavlik:
I think transportation gives another common non-dual scenario. Here commitments make it possibly more clear.

```yaml
'@context':
  '@vocab': https://w3id.org/valueflows#
  xsd: http://www.w3.org/2001/XMLSchema#
  qudt: http://qudt.org/schema/qudt/
  unit: http://qudt.org/vocab/unit/
  time: http://www.w3.org/2006/time#
  time:inXSDDateTimeStamp:
    '@type': xsd:dateTimeStamp
  alice: https://alce.example/
  bob: https://bob.example/
  claudia: https://claudia.example/
  notary: https://notary.example/
  faircoop: https://faircoop.example/

'@graph':

  - '@id': notary:9bd19194-a36d-4a1f-896b-8082887962cb
    '@type': ExchangeAgreement

  - '@id': claudia:33e8933b-ff73-4a01-964a-ca7a98893083
    '@type': Process
    skos:note: transporting Bob's package

  - '@id': bob:60f4204e-b8d2-4026-8577-102c3f82c0af
    '@type': EconomicResource
    skos:note: Bob's package
    currentQuantity:
      qudt:unit: unit:Unitless
      qudt:numericValue: 1

  - '@id': claudia:ad56a7ed-be3c-4937-a3fb-0f156bcd2c47
    '@type': Commitment
    action: load
    inputOf: claudia:33e8933b-ff73-4a01-964a-ca7a98893083
    involves: bob:60f4204e-b8d2-4026-8577-102c3f82c0af
    committedQuantity:
      qudt:unit: unit:Unitless
      qudt:numericValue: 1
    under: notary:9bd19194-a36d-4a1f-896b-8082887962cb
    provider: claudia
    receiver: bob

  - '@id': claudia:6078f30a-b15b-4b1c-8e77-797abae11378
    '@type': Commitment
    action: unload
    outputOf: claudia:33e8933b-ff73-4a01-964a-ca7a98893083
    involves: bob:60f4204e-b8d2-4026-8577-102c3f82c0af
    committedQuantity:
      qudt:unit: unit:Unitless
      qudt:numericValue: 1
    under: notary:9bd19194-a36d-4a1f-896b-8082887962cb
    provider: claudia
    receiver: bob

  - '@id': bob:e56fd654-7b94-4d96-8e60-de39e08329a7
    '@type': Commitment
    action: give
    resourceClassifiedAs: http://www.wikidata.org/entity/Q89 # apples
    committedQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 10
    under: notary:9bd19194-a36d-4a1f-896b-8082887962cb
    provider: bob
    receiver: claudia
```

Bob doesn't give her 5 apples for loading in one place and 5 apples for unloading in another place. He gives them for the whole transportation service which requires both loading and unloading.

fosterlynn:
Maybe we need to bring ExchangeAgreement back down to the boundary between planning and observation, and change the name to Exchange, and let it be the thing that ties reciprocal transfers together, not worrying about if it is an agreement.  It is an agreement if it has reciprocal Commitments.

Does that make any sense?  

Then just drop what I was trying to do, and just record events against the Exchange.

elf-pavlik:
I don't recall ever having Agreement just 'on planning layer', if we look at this old diagram it stays 'in between' layers just like Process does.

vf layers

I think you included a `vf:ExchangeAgreement` as subclass of `vf:Agreement` just to have word *exchange* somewhere. In the end the *provider* and *receiver* relations on flows *under* an Agreement define if we talk about exchange, pledge or whatever other constellation of flows.

fosterlynn:
>I don't recall ever having Agreement just 'on planning layer', if we look at this old diagram it stays 'in between' layers just like Process does.

How is an agreement in any way observable?

elf-pavlik:
In the same way as Process, both get referenced from Commitments and Events. Commitments we can consider as a record of a plan, while Events what we observe (ongoing) or observed (done).

PR #386, opened 2019-01-02
initial transportation use case snippets

elf-pavlik:
Initial examples for https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#transportation

* some variants similar to other variants have differences explained in comment blocks.
* load & unload for packaged and serial variants affect the same (transported) resource just in two different places
* snippets introduce *assemble* and *disassemble* actions as reversible variants of *consume* and *produce*

### TODO

* [x] variant of transporting non serial resource where transportation includes *load* from different resources with the same classification and *unload* to different resources with the same classification, while still allows to *tie together* matching *load* & *unload*. Similar to what current `vf:Transfer` does but without having anything to do with *rights and responsibilities*.

### DIFF
--- a/examples/transportation.yaml
+++ b/examples/transportation.yaml
@@ -58,6 +58,13 @@
       qudt:unit: unit:Kilogram
       qudt:numericValue: 40
 
+  - '@id': bob:a59e121a-89af-43c1-a247-e765137ba84c
+    '@type': EconomicResource
+    clasifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 110
+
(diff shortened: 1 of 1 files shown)


elf-pavlik:
Last commits adds a concurrent transportation example addressing *TODO* in issue description. It tries to reuse process nesting #293 to connect *load* and *unload* pairs in non ambiguous way.

elf-pavlik:
Latest commit adds snippet which involves passenger transportation. It works a little bit like package, and in this case *split* and *merge* sound little weird :package:

elf-pavlik:
#393 includes scenario which shows some limitations with *load* and *unload* - lack of 'on the vehicle' resource which we can get affected by various events.

fosterlynn:

PR #387, opened 2019-01-03
initial examples for shared use of independently controlled studios

elf-pavlik:
Based on https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#independent-shared-therapy-studios

Making commitment to oneself serves purpose of reserving the time and not making conflicting commitment to someone else.

I plan to add another variants where agents can exchange commitments. So when Alice makes commitment to Daniel for use of Alice's studio and then she gets commitment from Bob for use of his studio. Possibly she could exchange those commitments with Daniel and use the studio she controls while Daniel gets to use studio that Bob controls.

### DIFF
--- a/examples/shared-studios-usage.yaml
+++ b/examples/shared-studios-usage.yaml
@@ -2,16 +2,11 @@
 # https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#independent-shared-therapy-studios
 
 '@context':
-  '@vocab': https://w3id.org/valueflows#
-  xsd: http://www.w3.org/2001/XMLSchema#
-  time: http://www.w3.org/2006/time#
-  time:inXSDDateTimeStamp:
-    '@type': xsd:dateTimeStamp
-  alice: https://alice.example/
-  bob: https://bob.example/
-  claudia: https://claudia.example/
-  daniel: https://daniel.example/
-  erica: https://erica.example/
+  - https://git.io/vf-examples-jsonld-context
+  - alice: https://alice.example/
+    bob: https://bob.example/
+    claudia: https://claudia.example/
+    daniel: https://daniel.example/
 
 '@graph':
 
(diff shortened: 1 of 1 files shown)


fosterlynn:
I'm still pretty doubtful about keeping the main quantity (the one that is meaningful in `use`) in different properties for different use cases.  I'll take one of your examples like that and create some yaml for alternative option of vf:underlyingResource, so we can compare.

fosterlynn:
After some thought, I think that we want to use vf:underlyingResource for situations that involve usage or possession or similar, requiring a different unit of measure for the "same" resource.  

One reason is that there could be many different usage units, and usage arrangements.  Like an owner of an apartment may offer rights to use it by the month or by the week.  Or a rental car by miles and/or by days.

This is besides my original objection, which is that I think it is important for simplicity of processing that the quantity means the same thing across actions and kinds of resources.

I'll do a comparison yaml treatment, to find those devils in the details.

elf-pavlik:
> Like an owner of an apartment may offer rights to use it by the month or by the week.

This seems similar to how products get packaged. In https://github.com/valueflows/valueflows/pull/302#issuecomment-449200205 I experiment with `vf:atomicQuantity` to express minimum 'size' independent from the base physical unit.

> Or a rental car by miles and/or by days.

If we always use dedicated property for temporal intervals. We still have property for non-temporal quantities available. I think some services (eg. taxi) may combine distance quantity & time interval.

> This is besides my original objection, which is that I think it is important for simplicity of processing that the quantity means the same thing across actions and kinds of resources

Non-temporal quantity affects quantity of the resource (eg. 5 electric drills out of 20 managed as that resource). Temporal interval doesn't affect `vf:currentQuantity` of the resource, it affects its schedule which we haven't really work on.
We exist in spacetime and I don't see any problem with needing a dedicated way to handle the temporal aspect - schedules. At the same time I think we don't need to support 'user defined' dimensions :wink:

fosterlynn:
>At the same time I think we don't need to support 'user defined' dimensions wink

Didn't understand.  If you mean units, then I agree, was I doing that?

>We exist in spacetime and I don't see any problem with needing a dedicated way to handle the temporal aspect - schedules.

You don't see any problem with all reasoning over events needing to figure out which field to look at for quantity?

elf-pavlik:
> You don't see any problem with all reasoning over events needing to figure out which field to look at for quantity?

Let's implement something simple and see how it works. Affected quantity of a resource and durations of events sound like a distinct concerns to me that affect different things:
* quantities - current amounts of async stock resrouces
* duration - schedules

fosterlynn:
I think this one will work OK with what I proposed in https://github.com/valueflows/valueflows/pull/446.  The change would be:

```
    commitedQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 4
```

If we use the committedTime to record what needs to be counted for time-based resources, then we lose the ability to record a more general time to be used as calendar time rather than all the time something was actually used.  The way I've started thinking about it is the quantity is for what needs to be counted (with one additional quantity field if needed for partial stock resource use), and the time field is for scheduling, visualizing on a timeline, etc.

A possible compromise would be to use the main quantity field as you are doing, and add an additional one for measuring the usage - but I think that is more complex and would require more processing rules that need to be explained.

But in practice, as far as I've observed, a use case like the drills will be extremely rare.  If people want to record use of equipment or tools, it is important enough to keep track of for each uniquely identified resource, for things like knowing when to do preventative maintenance.  Below some threshold of actual need, people won't tend to record the usage.  I personally think it will be so rare that we shouldn't even document the option, although I'm willing to include it.

elf-pavlik:
> If we use the committedTime to record what needs to be counted for time-based resources, then we lose the ability to record a more general time to be used as calendar time rather than all the time something was actually used.

Since this scenario doesn't require any 'more general time', let's discuss that aspect with use cases that require it.

> The way I've started thinking about it is the quantity is for what needs to be counted (with one additional quantity field if needed for partial stock resource use), and the time field is for scheduling, visualizing on a timeline, etc.

To have clarity on capacity/availability of resources and services 'what needs to be counted' will differ depending on resource or service. Not always just one unit + quantity will suffice. Since we exist in time-space, temporal dimension will always play some role and we need to handle it well. I don't see it as 'some user defined unit of measuring', I say that only because I recall some conversation about 'user defined units' and I don't see temporal aspect as one of those.

fosterlynn:
>> The way I've started thinking about it is the quantity is for what needs to be counted (with one additional quantity field if needed for partial stock resource use), and the time field is for scheduling, visualizing on a timeline, etc.

>To have clarity on capacity/availability of resources and services 'what needs to be counted' will differ depending on resource or service. Not always just one unit + quantity will suffice. Since we exist in time-space, temporal dimension will always play some role and we need to handle it well. I don't see it as 'some user defined unit of measuring', I say that only because I recall some conversation about 'user defined units' and I don't see temporal aspect as one of those.

Yes capacity is not part of counting / accounting though.  Capacity is part of scheduling, planning.  And the times also place events into their place in time for various reasons, again not to be counted really.

I'm not saying to not handle the time dimension at all.  Just that it is a different dimension and we should carefully separate the concerns.  Not muddle counting into times.

I completely agree that temporal units can't be user defined.  They are a different thing - another reason to separate the concerns.  You can't imply a unit from a ProperInterval anyhow.

fosterlynn:

PR #388, opened 2019-01-04
simple transfer yaml

fosterlynn:
for the examples

### DIFF
--- a/examples/transfer-simple.yaml
+++ b/examples/transfer-simple.yaml
@@ -1,0 +1,68 @@
+# Simple transfer
+
+'@context':
+  '@vocab': https://w3id.org/valueflows#
+  xsd: http://www.w3.org/2001/XMLSchema#
+  time: http://www.w3.org/2006/time#
+  time:inXSDDateTimeStamp:
+    '@type': xsd:dateTimeStamp
+  alice: https://alice.example/
+  bob: https://bob.example/
+
+'@graph':
+
+  # initial resources
+
+  - '@id': alice:c7897c39-7f05-4a5d-a487-80e130a24345
+    '@type': EconomicResource
+    classifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 670
+    trackingIdentifier: lot-4399jj93
+
+  # Alice gives Bob 200 kg apples.
+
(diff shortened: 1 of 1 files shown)


fosterlynn:
@elf-pavlik a couple questions:
1. Can you tell me again how you generate those curie's?
2. Shouldn't we have qudt in the @context in our examples?

examples/transfer-simple.yaml
```diff
@@ -0,0 +1,68 @@
...
+    trackingIdentifier: lot-4399jj93
+
+
+  - '@id': bob:0f563083-7f05-4a5d-a487-80e130a24345
+    '@type': EconomicResource
+    classifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 200
+    trackingIdentifier: lot-4399jj93
```
elf-pavlik: It looks that to get this identifier here, system has to have access to both *give* and *receive* and the resource that *give* affects. You have mentioned that sometimes Bob will not have access to record about Alice's resource, sometimes not even have the *give* event, only *receive*. How do you see `trackingIdentifier` (LOT) propagating in those cases?

elf-pavlik: This also relates to recent discussions in #391 and #392

examples/transfer-simple.yaml
```diff
@@ -0,0 +1,68 @@
...
+    classifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 670
+    trackingIdentifier: lot-4399jj93
+
+  # Alice gives Bob 200 kg apples.
+
+  - '@id': alice:e7482f32-25b5-4f79-934b-5d0db3a5b77d
+    '@type': Transfer
```
elf-pavlik: can we just use `vf:Process` here since #389 will clarify that *give* and *receive* do NOT 'change existence' and based on meaning of those actions system should not consider this process as *transformation*?

examples/transfer-simple.yaml
elf-pavlik: We could remove `trackingIdentifier` on resources in this PR since we don't have a way for events to propagate it. Then in #392 we could sort out how to do it and then update this example with those LOT numbers as well as update #391 algorithm to use those tracking identifiers.

fosterlynn:
outdated

Issue #355, opened 2018-12-18
hard vs soft allocations


bhaugen:
These are used both in resource sales as well as issues to manufacturing processes.

A soft allocation is a type of rea:Commitment which allocates a quantity of an rea:ResourceType but does not allocate any particular rea:Resources.

For example, in a stock of apples, Pavlik may have a commitment with Bob for 5k of Apples. That is, any apples, not any particular apples.

Bob may have 20kg of Apples. A soft allocation would subtract 5kg from the available quantity, but not the onhand quantity. So Bob would only have 15kg left available for other requests.

A hard allocation would put 5kg of Apples in a container for Pavlik to pick up. 

The purpose of a hard allocation is to avoid somebody else coming along and taking more than 15kg of Apples, because the 5kg allocated to Pavlik was only known to the inventory system, not obvious in the bin of Apples.

PR #389, opened 2019-01-04
Define actions.

fosterlynn:
I started this branch to define one or two actions for "adjust", which we have used for things like beginning balances or inventory counts that differ from the system.  But, trying to be as consistent as possible, I decided to use `increment` and `decrement` for these adjustments.  Mostly because I couldn't think of one word actions for the adjust-up and adjust-down, and didn't want to use negative numbers instead of "decrement" behavior, since adjustments can go either way.

Then I decided we need to define more about actions anyhow, so I started that exercise.  Note ? on use and work, because we haven't defined time-based resources in detail yet.  As of now they just match what we already have defined.


docs/introduction/flows.md
```diff
@@ -12,15 +12,44 @@ Intent describes some potential future event which might require prior Commitmen
...
+accept | No effect | Process | Input | improve |
+give | Decrement | Transfer | Input | receive |
+
+We have defined a core set of actions, but expect that this will be extended with others. If extended, they should be defined as part of this or another formal vocabulary so that all can use them and assume the same meaning. 
+
+Note: Increment and decrement are usually not meant to be used directly, rather the actions inherit from them.  An exception is if an adjustment is needed that has no known provenance.  Examples of when this might happen: 
+* Beginning balances in a new system
+* Manual inventory count shows a different amount than what the system shows.
+
+Temporary note: The actions `use` and `work` are time-based actions, either with or without an explicit schedule.  How to handle time-based resources is still under discussion.  It is possible we will want to define the affect of those as `Decrement` in the future.
```
elf-pavlik: I think we should mention that we don't even know if we actually want to have time based `vf:Resource` or we will find Flows sufficient for represent usage, work and other services.

docs/introduction/flows.md
```diff
@@ -12,15 +12,44 @@ Intent describes some potential future event which might require prior Commitmen
...
+consume | Decrement | Process | Input | N/A |
+use | No effect | Process | Input | N/A |
+work | No effect | Process | Input | N/A |
+cite | No effect | Process | Input | N/A |
+accept | No effect | Process | Input | improve |
+give | Decrement | Transfer | Input | receive |
+
+We have defined a core set of actions, but expect that this will be extended with others. If extended, they should be defined as part of this or another formal vocabulary so that all can use them and assume the same meaning. 
+
+Note: Increment and decrement are usually not meant to be used directly, rather the actions inherit from them.  An exception is if an adjustment is needed that has no known provenance.  Examples of when this might happen:
```
elf-pavlik: I disagree with *Increment and decrement are usually not meant to be used directly*. We should clarify in detail what each of the action implies (how applications should interpret and reason over them) and just recommend to use action which has the most accurate implications.

docs/introduction/flows.md
elf-pavlik: This slide has *rental* event directly on the 'exhibition area' resource, it doesn't have 'rental of exhibition area' "resource". https://www.slideshare.net/Demetrius_Gallitzin/PavelHruby#16

docs/introduction/flows.md
elf-pavlik: #355 (hard vs. soft allocations) may need simple decrement on one resource and increment on another one, but since it does not change existence probably one can just use *give* (former *issue*) and *receive* without implying any changes to 'rights and responsibilities'

elf-pavlik:
@fosterlynn could you possibly add to the table *changes existence* column? If you prefer I could also add commit myself directly to the branch of this PR.

docs/introduction/flows.md
elf-pavlik: "How to handle schedules is still under discussion", I would also prefer to ~~It is possible we will want to define the affect of those as `Decrement` in the future.~~

docs/introduction/flows.md
fosterlynn: removed that sentence; removed from chart

docs/introduction/flows.md
elf-pavlik: I would still like to reformulate this note about schedules in direction i suggested above ^

docs/introduction/flows.md
```diff
@@ -30,36 +30,36 @@ Intents, Commitments, and Economic Events can occur at any granularity that is n
...
+* vf:unload -  transported resource leaves the process, the same resource will appear in input with *vf:load* verb
+* vf:accept - in processes like repair or maintentance, the same resource will appear in output with *vf:improve* verb
+* vf:improve - in processes like repair or maintentance, the same resource will appear in input with *vf:accept* verb
+* vf:give - give rights and responsibilities for the resource
+* vf:receive - receive rights and responsibilites for the resource
+* vf:raise - adjusts a quantity up based on beginning balance or inventory account adjustment
+* vf:lower - adjusts a quantity down based on beginning balance or inventory account adjustment
+
+Action | Affect | Related to | I/O | Changes existence | Pairs with |
+------ | ------ | ---------- | --- | ----------------- | ---------- |
```
elf-pavlik: latest version misses *increment* and *decrement*

docs/introduction/flows.md
fosterlynn: I was actually confused by the comment history, and am fine either way.  Now it says: "How to handle schedules is still under discussion".  Do you want to include one more sentence, the one "It is possible we will want to define the affect of those as Decrement in the future." ?

docs/introduction/flows.md
fosterlynn: I found a couple verbs for adjust: raise, lower.  So I don't think we need or want increment and decrement as actual actions, just abstract inheritance for behavior.  Make sense to you?  People should use the more defined actions imo, and define others as needed.

docs/introduction/flows.md
fosterlynn: actually, i'd like to do a PR and make this "reified" - if I am using the term correctly - add other properties to vf:Action, which would be a class instead of a property - along the lines of this chart - what do you think?

elf-pavlik: I think this would result in adding bunch of terms to the vocab just for the sake of represent that table, while this table seems mostly affecting implementations and not processing of data at run-time. I would prefer to finish this PR and if you see it a priority you could start another one but preferably not as part of *cleanup* milestone but as place to give consideration to this possibility.

fosterlynn: Agree, I didn't want to do that PR for this cleanup release.  Just explaining part of why I didn't want to include increment and decrement as actions in their own right.

elf-pavlik:
I think we those two changes this PR could land:

* [ ] add *increment* and *decrement* since we define them in vocabulary
* [ ] remove *Related to* column, we can always introduce it as part of [transfers milestone](https://github.com/valueflows/valueflows/milestone/5)

As for the second, if you like you could right after merging this one create PR that adds this column and assign it to *transfers* milestone.

fosterlynn:
>remove Related to column

done

Looking at the 'existence' column, wondering if actually decrementing or incrementing changes existence.  If so, i need to adjust the chart.

elf-pavlik:
>  Seems like consuming or producing could be the same (resource could exist throughout, just quantity changes)

If 'existance' doesn't change one shouldn't use *consume* and *produce* but for example *give* and *receive* which do not change existence.

I think *increment* and *decrement* should have
```
Action | Affect | I/O | Changes existence | Pairs with |
------ | ------ | --- | ----------------- | ---------- |
increment | Increment | Output | N/A | N/A |
decrement | Decrement | Input | N/A | N/A |
```

BTW *raise* and *lower* might also need `N/A` for 'change existence', food could have rotten or pests ate it,

fosterlynn:
Let's figure out what "change existence" means.  I can't tell from the original table, maybe need to re-read that paper.

But if it means new resource or gone resource, but also change of quantity to a resource (stock resource), then I need to change a few of those in the table.

PR #390, opened 2019-01-04
Explaining general status of the model/vocab.

fosterlynn:
DO NOT MERGE YET!!

I started updating the status page for the next release.  Mostly I wanted to explore how to think about the state of this vocabulary, how stable is it / can it be / do we want it to be, in general and in the next release in particular.  Discussion of this question would be welcome!

### DIFF
--- a/docs/introduction/status.md
+++ b/docs/introduction/status.md
@@ -2,23 +2,27 @@
 
 This documentation is for release 0.2.
 
-The relatively consolidated parts of the vocabulary are reflected below.  Those include: 
+The relatively consolidated parts of the vocabulary include: 
 * Agent model
 * Core Resource model
 * Core Input-Process-Output model
 * Core Commitments
-* Core Intents
-* Core exchanges, transfers, reciprocity
+* Beginning of Intents
+* Beginning of Exchange
 * Core Recipe model
 * Classifications
 
-Yet to come: 
+Some things yet to come: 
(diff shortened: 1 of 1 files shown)


docs/introduction/status.md
```diff
@@ -1,18 +1,24 @@
...
-* Recipes and Templates for processes and transfers
-* Intents (offers and requests)
+* Recipes - forking/versioning
+* Budgets, community analysis, plans
+* Transportation
 * Conversations for Action/Agreement and protocols
-* Serialized resources
 * Examples
+
+A note about the status.  This vocabulary is based on decades of work in academia and some limited work in the field in software implementations - and further development by the VF team through documenting and studying use cases.  The stability of the model reflects this history.  The core model is fairly stable.  However, there are many edge cases and possibly not-so-edge cases where the model will need to be adjusted for the reality in the field based on further experience. Also, this model represents something new, something that we want to support the next economy, with all the unknowns that will bring.
```
elf-pavlik: I'd like to add here approach to define status of each term in the vocab based on https://www.w3.org/2003/06/sw-vocab-status/note Of course if we agree on them:

* *unstable* - included in the vocabulary based on discussions
* *testing* - exemplified with data snippets based on diverse use cases
* *stable* - implemented in at least 2 open source applications which work with all included data snippets using that term

PR #391, opened 2019-01-06
starting documentation of traversing value flows

fosterlynn:
Capturing input from @bhaugen from issues and discussion, to start documentation on traversing resource flows.  This will need to be further developed as we nail down any changes/additions to transfer under development.

### DIFF
--- a/docs/appendix/track.md
+++ b/docs/appendix/track.md
@@ -12,4 +12,24 @@
 
 > When you "track" a cellphone, you monitor [the starting] location ... and follow it wherever it goes...
 
-One of the most-often used traces is to find the origins of health problems like mad cow disease and tainted drugs and food. In Value Flows terms, that would start with the product that caused a health problem, and tracing back along the chain of resource flows to the source or sources of the product.
+One of the most-often used traces is to find the origins of health problems like mad cow disease and tainted drugs and food. In Value Flows terms, that would start with the product (economic resource) that caused a health problem, and trace back along the chain of resource flows to the source or sources of the product.  Then it might be necessary to also track forwards from the source(s) to find everything else that might include that source(s).
+
+Here is the logic for tracking and tracing.
+
+* For an EconomicResource:
+    * after: EconomicEvents that are inputs to Processes or Transfers
+    * before: EconomicEvents that are outputs from Processes or Transfers
+* For a Process or Transfer:
+    * after: EconomicEvents that are outputs
+    * before: EconomicEvents that are inputs
+* For an EconomicEvent:
+    * after: 
+        * a Process or Transfer to which it is an input, or
+        * an EconomicResource which it affected as the output of a Process or Transfer
+    * before:
+        * a Process or Transfer from which is is an output, or
+        * if it is an input to a Process or Transfer, the EconomicResource which it affects.
+
+When the same economic resource is both input and output of a process, and more than one resource is involved, the tracking identifier is used.  And example of this would be partial load shipments, where many separate shipments are included in the same shipment process.
+
+When the same economic resource is both input and output of a process, sometimes a series of processes, such as for repair or quality testing or a workflow where a resource is refined through stages like writing/editing/etc, the stage must be identified, based on the kind of process the resource was last output of.
--- a/docs/appendix/track.md
+++ b/docs/appendix/track.md
@@ -17,8 +17,8 @@ One of the most-often used traces is to find the origins of health problems like
 Here is the logic for tracking and tracing.
 
 * For an EconomicResource:
-    * after: EconomicEvents that are inputs to Processes or Transfers
-    * before: EconomicEvents that are outputs from Processes or Transfers
+    * after: EconomicEvents affecting it that are inputs to Processes or Transfers
+    * before: EconomicEvents affecting it that are outputs from Processes or Transfers
 * For a Process or Transfer:
     * after: EconomicEvents that are outputs
     * before: EconomicEvents that are inputs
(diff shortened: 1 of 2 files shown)


docs/introduction/concepts.md
```diff
@@ -31,3 +31,27 @@ Producing and exchange olive oil and coffee, even more simplified.
...
+* For an EconomicResource:
+    * after: EconomicEvents that are inputs to Processes or Transfers
+    * before: EconomicEvents that are outputs from Processes or Transfers
+* For a Process or Transfer:
+    * after: EconomicEvents that are outputs
+    * before: EconomicEvents that are inputs
+* For an EconomicEvent:
+    * after: 
+        * a Process or Transfer to which it is an input, or
+        * an EconomicResource which it incremented
```
elf-pavlik: I think it may need to match *before*:
 if it is an output to a Process or Transfer, the EconomicResource which it affects.

Not all output actions on a resource have to increment it - eg. *improve*

elf-pavlik:
I think it might fit together with https://www.valueflo.ws/appendix/track.html

elf-pavlik:
I recall discussing *service* action which could reference two different processes:
* one using `vf:outputOf`
* second using `vf:inputOf`

We should remember in PR that may propose it to modify *after* and *before* of EconomicEvent to also reflect that use of *service* flow.

docs/introduction/concepts.md
fosterlynn: good point, will fix

fosterlynn:
>@fosterlynn your mother says it's "traverse" not "transverse".

Well I should always listen to my mother, so I'll fix it!

fosterlynn:
>I think it might fit together with https://www.valueflo.ws/appendix/track.html

Yes, thanks.  I'll get it all untangled in the documentation.  Do you think it should all go in that appendix entry?  I can move it there, and leave a small explanation up in the Resource Flows section, and/or a link.

fosterlynn:
>We should remember in PR that may propose it to modify after and before of EconomicEvent to also reflect that use of service flow.

Good point.  I don't know if we ever actually agreed on that, but I like it as a working assumption.  I think that it is covered in Bob's sequencing though, true?  It would find both before and after Processes OK?  (After the fix for "incrementing".)  Will push in one minute.

>

fosterlynn:
Fixes pushed.

docs/introduction/concepts.md
```diff
@@ -31,3 +31,27 @@ Producing and exchange olive oil and coffee, even more simplified.
 ### ... to thinking bigger.
 
 ![river flow](https://rawgit.com/valueflows/valueflows/master/release-doc-in-process/vf-river.png)
+
+### Traversing value flows
+
+Value flows can be traversed forwards ("tracing") or backwards ("tracking").  Often people use the term "provenance" when looking backwards towards the source of some resource, sometimes when a problem emerges (as in a disease outbreak from food), sometimes to know the quality of the resource (as in wanting local humanely produced food with lower ecological impact).
```
elf-pavlik: @bhaugen in https://github.com/valueflows/valueflows/pull/392#issuecomment-451946438
> In the usual meanings, although I get confused and so does everybody else but people who do it for a living, "track" means starting with a source and seeing where it went, like starting with a mad cow and see who got parts of it. So it's traversing forwards.

>"Trace" means you got a part of the mad cow and want to find out which animal it came from, what farm, etc. So it is traversing backwards. Provenance is similar to tracing.

I think you two might want to double check which one (track, trace) matches which one (forwards, backwards)

bhaugen: So anyway, @elf-pavlik and @fosterlynn ,the definitions in that intro seem to be backwards. Might be my fault....

fosterlynn:
>So anyway, @elf-pavlik and @fosterlynn ,the definitions in that intro seem to be backwards. Might be my fault....

No, my fault, they just aren't meaningful to me.  (Not saying we shouldn't use them if there is some standard terminology there, which I understand there is.)  I'll fix.

docs/introduction/concepts.md
```diff
@@ -31,3 +31,27 @@ Producing and exchange olive oil and coffee, even more simplified.
...
+    * before: EconomicEvents that are inputs
+* For an EconomicEvent:
+    * after: 
+        * a Process or Transfer to which it is an input, or
+        * an EconomicResource which it affected as the output of a Process or Transfer
+    * before:
+        * a Process or Transfer from which is is an output, or
+        * if it is an input to a Process or Transfer, the EconomicResource which it affects.
+
+When the same economic resource is both input and output of a process, and more than one resource is involved, the tracking identifier is used.  And example of this would be partial load shipments, where many separate shipments are included in the same shipment process.
```
elf-pavlik: pretty much on of the scenarios discussed in #392 which also shows variant where we have different resource on input and output but they need to keep distinct *provenance anchor* / *processing transaction id* / *tracking id*, i think we should remove that *same economic resource is both input and output* and just keep something like *both input and output reference the same tracking (processing transaction) identifier*

elf-pavlik:
Very early implementation https://github.com/valueflows/vf-cli.ts/blob/master/index.ts#L83-L141

fosterlynn:
I moved the detail down to the appendix.  Is this merge-able?

Status: I think what we have works for services that are both output and input.  We haven't defined detail of process stages for when the resource is the "same" before and after the process (accept/improve).  We may not agree on using the tracking identifier for partial load transportation.... so if not, I can take that out for now.  If we do agree that this is at least one way it can be done, I could integrate that paragraph into the logic bullet points (or not).

docs/appendix/track.md
```diff
@@ -12,4 +12,24 @@
...
 
-One of the most-often used traces is to find the origins of health problems like mad cow disease and tainted drugs and food. In Value Flows terms, that would start with the product that caused a health problem, and tracing back along the chain of resource flows to the source or sources of the product.
\ No newline at end of file
+One of the most-often used traces is to find the origins of health problems like mad cow disease and tainted drugs and food. In Value Flows terms, that would start with the product (economic resource) that caused a health problem, and trace back along the chain of resource flows to the source or sources of the product.  Then it might be necessary to also track forwards from the source(s) to find everything else that might include that source(s).
+
+Here is the logic for tracking and tracing.
+
+* For an EconomicResource:
+    * after: EconomicEvents that are inputs to Processes or Transfers
+    * before: EconomicEvents that are outputs from Processes or Transfers
```
elf-pavlik: I think we may wan to clarify "EconomicEvents affecting it that ...."

fosterlynn: good idea

docs/introduction/concepts.md
fosterlynn: I very much disagree with having the provenance anchor (either a process or transfer) as sub to the main process.  And I didn't understand your suggested sentence.  I tried to clarify the text without implying more than what we have.  See if this is OK, or if I am going too much in "my" direction without taking into account your concerns!  Will continue to fix per discussion...

Changed to:
```
When there is more than one economic resource going from input to output of a process without transformation of that resource (except for location), the tracking identifier is used.
```

fosterlynn:
Made those changes... see what you think.

Issue #148, opened 2016-08-13
Agents in relation to process events


fosterlynn:
We have a general question on if agents should have a place in processes, or do they just relate to transfers, i.e. which agent has which role (rights and responsibilities) in relation to a resource.  From a gitter chat, too hard to find it now. :disappointed:   But I think this is maybe the biggest conceptual hurdle left in the core model after our marathon discussions.....??

I think there is a lot of agency involved in processes, and several kinds of process related events require one or possibly two agents to have complete information.  For example work, load, unload. 

@elf-pavlik please add your thoughts or objections to this, I don't think I can do it justice and don't want to mis-represent.

bhaugen:
Unless you got some robotic agents, your process will need some human agents.

elf-pavlik:
> I think there is a lot of agency involved in processes, and several kinds of process related events require one or possibly two agents to have complete information. For example work, load, unload. 

If we model work as *vf:ServiceResource*, than we can reference agent as provider (owner doesn't seem to fit services well). For scheduled services (possibly modeled as [planned resources](https://github.com/valueflows/resource/issues/14#issuecomment-239677311)) we also need provider and recipient information.

If we want to capture the actual work of loading and unloading... again we have work which we can represent as *vf:ServiceResource* which in turn also can reference *load* and *unload* events (we would need to define property for it, but we will also need something like that for referencing *load* and *unload* from the transportation service - output of transportation process)

Maybe for each YAML snippet where you reference agent directly from a process event I will try to write alternative where we have a resource with a relationship to an agent? If you paste couple of links to existing YAML from other issues I could give it a try.

fosterlynn:
>If we want to capture the actual work of loading and unloading... again we have work which we can represent as vf:ServiceResource

I was thinking more about this too. I think you are right about load and unload - creation of value all comes down to work in the end.  It also takes work to create a service (and sometimes other inputs too).  

Load and unload provider, carrier, receiver might be just context agents?  Would need to do some examples to be sure, because different viewpoints might impact this, I'm not sure.  Context agents have agency too, but I think you are .

Then again some people might not want to record work, but only record what was created/produced.  Could be a service or a material item.

>If we model work as vf:ServiceResource

I'd like to understand better why you don't like vf:work?  Seems to me to clearly and directly describe what is actually happening.  Which is what I think rea:EconomicEvents should do.

[...]

PR #392, opened 2019-01-06
examples of ProvenanceAnchor (generalized version of what current Transfer tries to do)

elf-pavlik:
Let's consider `vf:ProvenanceAnchor` and `vf:anchoredWith` as working names.

It provides the same provenance tracking capability as using `vf:Transfer` but doesn't make any assumption about which actions flows will use.

It includes alternative to using sub processes in https://github.com/valueflows/valueflows/pull/386/commits/fab19b610b0548f7beef8df5f18fb29fcd17da13

We still need to see how `vf:anchoredWith` plays with using `vf:trackingIdentifier`.

I think to evaluate how it compares to just using granular sub processes we will need to evaluate algorithms and queries required to gather provenance information with either approach.

### DIFF
--- a/examples/provenance.yaml
+++ b/examples/provenance.yaml
@@ -1,0 +1,236 @@
+'@context':
+  '@vocab': https://w3id.org/valueflows#
+  xsd: http://www.w3.org/2001/XMLSchema#
+  qudt: http://qudt.org/schema/qudt/
+  unit: http://qudt.org/vocab/unit/
+  time: http://www.w3.org/2006/time#
+  time:inXSDDateTimeStamp:
+    '@type': xsd:dateTimeStamp
+  al: https://al.example/
+  bea: https://bea.example/
+  cloe: https://cloe.example/
+  faircoop: https://faircoop.example/
+
+'@graph':
+
+# https://github.com/valueflows/valueflows/issues/385
+
+  - '@id': al:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': EconomicResource
+    clasifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 100
+
+  - '@id': cloe:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    '@type': EconomicResource
+    clasifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 0
+
+  - '@id': bea:6b97b1be-8e07-44ac-82e5-214f1b2aaf33
+    '@type': EconomicResource
+    clasifiedAs: http://www.wikidata.org/entity/Q89 # apples
+    currentQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 0
+
+  # Al arrives with 3 apples. He gives them to Chloe, Chloe receives them.
+
+
+  - '@id': al:52f0e212-3c4f-4d27-b345-5e964c135824
+    '@type': ProvenanceAnchor
+
+  - '@id': al:b52a5815-fae9-43bf-be95-833b95dc0adb
+    '@type': EconomicEvent
+    anchoredWith: al:52f0e212-3c4f-4d27-b345-5e964c135824
+    action: give
+    affects: al:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    observedQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 3
+    
+  - '@id': cloe:b90b0b77-09a2-42e2-8bd4-e9ae2c1c6172
+    '@type': EconomicEvent
+    provider: al
+    receiver: cloe
+    anchoredWith: al:52f0e212-3c4f-4d27-b345-5e964c135824
+    action: receive
+    affects: cloe:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    observedQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 3
+
+# https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#transportation
+
+  # Al provides 100kg of apples to Cloe and she picks them up from him.
+
+  - '@id': al:02b39a30-3e04-4305-9656-7f261aa63c84
+    '@type': Process
+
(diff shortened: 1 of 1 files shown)


bhaugen:
"Provenance" has a different meaning and use cases from "tracking and tracing" and other use cases involving traversing value flows.

Provenance refers to the origin and history of something, often applied to scientific papers, paintings, or food. The traversal starts from the resource being analyzed and goes backwards from there, often forming a tree with more than one starting point.

In the usual meanings, although I get confused and so does everybody else but people who do it for a living, "track" means starting with a source and seeing where it went, like starting with a mad cow and see who got parts of it. So it's traversing forwards.

"Trace" means you got a part of the mad cow and want to find out which animal it came from, what farm, etc. So it is traversing backwards. Provenance is similar to tracing.

https://www.valueflo.ws/appendix/track.html

In cases of food and drug contamination, the analysis will go both ways: trace back to the source batch and ingredients, and then track forward to wherever that source batch went.

Another use case for traversing value flows is in a contribution economy like Sensorica where income is distributed according to the contributions to the creation of whatever brought the income.

elf-pavlik:
This PR does not have any impact on 'rights and responsibilities' currently re-discussed in #148 

> Or would process events have PA's basically when there is not transformation? So transportation or transfer?

In *Al splits out two piles of apples from two different resources* scenario just the process doesn't capture which pile got split out of which initial resource. Here PA's serve as alternative to nested Processes from last example in #386 

If we just try to rely on algorithm in #391 it would connect apples from split from resource A with resource B, just because splitting two apples got recorded as part of the same process even that apples didn't really get mixed. Possibly LOT identifier could get used to track provenance here but mentioned algorithm doesn't mention use of any tracking identifier.

I see need to reconcile proposed here `vf:anchoredWith` supposed to work and how value of `vf:trackingIdentifier` gets used in flows (currently we only specify that resource can have one).

In practice we just need all the flows that should stay consider as part of one *transaction* (provenance transaction nothing to do with rights and responsibilities) to reference the same identifier. Here they reference it with `vf:anchoredWith` but we could call it `vf:trackedWith` or `vf:tracedWith` - all working naming in here.

While sometimes we can't expect 'vf:trackingIdentifier' to have IRI as value (we define it as `owl:DatatypeProperty`). I see it useful that flows could use property defined as `owl:ObjectProperty`) that has IRI value and gets used for tracking/tracing provenance.

fosterlynn:
>Possibly LOT identifier could get used to track provenance here but mentioned algorithm doesn't mention use of any tracking identifier.

I wrote a sentence about it separate from the bullet points.  But yes, it would probably need to be better detailed out.

In general, I expect people who do lot tracking are pretty careful to not muddy the lots.  Since this example does not, I think trackingIdentifier would be fine.  If people combine lots, they create a new lot number though, so then you need the Process to not have all kinds of parallel activity.  Which is how transformations are supposed to work anyhow.  Al could split out his work.

And I think we can count on shippers to have tracking identifiers for everything on a mixed shipment.

So I tentatively don't think we need anything to tie events/resources across processes for tracking/tracing.  Could still be missing something.

So, I'm not sure we need anything like

elf-pavlik:
> Al could split out his work.

I don't think Al should have to do that, last example in #386 shows variant which uses two sub processes. In that case tracking/tracing algorithm #391 would need to do something like.

*If flow references multiple processes, use process which stays referenced by the least number of other flows*

fosterlynn:
>>Al could split out his work.

>I don't think Al should have to do that, 

"By Process, we mean an activity that transforms inputs into outputs." (definition from our doc)  I don't think processes are driven by work - although if Al wants to log his work all at once, that seems fine, if he doesn't need to know how much work went into each resource.  But then that should not drive construction of processes, which are fundamentally about creating outputs from inputs.  The process is the data structure that connects the inputs that created an output (maybe with co-products or by-products).

elf-pavlik:
> "By Process, we mean an activity that transforms inputs into outputs."
> The process is the data structure that connects the inputs that created an output (maybe with co-products or by-products).

The action defines if existence changes or not etc.
* Process with consume & create - one can consider as transformation
* Process with split (give|load) and merge (take|unload) one can consider as stock management
* use and work can apply to any process

What we discuss as transportation will have aspects of both
* transformation - consumes fuel / power from battery and creates CO2 or horsepower
* stock management - can split in one place and merge in other place

Also as discussed in #290 and #293 people can group flows in processes with different granularity. I think with `vf:includedIn` and `vf:service` (action) we can do a whole lot to provide flexible and unambiguous way to model all those scenarios.

elf-pavlik:
If #388 gets merged without `vf:trackingIdentifier`, this PR should address how tracking identifiers used by resources get propagated through flows and update that example. It will also affect #391

fosterlynn:
I would like to move this to the transfer milestone.  I don't think it can be considered separately. OK with you?

elf-pavlik:
I only mentioned current `vf:Tranfer` here because it only serves tracking provenance and doesn't really take any responsibilities related to 'rights and responsibilities' #148 

As part of the cleanup we may only need to provide way for `vf:trackingIdentifier` to propagate with flows. For example #388 doesn't make it clear how did `bob:0f563083-7f05-4a5d-a487-80e130a24345` ended up with `lot-4399jj93`

I think part of the issue relates to *composition* milestone where 'lower' level processes could get used to disambiguate provenance in 'higher' level processes, pretty much what `vf:Transfer` currently seem to do and what more general `vf:ProvenanceAnchor` proposed in this PR supposed to do.

fosterlynn:
>I think part of the issue relates to composition milestone where 'lower' level processes could get used to disambiguate provenance in 'higher' level processes, pretty much what vf:Transfer currently seem to do and what more general vf:ProvenanceAnchor proposed in this PR supposed to do.

Yeah, the 'higher' level processes is what I don't think we can include in this cleanup release, that has a bunch to work out yet.  I don't think that is what current vf:Transfer does, it is inline with other value flows in a directed graph, not part of a hierarchy.

elf-pavlik:
> As part of the cleanup we may only need to provide way for vf:trackingIdentifier to propagate with flows. For example #388 doesn't make it clear how did bob:0f563083-7f05-4a5d-a487-80e130a24345 ended up with lot-4399jj93

How about we just focus on that and stay away from *composition* and *transfer* for now?

fosterlynn:
>How about we just focus on that and stay away from composition and transfer for now?

Sounds good, I'll enlist @bhaugen to help document that....

fosterlynn:

Issue #300, opened 2018-11-27
Categorizing / classifying and specifying work and services.


elf-pavlik:
Capturing from https://github.com/valueflows/valueflows/pull/296#discussion_r236713077

> @elf-pavlik I would not use `ResourceSpecification` since we don't reference any resource from *work* flows and this would just make everything more confusing.
Only flows which at event stage will reference an actual resource with *affects* should use `vf:resourceConformsTo` and reference specification of a resource which event will reference using `vf:affects`.

elf-pavlik:
For quite some time I would like to explore 'expanding' *work* flow into a process that outputs a *service* and takes as input *work* time of a person as well as *use* of this person skills. This way `vf:ProcessRecipe` that will land with #296 could get used as clear specification of a service.

This might work with other services, let's say mowing a grass requires inputs
* 1h work
* 1h use of string trimmer https://www.wikidata.org/wiki/Q1340233

Recipe of that process seems like a clear way to specify a service. While for product static `vf:ResourceSpecification` makes sense, for dynamic services (which for me include work) `vf:ProcessRecipe` itself looks reasonable to specify / define a service.

elf-pavlik:
>> Could you please amend that fully automated post creation example to add some work flows?
```
ResourceSpecification
--id: 'rs1-post'
--name: 'Post'

ResourceSpecification
--id: 'rs2-image'
--name: 'Image'

ResourceSpecification
--id: 'rs3-text'
--name: 'Text'

ResourceSpecification
--id: 'rs4-designwork'
--name: 'Graphic Design'

ResourceSpecification
--id: 'rs5-writingwork'
--name: 'Writing (EN)'

RecipeProcess
-- id: 'recipeprocess1'
-- name: 'Write Post'

RecipeProcess
-- id: 'recipeprocess2'
-- name: 'Create Image'

RecipeProcess
-- id: 'recipeprocess3'
-- name: 'Format Post'

RecipeFlow
--id: 'recipeflow1'
--action: produce
--resourceConformsTo: rs1-post
--outputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow2'
--action: produce
--resourceConformsTo: rs2-image
--outputOf: recipeprocess2

RecipeFlow
--id: 'recipeflow3'
--action: cite
--resourceConformsTo: rs2-image
--inputOf: recipeprocess3

RecipeFlow
--id: 'recipeflow4'
--action: produce
--resourceConformsTo: rs3-text
[...]

Issue #177, opened 2017-02-15
Process vs Exchange Fulfillment


fosterlynn:
In combining the process and exchange models that we have more or less agreed on, it became clear that we have different ways of modeling event fulfillment of commitment in process vs exchange. 

Exchange here: https://github.com/valueflows/exchange/issues/32#issuecomment-267834962

Process: We have events directly fulfilling commitments.  1:M, commitment to event.  (For exchanges, M:M.)

This issue is to take some specific use cases and work them through in detail to see how this should work.

For example, someone commits to do some work for a process.  They also commit to trade this work for a bushel of apples.  They do the work, recording an event. The work fulfills the commitment to do work.  The work also fulfills the commitment to "transfer" (little t) the work as part of the exchange......

elf-pavlik:
> The work fulfills the commitment to do work. The work also fulfills the commitment to "transfer" (little t) the work as part of the exchange......

I don't see 2 commitments here but only one, to do some work.

fosterlynn:
So I suppose the commitment belongs to both a process and an exchange then?  Could work.

elf-pavlik:
When it comes to work, the only case where we would see *provider* & *receiver* the same would happen when someone just tracks work requested by oneself and committed to oneself. In any case where we have work with different agents referenced as *provider* and *receiver* one can consider it implying transfer between those agents.

fosterlynn:
>In any case where we have work with different agents referenced as provider and receiver one can consider it implying transfer between those agents.

That makes sense when the work will actually be involved in a direct exchange with another agent.

But what about the cases where people are working on a process together?  Let's say an open source project for the commons.  Or in Sensorica working on an R&D process.

elf-pavlik:
> But what about the cases where people are working on a process together? Let's say an open source project for the commons. Or in Sensorica working on an R&D process.

I don't see why this changes anything, as long as we have event with different agent referenced as *provider* and *receiver* I think one can interpret that as 'transfer'. It may or may not stay included in some Agreement or at some point have Appreciation but I don't see why this would matter.

If we look at VF, work events could have *provider* but wouldn't have any agent referenced as *receiver*.

fosterlynn:
> In any case where we have work with different agents referenced as provider and receiver one can consider it implying transfer between those agents.

I think this makes sense.  Let's see how it plays out as we make yaml with provider and receiver.

fosterlynn:
[...]

PR #393, opened 2019-01-07
initial service examples

elf-pavlik:
First YAML examples inspired by those recent conversations:
* https://github.com/valueflows/valueflows/pull/391#issuecomment-451949728
* https://github.com/valueflows/vf-apps-traversing-the-flows/issues/4#issue-395979472
* https://github.com/valueflows/valueflows/issues/290#issuecomment-447584697
* #386 
* #392 
* #300

### DIFF
--- a/examples/service.yaml
+++ b/examples/service.yaml
@@ -94,6 +94,13 @@
     '@type': EconomicResource
     classifiedAs: http://www.wikidata.org/entity/Q21002847 # Faircoin
 
+  - '@id': miner:566a2c01-96f8-4b3c-adcf-f197eced3092
+    '@type': EconomicResource
+    skos:note: computing unit
+    currentQuantity:
+      qudt:unit: unit:Unitless
+      qudt:numericValue: 1 
+
   - '@id': miner:23799c14-c368-4653-a584-83bf9ae8b82a
     '@type': Process
     skos:note: transaction processing
@@ -141,6 +148,17 @@
       qudt:numericValue: 1
     provider: alice
     receiver: miner
+
+  - '@id': miner:8791913c-0152-4c32-bb3f-f25d61df736c
+    '@type': EconomicEvent
+    action: use
+    affect: miner:566a2c01-96f8-4b3c-adcf-f197eced3092 # computing unit
+    inputOf: miner:23799c14-c368-4653-a584-83bf9ae8b82a
+    observedTime:
+      time:intervalStarts:
+        time:inXSDDateTimeStamp: 2019-01-02T02:10:43.003-5:00
+      time:intervalEnds:
+        time:inXSDDateTimeStamp: 2019-01-07T02:10:43.007-5:00
 
   - '@id': miner:e72f7d1d-4f0a-4518-a44c-96658cafda95
     '@type': EconomicEvent
--- a/examples/service.yaml
+++ b/examples/service.yaml
@@ -11,6 +11,8 @@
   claudia: https://claudia.example/
   notary: https://notary.example/
   miner: https://miner.example/
+  jesus: https://jesus.example/
+  joe: https://joe.example/
 
 '@graph':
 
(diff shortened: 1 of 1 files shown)


elf-pavlik:
I may need to make a visual diagram for snippet in the last commit https://github.com/valueflows/valueflows/pull/393/commits/f6bd73357c4d7600b1cd071162dc68cbcbb43651

It should feed a lot into conversations happening in various issues and PRs

elf-pavlik:
service

elf-pavlik:
@fosterlynn I've added use of computing unit as input to processing faircoin transaction snippet

@sqykly from what I understand Holo currency will get exchanged for computing power, have you thought on how to represent those flows (economic events, commitments) which use computing power?

fosterlynn:
I'm worried about the complexity of "included in" kinds of things on the operational level (as opposed to a summary level, which might be a different story). Seems like it messes with the simplicity of the base directed graph structure of the value flows.  I'll try an alternative treatment of this use case (and some other PRs that currently have this structure), to see if I can make it better or there are devils I haven't thought of.

fosterlynn:
What are the definitions of actions "split" and "merge", and what do they replace?

elf-pavlik:
> What are the definitions of actions "split" and "merge", and what do they replace?

Please think of them as plain *decrement* and *implement* which does not 'change existence'. I could have used *pop* and *push* instead.

elf-pavlik:
I made a simple script for *track* from #391 and for diagram above when we start from *Alice's apples* it gives me correct: 

screenshot at 2019-01-10 18-53-14

It needs to keep track of visited nodes not to fall into looping over a cycle. I'll clean it up on Saturday and push a new repo to this organization. It does not rely on any specific serialization (JSON-LD, Trig, N-quads) and just builds on http://rdf.js.org/ interfaces.

Tree representation doesn't work that well since we have a graph, for now I'll just add a simple way to mark where nodes would start repeating. Above the last event in the bottom also goes as input to process on line 5.

fosterlynn:
Here's a more elaborate treatment, which will be able to support inventory on the truck and lot numbers in a more specific way. ( "Merge" doesn't seem that good, because if Joe was going to use the fruit for anything that would go to others, he would keep them as separate resources to maintain the lot numbers coming from the growers.) 
delivery-service-2

elf-pavlik:
In your last diagram:
* what quantity would you record as *unload*? just the quantity that Joe will take home? 
* can Jesus record work time for each of those small 4 processes related to specific operations on apples and pears?
* how would you see Jesus recording a distinct process for each 'processing of transaction with a consumer' which could have a work event specific to handling the whole transaction with a specific consumer? he may need that data for example to plan route based on statistics how long transactions with each of the consumers tend to last.

fosterlynn:
>what quantity would you record as unload? just the quantity that Joe will take home?

yes

>can Jesus record work time for each of those small 4 processes related to specific operations on apples and pears?

if he wants -- not sure it would make sense for this use case since he is not splitting out his other inputs, but no problem with it, even if they are transfers imo

>how would you see Jesus recording a distinct process for each 'processing of transaction with a consumer' which could have a work event specific to handling the whole transaction with a specific consumer? he may need that data for example to plan route based on statistics how long transactions with each of the consumers tend to last.

that is included in the last version, no?

elf-pavlik:
> that is included in the last version, no?

No you have two separate processes to split apples for Joe and split pears for Joe. I see a case where Jesus would want to record the whole transaction with Joe as one process which would involve splitting any number of resources and negotiating agreements based on available proposals. We can skip that negotiating agreements part for now and at least have one work event related to splitting both apples and pears for Joe, just as in my diagram for "Transaction with Joe" process.

fosterlynn:
>No you have two separate processes to split apples for Joe and split pears for Joe

I suppose you could make one process, and rely on the tracking identifiers.  
deliver-joe

Better maybe to just log each separate delivery work event against the main process.

But, I think almost anything is better than adding the complication of nesting this type of process.

fosterlynn:
>If we define vf:includedIn and vf:service as a way to nest flows and relate processes.

I don't think that we should define vf:includedIn, haven't seen a need for it yet.  We may find a need when we talk about non-operational higher-level processes, which is where it first came up.  But what I have seen so far operationally is that it adds more complexity than it is worth.

In terms of vf:service (as an action, right?) - I think it fits fine into the directed graph structure we have now without any nesting.  Should work very much like vf:produce, an output to a process whose inputs were used or consumed to create the service.  In the occasional cases where a service is created and is also input to the next process, then we talked about trying out using it as both an output and an input without the resource in between.  But that is still basically a flow, not a hierarchy, in terms of structure.

fosterlynn:
>It looks like one of the approaches for #177 (comment) also to log more specific work events happening during longer work session.

Not quite sure of the implications to what you are considering, but our experience is that people often log a lot of work events for one commitment.  Sometimes this is logging every working session, which is convenient.  Or logging different pieces of the "task" so it is recorded separately for whatever reason (summary of workflow, communication of stages of the work, etc.).

But that doesn't mean that we need a summary work event for those.  As you point out, it is handled by the M:M between event and commitment.

fosterlynn:
>If you prefer I could separate all the vf:includedIn occurrences into a separate PR based on discussions in #290 and #293

That might be helpful.  Or if there is one PR it is needed as part of, it can stay in that one.  

I'm thinking the examples are not a good place to introduce new things though?  Proposal: PRs for examples only include accepted vocabulary, or vocabulary that is defined in another PR where the spec is updated, instead of just examples.  OK with you?

elf-pavlik:
> I'm thinking the examples are not a good place to introduce new things though?

I think we should never introduce new things if PR doesn't include examples. Possibly we can always start with examples and once we see consensus arising, then add missing terms and documentation to that PR.

fosterlynn:
>I think we should never introduce new things if PR doesn't include examples.

This also seems like a good idea.  But changes need explanation, at least a definition. So let's figure out how to do both in PRs?

fosterlynn:

PR #404, opened 2019-01-17
Matching UML and ttl, other tweaks per discussions.

fosterlynn:
Note: UML will have to be changed again, but it is a a lot better than what is there now.


fosterlynn:
cut a webvowl pic also, just for now

PR #407, opened 2019-01-19
removed Transfer

elf-pavlik:
Currently we discuss how to model *transfer of rights and responsibilities* in #148 as well as we have a dedicated [*transfer milestone*](https://github.com/valueflows/valueflows/milestone/5)

At this moment `vf:Transfer` only attempts to track provenance in very special case, at the same time #392 tries to look for general way of tracking provenance and very likely we will need to defer it to [*composition milestone*](https://github.com/valueflows/valueflows/milestone/3)

#389 clarifies definitions of actions, including which do *change existence* and which not. Given that one can easily filter processes which have aspect of transformation from those which do not. Also based on #148 and *transfer milestone* one will have a way to filter processes referenced from flows that involve any kind of transfer.

Given above and in addition as shown in #393 that some processes currently recommended to record as `vf:Transfer`, will require flows for *use* and *work*, it seems reasonable to simple use Process for all those scenarios and document how to reason weather any Process that have aspect of transformation, transportation, transfer etc.

NOTE: github doesn't present diff very well here, this PR mostly rearranges order of sections and makes minor edits to few of the sections.


fosterlynn:
>At this moment vf:Transfer only attempts to track provenance in very special case

Between Transfer and Process, provenance can be tracked in all cases as far as I know.  That was certainly the idea.

elf-pavlik:
If one replaces every instance of `vf:Transfer` with an instance of `vf:Process` how does it affect tracking & tracing (provenance) ?

> #389 clarifies definitions of actions, including which do change existence and which not. Given that one can easily filter processes which have aspect of transformation from those which do not.

once again one can filter flows based on their actions so we don't seem distinct Process and Transfer for that functionality

fosterlynn:
Transfer is gone now.

PR #411, opened 2019-01-23
First simple example: agents

fosterlynn:
Starting with agents for simple examples.

Questions I have: 
* what allows us to say `alice` in something like this? `subject: alice`   Can we do the same kind of thing for something like relationship? `relationship: fablab:52f0e212-3c4f-4d27-b345-5e964c135824 # member`  
* should we review the rdfs:label and vf:inverseLabel in AgentRelationshipRole?  Like, should we have vf:label with range of rdfs:label?  Seems like it might be easier on people?

Any other things I am not doing right technically, in the yaml, please let me know @elf-pavlik !!

### DIFF
--- a/examples/agent-examples.yaml
+++ b/examples/agent-examples.yaml
@@ -1,0 +1,71 @@
+# Example: Agents and Agent Relationships
+
+'@context':
+  '@vocab': https://w3id.org/valueflows#
+  rdfs: http://www.w3.org/2000/01/rdf-schema#,
+  alice: https://alice.example/
+  bob: https://bob.example/
+  fablab: https://fablab.example/
+  coop: https://coop.example/
+
+'@graph':
+
+  - '@id': alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': Person
+    name: 'Alice'
+    image: 'https://alice.example/avatar.png'
+    url: 'https://alice.example'
+    primaryLocation: 'https://somelocation.example'
+    note: 'Alice is a mechanical engineer who likes to work in the fablab.'
+
+  - '@id': bob:b90b0b77-09a2-42e2-8bd4-e9ae2c1c6172
+    '@type': Person
+    name: 'Bob'
+
+  - '@id': fablab:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    '@type': Organization
+    name: 'Driftless Fablab'
+
+  - '@id': coop:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    '@type': Organization
+    name: 'Community Tool Lending Coop'
+
+  - '@id': fablab:52f0e212-3c4f-4d27-b345-5e964c135824
+    '@type': AgentRelationshipRole
+    name: 'member'
+    rdfs:label: 'is member of'
+    inverseLabel: 'has member'
+    note: 'Both persons and organizations can be members of this fablab.'
+
+  - '@id': fablab:02b39a30-3e04-4305-9656-7f261aa63c84
+    '@type': AgentRelationshipRole
+    name: 'trading partner'
+    rdfs:label: 'is supplier of'
+    inverseLabel: 'is customer of'
+
+  - '@id': fablab:a25500e0-0106-43cd-8cbb-e74779488835
+    '@type': AgentRelationshipRole
+    name: 'mentor'
+    rdfs:label: 'mentors'
+    inverseLabel: 'has mentor'
+
+  - '@id': fablab:6b97b1be-8e07-44ac-82e5-214f1b2aaf33
+    '@type': AgentRelationship
+    subject: alice
+    relationship: fablab:52f0e212-3c4f-4d27-b345-5e964c135824 # member
+    object: fablab
+
+  - '@id': fablab:a8236bbb-81e0-422d-9861-56d2417db0fb
+    '@type': AgentRelationship
+    subject: coop
+    relationship: fablab:02b39a30-3e04-4305-9656-7f261aa63c84 # trading partner
+    object: fablab
+    note: 'The coop is a supplier of tools for the fablab.'
+
+  - '@id': fablab:6f438393-7f87-4914-806c-e23a4fd15e89
+    '@type': AgentRelationship
+    subject: alice
+    relationship: fablab:a25500e0-0106-43cd-8cbb-e74779488835 # mentor
+    object: bob
+    inScopeOf: fablab
+    note: 'Alice mentors Bob at the fablab.'


fosterlynn:
Another question: when to use quotes and when not to?

examples/agent-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - '@id': alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': Person
+    name: 'Alice'
+    image: 'https://alice.example/avatar.png'
+    url: 'https://alice.example'
+    primaryLocation: 'https://somelocation.example'
+    note: 'Alice is a mechanical engineer who likes to work in the fablab.'
+
+  - '@id': bob:b90b0b77-09a2-42e2-8bd4-e9ae2c1c6172
+    '@type': Person
```
elf-pavlik: since `vf` doesn't define Person I think we should use `foaf:Person` to avoid confusions

examples/agent-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+    name: 'Bob'
+
+  - '@id': fablab:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    '@type': Organization
+    name: 'Driftless Fablab'
+
+  - '@id': coop:3129ca8b-fcda-45be-bbda-294dc924d3b9
+    '@type': Organization
```
elf-pavlik: since `vf` doesn't define Organization I think we should use `org:Person` to avoid confusions

examples/agent-examples.yaml
```diff
@@ -0,0 +1,71 @@
+# Example: Agents and Agent Relationships
+
+'@context':
+  '@vocab': https://w3id.org/valueflows#
+  rdfs: http://www.w3.org/2000/01/rdf-schema#,
+  alice: https://alice.example/
```
elf-pavlik: Instead of three lines above we can use 
```yaml
  - https://git.io/vf-examples-jsonld-context
  - alice: https://alice.example/
```
lines below just need to align bob and alice, this way we have array with two elements, IRI to external context and object with local prefixes for agents

elf-pavlik:
To stay honest, I wouldn't like to dedicate to much time to all that roles and relationships between agents. It doesn't impact anything else in the vocab and I would like that at some point we reconsider if we even want to include any of that in core VF.
We already rely on `foaf:` and `org:` and possibly could just leave all that roles and relationships stuff to dedicated vocab, it doesn't seem to tie directly to anything else in VF anyways.

examples/agent-examples.yaml
fosterlynn: ok - so we would need to make one context that all examples could use, right?

elf-pavlik: yes, I made https://git.io/vf-examples-jsonld-context to use it in all examples and keep them consistent

fosterlynn:
>To stay honest, I wouldn't like to dedicate to much time to all that roles and relationships between agents. 

I haven't seen any agent model that is as flexible and simple as this one, so I'd like to keep it.  Anyway, not much work left, some tweaks here and there.  And we spent a ton of time on it early on.

Also, we'll have agents all over the place - so it would be nice to be able to refer to them in the same way, no?  And there is REA.... where "Economic Agent" is defined.

examples/agent-examples.yaml
elf-pavlik: please note that first two elements need `-` since we create an array here

elf-pavlik:
let's do each file as separate PR

examples/agent-examples.yaml
```diff
@@ -0,0 +1,76 @@
...
+  bob: https://bob.example/
+  fablab: https://fablab.example/
+  coop: https://coop.example/
+
+'@graph':
+
+  # Agents
+
+  - '@id': alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
```
elf-pavlik: here you also need to use just `alice:` if you want to later reference that person with just `alice:`

fosterlynn:
>let's do each file as separate PR

oops just saw this

examples/agent-examples.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+'@graph':
+
+  # Agents
+
+  - '@id': alice:
+    '@type': foaf:Person
+    name: 'Alice'
+    image: 'https://alice.example/avatar.png'
+    url: 'https://alice.example'
```
elf-pavlik: can we remove this `url` not to cause confusion on how does it differ from `@id`?

examples/agent-examples.yaml
```diff
@@ -0,0 +1,76 @@
...
+    name: 'Community Tool Lending Coop'
+
+  # Roles
+
+  - '@id': fablab:52f0e212-3c4f-4d27-b345-5e964c135824
+    '@type': AgentRelationshipRole
+    name: 'member'
+    rdfs:label: 'is member of'
+    inverseLabel: 'has member'
+    note: 'Both persons and organizations can be members of this fablab.'
```
elf-pavlik: don't we use `skos:note` ?

fosterlynn: hmmm - i think this one is defined as vf:note with range of skos:note -- what is the best way to do this kind of thing?

elf-pavlik: > defined as vf:note with range of skos:note

such definition makes no sense, let's just use skos:note directly

fosterlynn: ok

fosterlynn: you can cover that in the context file, right?

bhaugen:
@elf-pavlik 
> To stay honest, I wouldn't like to dedicate to much time to all that roles and relationships between agents. It doesn't impact anything else in the vocab and I would like that at some point we reconsider if we even want to include any of that in core VF.
> We already rely on `foaf:` and `org:` and possibly could just leave all that roles and relationships stuff to dedicated vocab, it doesn't seem to tie directly to anything else in VF anyways.

Agents tie into everything in VF. We've used AgentRelationships using the same model as now in VF, although the names varied, in several systems which have been used and forked by many people, and several now have VF APIs. They will also be important for all of the new projects we are working on that will be based on VF from the beginning.

This part of our whole practice of spiraling between working with concrete groups and then back to summarize and generalize, which we have done in VF since it started.

So we need AgentRelationships. And we also need Transfers and Intents.

As far as I know, we are the only people who have developed and deployed systems using all of the VF concepts, and they are used by real people. Every concept we have introduced or suggested in VF comes from real experience.

elf-pavlik:
> Agents tie into everything in VF.

Currently only `foaf:Agent`, `foaf:Person` and `org:Organization` 'tie into other things', all the rest 'agent related' currently included in VF vocab doesn't really impact anything in the rest of the vocab. If another vocab defined all that stuff, not keeping anything from VF in mind (or not even knowing of VF existence) it would make no difference on the rest of the VF as currently defined.


> They will also be important for all of the new projects we are working on that will be based on VF from the beginning.

Which other vocabs / ontologies besides VF this projects will use?

bhaugen:
> > Agents tie into everything in VF.
> 
> Currently only `foaf:Agent`, `foaf:Person` and `org:Organization` 'tie into other things', all the rest 'agent related' currently included in VF vocab doesn't really impact anything in the rest of the vocab. If another vocab defined all that stuff, not keeping anything from VF in mind (or not even knowing of VF existence) it would make no difference on the rest of the VF as currently defined.
> 

That's a limited view of how those relationships work. For example, we will need to define access, validation, agreements, countersignatures, distributed transactions, and distributed consensus, among other functions, for at least 3 and maybe 4 projects that are now in motion.  All of the more purely economic interactions between agents will be affected by those security and validity related functions. And the agent relationships will be used in defining them. So:
* We have found no other adequate model of agent relationships, and
* if we did, and it was not connected to the rest of VF, it would not help.

> > They will also be important for all of the new projects we are working on that will be based on VF from the beginning.
> 
> Which other vocabs / ontologies besides VF this projects will use?

Several that have been mentioned or referenced in VF. We will need at least a units of measure vocab, and qudt seems to be having trouble with their upgrade and findability, and also vocabs for locations and time. But qudt, for example, is explicitly referenced in VF, so it's clear how it ties in. Assuming an adequate agent relationship vocab could be found, it would need to be similarly tied in. 

Overall, we need a whole ontology for economic networks. Can be composed of different vocabs, but needs to be explicitly composed to provide a whole computable model for network planning, operations, accounting, and analysis.

PR #412, opened 2019-01-23
Simple examples: process (manufacturing)

fosterlynn:
moved process mfg example to separate PR


fosterlynn:
Fixed format.  And see comment in the other process issue re. Unitless.

PR #413, opened 2019-01-23
Simple examples: process (workflow)

fosterlynn:
moved process workflow example to create separate PR

### DIFF
--- a/examples/process-workflow-examples.yaml
+++ b/examples/process-workflow-examples.yaml
@@ -3,8 +3,9 @@
 '@context':
   - https://git.io/vf-examples-jsonld-context
   - alice: https://alice.example/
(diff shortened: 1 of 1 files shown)


fosterlynn:
Fixed format.  

Also, in this one and the other process one, I think we need to deal with the QUDT "Unitless" unit.  That is just confusing.  

@bhaugen  did some looking around for other similar vocabularies, hasn't found anything worth considering.  But, he also found this: http://qudt.org/schema/qudt.  Look down a little at Counting Unit, with examples including Number.

I don't like Number too well, but I like it a lot better than Unitless.

Bob also took a look at some ERP systems, and said that all of them are still using Each.

Here is another possible point of research: https://en.wikipedia.org/wiki/ISO/IEC_80000.

PR #415, opened 2019-01-24
external terms

elf-pavlik:
`rdfs:range skos:note` doesn't make any sense, we should simply use `skos:note` directly

### DIFF
--- a/docs/specification/external-terms.md
+++ b/docs/specification/external-terms.md
@@ -1,0 +1,22 @@
+# External terms
+
+Most applications using VF will need terms defined in various other vocabularies. Below we recommend reusing some from well established vocabularies (web ontologies).
+
+## SKOS
+
+https://www.w3.org/TR/skos-reference/
+
+* [`skos:note`](https://www.w3.org/TR/skos-reference/#note)
+
+## FOAF
+
+http://xmlns.com/foaf/spec/
+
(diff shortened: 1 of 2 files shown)


fosterlynn:
Question: how is note different than image or url?  I think we want to include those kind of things where appropriate in the model, so people don't have to guess at using other vocabs.... but how best to do it?  Maybe owl:equivalentClass or similar (if I remember that right)?

elf-pavlik:
attempted definition suggests that it supposed to match `skos:note` so it seems reasonable just to go with that. i think we will need to do something about `url` and `image` at some point but i don't see them as obvious as this one.

fosterlynn:
How do we signal that a note (or description or whatever) is part of this vocabulary and appropriate (or really needed) in some classes?  I can do it in the UML, but otherwise it doesn't show up anywhere.  We have the same for rdfs:label in AgentRelationshipRole.  I guess maybe examples can help.  Can we just put skos:note into the turtle?  Like foaf:Person?

It occurs to me that this might be why vocabularies tend to re-use the same concepts but put them into the vocabulary itself instead of just using them from other vocabularies?

elf-pavlik:
I think most people will follow examples, we can also add list of recommended terms somewhere in documentation.

elf-pavlik:
#424 explains that one should not expect to only use VF, in further PRs we should add list of recommended terms from other vocabs

PR #416, opened 2019-01-24
first cut flows example

fosterlynn:
intent-satisfaction-commitment-fulfillment-event


elf-pavlik:
nice!

could we remove `coworking` here and maybe have `mfg` having that initial produce 10 and they would just produce some themselves and get commitment from `maker` to produce 5 ?

PR #417, opened 2019-01-25
resource example

fosterlynn:


examples/resource-examples.yaml
```diff
@@ -0,0 +1,42 @@
+# Example: Resources
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - farm: https://farm.example/
+    ose: https://www.opensourceecology.org
```
elf-pavlik: misses trailing `/`

examples/resource-examples.yaml
```diff
@@ -0,0 +1,42 @@
...
+    name: Tractor - LifeTrac 6
+    trackingIdentifier: 889jcd00s # the serial number assigned by the manufacturer
+    resourceConformsTo: ose:822b6baf-d66f-45a0-98f2-57450e2aec23
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': farm:07819789-dd51-44c3-b35c-9210165bc832
+    '@type': EconomicResource
+    name: 12 kg case organic carrots
```
elf-pavlik: can we start without cases, just 600kg of carrots and later we can add variant which expresses 12kg cases in machine readable way

examples/resource-examples.yaml
fosterlynn: done thanks

examples/resource-examples.yaml
fosterlynn: done

fosterlynn:
Fixed/changed per review.  

Also made a list of classifiedAs links, could you please check and see if the format/syntax is right?  Thought it might be good to get that into an example.  Looked for agrivoc classifications, which would be better, but couldn't parse their website in a reasonable amount of time.

examples/resource-examples.yaml
```diff
@@ -0,0 +1,44 @@
...
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': farm:07819789-dd51-44c3-b35c-9210165bc832
+    '@type': EconomicResource
+    name: Loose organic carrots
+    trackingIdentifier: 54988 # the lot number assigned by the farm
+    classifiedAs: 
+      - https://www.wikidata.org/wiki/Q81 # carrot
+      - https://www.wikidata.org/wiki/Q9566427 # organic certification
```
elf-pavlik: Maybe let's not experiment with it at this moment? Some agent may offer organic certification for produce and that service could have this classification. Just using `Q81` seems enough to get this use case started.

fosterlynn:
removed organic cert

Issue #306, opened 2018-12-13
Named bi-directional properties


fosterlynn:
From @elf-pavlik in another issue:
>inverse properties just complicate queries, one has to always query for both directions to make sure all results get returned. 

>on can always adjust position of object and subject in the statement without need to have inverse property, in JSON-LD @reverse keyword provides some convenience

An example is Process.inputs and EconomicEvent.inputOf.

I have never tried sparql, so I don't know how it works.  Why would you have to query both directions?

From the non-rdf world, it seems like having named properties for both directions will be helpful, perhaps essential, as people will need to approach querying from either direction.  For example, I just wrote a graphql api using VF to serve a UI app that wants to be able to plug into any VF backend.  I ended up making up a lot of names (some of them not so good in retrospect) to accommodate many "other direction" queries.

If we truly can't have a named reverse property, we will need to decide if we do relational (foreign key) or OO (containers) direction as a consistent choice.  Does rdf have a preference?

elf-pavlik:
> An example is Process.inputs and EconomicEvent.inputOf.

I don't know how `vf:inputs` slipped in. Besides using anti-pattern of plural name, it would force everyone to check for both `vf:inputs` and `vf:inputOf`.

> Why would you have to query both directions?

Because data any creator could have used either one, and as a consumer of all that data we now must to check both options. For example if Alice uses `vf:inputs` in her data and Bob uses `vf:inputOf` in his data, for my application to work with Alice's, Bob's and everyone else's data I have to always check those both possibilities.

> If we truly can't have a named reverse property, we will need to decide if we do relational (foreign key) or OO (containers) direction as a consistent choice. Does rdf have a preference?

Could you please provide an example to clarify what you meant? In RDF we have an underlying model of *directed labeled graph* where properties / predicates get used as labels on those directed edges. The whole data model then boils down to making *Statements* (represented as triples), on can state
```
Bob likes pineapples  (subject: Bob predicate: likes object: pineapples)
```
While for 'human consumption' it might sometimes make more sense to say something like  
```
pineapples are liked by Bob (subject: pineapples, predicate: are liked by, object: Bob)
```
for the machine we just need to provide a triple that will get used in reconstructing this underlying *directed labeled graph*. Having property / predicate just in one direction, one can always make appropriate statement by choosing what to use as subject and what as object in a statement.

This works very similar to *entity attribute value* model, just the full description of certain entity will need to include both cases, where we have it in entity (subject) position and where we have it in value (object) position.

[...]

PR #418, opened 2019-01-25
direction change of input/output relationships

fosterlynn:
I want to change some of the directions in our relationships, as discussed previously.  Starting with inputs and outputs, making them "collections".

Next will try to do fulfillment, satisfaction, and maybe reciprocity.  Will see how it goes and do as a separate PR.  I think those 2 sets are the main ones that would be more useful as "collections" rather than ERD style.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -161,39 +161,71 @@ vf:action
         rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent vf:RecipeFlow) ] ;
         rdfs:range          vf:Action ;
         vs:term_status      "unstable" ;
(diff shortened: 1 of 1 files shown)


fosterlynn:
>This PR would require some implementations to update Process every time new flow gets added.

That's a good point.  I was thinking about the query aspect more.

So, in that case I think we should define selective inverse terms, like schema.org does, or whatever is the recommended method.

elf-pavlik:
> So, in that case I think we should define selective inverse terms, like schema.org does, or whatever is the recommended method.

We have #306 to decide how we want to document suggested naming for reverse directions. Maybe you could close this PR and we look at creating another one that would address #306 ?

elf-pavlik:
Can we close this one?

PR #419, opened 2019-01-25
changed direction of fullfills and satisfies

fosterlynn:
Made them from commitment to event; and from intent to commitment.


fosterlynn:
>Both Fulfillment and Satisfaction act as Qualified Relations. 

Yes, I know.  If it is too hokey, let's look at some inverse properties so the common queries can be supported.

elf-pavlik:
I don't think we will merge this one and all the inverse aliases related stuff we can track in #306

fosterlynn:
agreed

PR #420, opened 2019-01-25
Clarify event / reciprocity / claim / commitment, naming and directions.

fosterlynn:
... and also cardenalities.  An event would create one claim, a commitment can cover many claims.

Fixes the concept that the commitment should be created only by the agent committing.  The claim is implies by the event.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -137,10 +137,10 @@ vf:Fulfillment  a           owl:Class ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The quantity that the economic event fulfilled towards the commitment." .
 
-vf:Reciprocity  a           owl:Class ;
-        rdfs:label          "vf:Reciprocity" ;
+vf:Claim  a                 owl:Class ;
+        rdfs:label          "vf:Claim" ;
         vs:term_status      "unstable" ;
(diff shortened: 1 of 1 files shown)


fosterlynn:
>I think we might again stumble on non-dual cases where one event would result in multiple claims or it would require multiple events to result in one claim.

Yes, that was the original model.  But I thought about it, and am not sure we need that for the event side.  I'll go review what we did on that.

And yes, definitely needs examples, which I will work on.

One thing we could do for cleanup milestone is to just change the name to vf:Claim, and not do the rest.  Or just take it out of cleanup, not used by any user group we have going now.

elf-pavlik:
> One thing we could do for cleanup milestone is to just change the name to vf:Claim, and not do the rest. Or just take it out of cleanup, not used by any user group we have going now.

Maybe we take it out altogether for now and reintroduce as part of [*composition* milestone](https://github.com/valueflows/valueflows/milestone/4) ?
I think it goes together with recent discussions in #307 and #384 and it looks that to support non-dual scenarios we may need to use nested agreements.

fosterlynn:
Took it out of cleanup.  Maybe after we do this release, we can re-look at the other milestones and slot open issues and PRs into them.

fosterlynn:
Closing in favor of https://github.com/valueflows/valueflows/pull/489.

PR #421, opened 2019-01-25
New relationships between exchange agreement and commitments/events.

fosterlynn:
Other direction than vf:under.  Exchange contains commitments and/or events.  Naming needs some work.  Also still thinking about separating reciprocal ones, but we can look at that later when we do proposals if there is no immediate agreement.

### DIFF
--- a/examples/exch-agreement.yaml
+++ b/examples/exch-agreement.yaml
@@ -13,7 +13,7 @@
   # Exchange agreement with commitments
 
   - '@id': alice:57f1c1d0-432e-4bfa-9d32-002b8955a708
-    '@type': ExchangeAgreement
+    '@type': Agreement
     skos:note: Alice commits to giving Bob 50 kg of apples in exchange for 10 liters of apple cider.
 
   - '@id': alice:2342d456-5d6f-46d5-a7ed-3ac7bfd5a86c
@@ -23,7 +23,7 @@
     provider: https://alice.example/
     receiver: https://bob.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q89 # apples
-    affectedQuantity:
+    quantifiedAs:
       qudt:unit: unit: Kilogram
       qudt:numericValue: 50
 
@@ -34,25 +34,35 @@
     provider: https://bob.example/
     receiver: https://alice.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q5977438 # soft apple cider
-    affectedQuantity:
+    quantifiedAs:
       qudt:unit: unit: Liter
       qudt:numericValue: 10
 
   # Exchange without commitments
 
   - '@id': store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
-    '@type': ExchangeAgreement
+    '@type': Agreement
     skos:note: Carol purchased a new bucket at the hardware store and paid 5 dollars for it.
 
   - '@id': store:a8356625-bf64-4c16-9099-28aa1b718c4b
     '@type': EconomicEvent
-    clauseOf: store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
+    realizationOf: store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
     action: give
-    provider: https://alice.example/
-    receiver: https://bob.example/
-    resourceClassifiedAs: https://www.wikidata.org/wiki/Q89 # apples
-    affectedQuantity:
-      qudt:unit: unit: Kilogram
-      qudt:numericValue: 50
+    provider: https://store.example/
+    receiver: https://carol.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q47107 # bucket
+    quantifiedAs:
+      qudt:unit: unit: Number
+      qudt:numericValue: 1
 
-  
+  - '@id': store:a8356625-bf64-4c16-9099-28aa1b718c4b
+    '@type': EconomicEvent
+    realizationOf: store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
+    action: give
+    provider: https://carol.example/
+    receiver: https://store.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q4917 # US dollar
+    resourceInventoriedAs: carol:e56fd654-7b94-4d96-8e60-de39e08329a7 # Carol's bank account
+    quantifiedAs:
+      qudt:unit: unit: Number
+      qudt:numericValue: 5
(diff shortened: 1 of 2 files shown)


elf-pavlik:
Could you please include YAML example(s) straight away?

fosterlynn:
Good examples for a beginner in VF can't be done without transfers.  So I guess we need to wait on this one.  I committed suggested terms in the ttl for discussion, and the start of examples for later.

examples/exch-agreement.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  # Exchange without commitments
+
+  - '@id': store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
+    '@type': ExchangeAgreement
+    skos:note: Carol purchased a new bucket at the hardware store and paid 5 dollars for it.
+
+  - '@id': store:a8356625-bf64-4c16-9099-28aa1b718c4b
+    '@type': EconomicEvent
+    clauseOf: store:ac9ec98d-db80-44dc-97be-7aa149b2fe5d
```
elf-pavlik: I guess you wanted to use `realizationOf` here?

fosterlynn: yeah, just didn't get that far, left the examples for later

fosterlynn:
>Can we make both ranges vf:Agreement ?

I think of ExchangeAgreement as this, not other kinds of agreements.  Agreement covers a lot of territory.  (We called it Exchange in NRP, and I actually think that may be better.  Maybe Agreement shouldn't even be part of the vocab, just create property references like we do for classifiedAs?

elf-pavlik:
I don't think only exchange agreements can have Commitments as its clauses and Events as realizations of those clauses.

elf-pavlik:
Someone got caught on photo-radar speeding and they lost their driving license, later entered agreement to do some specific public service work and after that get the driving license back.
I wouldn't call it exactly an 'exchange agreement' and I don't see why use of those properties should imply ExchangeAgreement.

elf-pavlik:
Or simpler, someone just agreed to *load* something in one place and *unload* in another place. Both commitments stay 'clauses' of the same agreement but we don't have exchange in this case.

fosterlynn:
>Or simpler, someone just agreed to load something in one place and unload in another place. Both commitments stay 'clauses' of the same agreement but we don't have exchange in this case.

That is part of a process.  I don't see any larger agreement than that where this is a clauseOf, just commitments.

>Someone got caught on photo-radar speeding and they lost their driving license, later entered agreement to do some specific public service work and after that get the driving license back.  I wouldn't call it exactly an 'exchange agreement'

They agreed to do public service in exchange for getting their license back.  Sounds like an exchange agreement to me.

elf-pavlik:
> That is part of a process. I don't see any larger agreement than that where this is a clauseOf, just commitments.

Flows don't reference either Process or Agreement, any flow can reference both Process(es) and Agreement(s). Mentioned *load* and *unload* can happen in different processes, but still can stay as clauses of one Agreement.

Could you explain why do you think *clauseOf* should imply exchange agreement?

fosterlynn:
>Flows don't reference either Process or Agreement, any flow can reference both Process(es) and Agreement(s). Mentioned load and unload can happen in different processes, but still can stay as clauses of one Agreement

Flows can reference both, we have made that clear in the vocab.  But still I don't think commitments to load and unload rise to the level of an Agreement - sounds more like a plan.

The more I think about it, the more I think Agreement can just go away from VF.  We can keep vf:under and it can reference owl:Thing or whatever we can find that is more appropriate.  Maybe in that case vf:underAgreement is better.  Agreements that are not exchange agreements aren't really core imo, and are way less important to the model than Agent for example.  And people have done lots of work on smart contracts, etc., and I don't even know what vocabularies exist for agreement.  I propose we put it out of scope except for the property references.

DistributionAgreement can go away too, I think if someone wants to do that, it will be an extension to VF.

elf-pavlik:
>  I think commitments can stand on their own.

Commitments do stand on their own as well as 'clauses' of Agreements. 

> They might become part of an agreement if there is some reciprocity expected.

I think people can decide themselves why they would want to have multiple Commitments as 'clauses' of the same Agreement. I don't see any reason why vocab should imply an ExchangeAgreement when one uses Commitment as 'clause' in an Agreement.

> If we are stuck for the moment on this one, I'll take the cleanup milestone off of it, and we can address later.

We might need to create a dedicated *agreements* milestone and maybe work on it between *schedules* and *composition* ?

fosterlynn:
>We might need to create a dedicated agreements milestone and maybe work on it between schedules and composition ?

Sounds good.

Do you think we should do anything for cleanup on agreement?  Or just leave as unstable?

elf-pavlik:
I think we can just create *agreements* milestone and schedule it right after *schedules* milestone at the same time pushing all the other milestones accordingly.

release-doc-in-process/all_vf.TTL
```diff
@@ -541,7 +529,19 @@ vf:resourceConformsTo
...
+        rdfs:range          vf:Agreement ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "This commitment is part of the agreement, which could not exist without it." .
 
+vf:realizationOf a           owl:ObjectProperty ;
+        rdfs:label          "realization of" ;
+        rdfs:domain         vf:EconomicEvent ;
+        rdfs:range          vf:Agreement ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "This economic event occurs as part of this exchange." .
```
elf-pavlik: s/exchange/agreement/

fosterlynn:
@elf-pavlik we need something in place for this, so I expedited it, I think I changed everything to resolve your concerns.  Also got rid of ExchangeAgreement and DistributionAgreement, doesn't seem like there is any reason left to define them.  We can revisit some time later as needed.

I tried to rebase it, since it is pretty far behind, and it seemed to work locally, but wouldn't push.  This is my first try at rebasing, and I'm not sure how to fix that atm.

But if you can take a look anyhow and see if this seems mergeable in the meantime, that would be helpful.  If so, you could probably merge it, I don't really expect conflicts since I didn't get any in the rebase.

fosterlynn:
OK, those are confusing errors on the checks, especially Mixed content.  I think to save some pain, I'll just start this one over from scratch, it isn't that big.

fosterlynn:
Closing this one, too many technical issues.  Here is the new one: https://github.com/valueflows/valueflows/pull/500.  (Wow, 500, is that a milestone?? :smile: )

elf-pavlik:
after rebasing you can force push the branch

PR #426, opened 2019-01-26
minimal replacement of lod-converter

elf-pavlik:
work in progress, still checking how it works on TravisCI


elf-pavlik:
it should work now, ready to merge!

PR #431, opened 2019-01-28
added intended qty

fosterlynn:
define better later (we have been looking at several different quantities in Intents)

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -292,8 +292,8 @@ vf:currentQuantity a        owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The current amount and unit of the economic resource. This can be stored or derived from economic events affecting the resource" .
 
-vf:observedQuantity a       owl:ObjectProperty ;
-        rdfs:label          "observed quantity" ;
+vf:affectedQuantity a       owl:ObjectProperty ;
+        rdfs:label          "affected quantity" ;
         rdfs:domain         vf:EconomicEvent ;
         rdfs:range          qudt:QuantityValue ;
         vs:term_status      "unstable" ;
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -292,19 +292,26 @@ vf:currentQuantity a        owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The current amount and unit of the economic resource. This can be stored or derived from economic events affecting the resource" .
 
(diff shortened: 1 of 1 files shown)


elf-pavlik:
Since we have `affects` and `affectedQuantity` for `EconomicEvent`, and we have `involves` for `Commitment` and `Intent` we could have matching `involvedQuantity` for both `Commitment` and `Intent`

fosterlynn:
How about this?  

observedQuantity: this applies to all actions, and is the quantity that will be accounted for (then when we do time-based resources, we can add whatever is appropriate, and differentiate in whatever different ways we need to - but the observedQuantity will be always counted when things get summarized - whether there is an inventoried resource or not - makes more sense than affectedQuantity, which implies only that which affects resources

committedQuantity: also makes more sense than involvedQuantity, which isn't even a good descriptive word in the first place - what we need to know is what the person committed to (in NRP for example when a person commits to some work, they can change the originally intended time estimate to what they estimate - also applies to all actions

IntendedQuantity: just following the pattern, again I don't think involvedQuantity is accurate, since nothing is involved at the moment - and we can define later as needed, but we know we need some quantity as part of the intention

I think this would be a great piece of cleanup, and doesn't preclude other quantities and durations also used when we get to those use cases.

elf-pavlik:
Please go ahead with your proposal as long as all those properties stay *unstable*. Which means we leave it open to go back to `vf:affectedQuantity` and add just `vf:involvedQuantity`.
Actually we might even go with `affectedClassifiedAs` and `involvedClassifiedAs` to keep it all consistent with what describes the flow and what describes *affected* or *involved* resource.
We should also clarify if we need different properties or can just have same one for any flow. I can't recall why we currently have distinct ones for Intent, Commitment and Flow.
 
> and is the quantity that will be accounted for (then when we do time-based resources, we can add whatever is appropriate, and differentiate in whatever different ways we need to - but the observedQuantity will be always counted when things get summarized

I don't think we can decide anything like that at this moment. Luckily we have *schedules* milestone coming up next which should give us chance to clarify it all further. Once again, we may reverse this change based on this upcoming work.

fosterlynn:
>We should also clarify if we need different properties or can just have same one for any flow. I can't recall why we currently have distinct ones for Intent, Commitment and Flow.

That is a good question.  I think we ended up with involves because we started with affected, and affected didn't work as well for commitments.  I could be fine with just one.

fosterlynn:
OK, went with proposal, see what you think.

elf-pavlik:
I think we still miss property to use with Commitment

fosterlynn:
>I think we still miss property to use with Commitment

right above intendedQuantity (committedQuantity)

PR #439, opened 2019-01-29
changed a few statuses to 'testing'

fosterlynn:


elf-pavlik:
Let's don't promote any other terms till we add more extensive examples

fosterlynn:
>Let's don't promote any other terms till we add more extensive examples

Nope, all done for now.

PR #440, opened 2019-02-01
start json-schemas, minor fixes to ttl file

fosterlynn:
Do not merge yet, in process!


docs/specification/json-schemas.md
```diff
@@ -0,0 +1,31 @@
+# Json-schema definitions
+
+These schemas are offered for those who want to use ValueFlows but are not using Linked Open Data / Semantic Web technology.  See https://json-schema.org/ for more information.
```
elf-pavlik: I think we need to include here something like:
**VF doesn't require any specific serialization, publishers can publish valid VF data using any of [RDF1.1 Serializations](https://www.w3.org/TR/rdf11-new/#section-serializations) including [JSON-LD](https://json-ld.org/). Even valid VF data published in JSON-LD format doesn't need to follow those schemas. Applications assuming use of those schemas will only interoperate with the subset of VF ecosystem which uses JSON and conforms to schemas below.**

fosterlynn: Sure I'll work something like that in somewhere, maybe a more general page?

fosterlynn:
>I haven't looked at any of the schemas and if you would like them reviewed please request someone else to review them.

Sure.  Need to test/validate them too.  And have several questions I need to resolve.  Just pushed so I could debug the gitbook links, all in process, sorry if it was confusing....

docs/specification/json-schemas.md
elf-pavlik: I think we need both. A more general page elaborating in depth on RDF, JSON-LD, JSON, JSON Schemas, Media Types etc.
Each specification page might have a shorter note like the one above which links to that more general in-depth page.

fosterlynn: :+1:

elf-pavlik: I'll try to find a little bit of time today to make separate PR with dedicated page. This PR would just need to include a paragraph in lines of what I proposed above and link to that dedicated page.

elf-pavlik:
Could you please extract changes to `all_vf.TTL` into separate PR?

Issue #427, opened 2019-01-26
RDF spec plus non-RDF spec


fosterlynn:
Right now we are RDF centric (which is fine) in the actual spec, while noting that we will support any format needed.  The all_VF.ttl file is our "system of record" and the only machine-readable spec we have.  We are also discussing it not being good practice to include other vocabulary terms inside our vocabulary (skos:note, but also foaf:Agent, etc.).  See https://github.com/valueflows/valueflows/issues/423 and https://github.com/valueflows/valueflows/pull/424 and https://github.com/valueflows/valueflows/pull/415.

But, also we want to support non-rdf implementations (and are doing a couple of them now).  Even many of the ActivityPub developers want to use just plain json.  

Proposal: 
* We keep our machine-readable ttl spec with its generated human-readable spec as it is (and we can eliminate all other vocab elements).
* We add a parallel spec that is also machine-readable, and has the same content but includes the elements we want to have that are already represented in other rdf vocabularies.  One idea: json-schemas.  I think that is what we have in holochain (can check); and that is what ssb uses.

elf-pavlik:
For non-RDF applications, I would also like to know strategy for using VF together with other non-RDF vocabs and having a way to know which terms come from which vocab, as well as not having collisions int terms defined by different vocabs.

Currently non-RDF spec sounds to me more like "let's make a VF-centric bubble" rather than "let's contribute to an open ecosystem". Possibly because I don't see how those non-RDF approaches address challenges I mention above.

fosterlynn:
>When it comes to JSON Schema, AFAIK it can't properly represent RDF and similar solutions exist which can

But we have rdf definitions already, so we don't need to do that.  We need something that people can easily use that doesn't require rdf conventions.  I'm not suggesting we get rid of what we have of course.

>For non-RDF applications, I would also like to know strategy for using VF together with other non-RDF vocabs

Sort of interesting, I think the fact that rdf has properties that don't have to be part of a class makes a big difference here.  Like in non-rdf land (at least afaik), you can't have something like a free-floating skos:note, so you just put a note wherever you need one, and that is that, you aren't conflicting with anything else.  So the only overlap will be classes.

And I don't know how easy it is to mix and match vocabs that aren't rdf-based, that is something that the semantic web people have done relatively well imo.  But anyhow, I don't have experience with other kinds of standards.  @bhaugen was involved in some ISO and UNCEFACT standards for REA, he may have some knowledge there.

[...]

PR #444, opened 2019-02-04
overview of specifications

elf-pavlik:
Related to #440 and #427 

I have obvious pro Linked Data bias. Please help me with getting this overview little more balanced.

### DIFF
--- a/docs/specification/overview.md
+++ b/docs/specification/overview.md
@@ -1,0 +1,37 @@
+# Specifications
+
+ValueFlows aims at providing model which can get used in different information ecosystems.
+
+## The Web of Linked Data
+
+We provide web ontology which uses [RDF Schema 1.1](https://www.w3.org/TR/rdf-schema/) and [OWL 2](https://www.w3.org/TR/owl2-overview/) constructs. It builds on top of well established standards:
+
+* [IRIs](https://tools.ietf.org/html/rfc3987) as global identifiers
+* [HTTP/2](https://tools.ietf.org/html/rfc7540) as universal dereferencing protocol
+* [RDF 1.1](https://www.w3.org/TR/rdf11-concepts/) as well defined (web native) data graph model
+* [SHACL](https://www.w3.org/TR/shacl/) and [ShEx](https://shex.io/) as constraints and validation languages
+* [Web Architecture](https://www.w3.org/standards/webarch/) as omnipresent platform
+
+This leads to certain advantages:
+
+* namespacing helps prevent naming collisions
+* abstract data model which can adopt new serializations, eg. [HDT](http://www.rdfhdt.org/)
+* existing ecosystem of open vocabularies https://lov.linkeddata.es/
+* active community and ongoing academic research [ISWC](https://en.wikipedia.org/wiki/International_Semantic_Web_Conference), [ESWC](https://en.wikipedia.org/wiki/Extended_Semantic_Web_Conference), [SEMANTiCS](https://semantics.cc/), ...
+
+RDF 1.1 has [multiple serialization formats](https://www.w3.org/TR/rdf11-new/#section-serializations), we recommend publishing Turtle/TriG, N-Triples/N-Quads and JSON-LD (expanded and flattened), at the same time we discourage use of RDFa and RDF/XML. If one needs to include one of recommended above RDF in HTML source, [embedding in script tag](https://w3c.github.io/json-ld-syntax/#embedding-json-ld-in-html-documents) also stays available.
+
+## JSON & JSON Schema
+
+Many emerging environments tie themselves more closely to very popular in last years JSON based technologies. For example [Scuttlebutt](https://www.scuttlebutt.nz/) or [Holochain](https://holochain.org/)
+
+* [JSON](https://www.json.org/) as common data serialization
+* [JSON Schema](https://json-schema.org/) as vocabulary to annotate and validate JSON
+
+While it comes with certain simplicity it also leads to certain disadvantages:
+
+* lack of specific underlying graph model
+* no namespacing which can result in naming collisions or proliferation of media types
+* next generation serialization might replace JSON just like it replaced XML
+
+Keeping above in mind this documentation still provides reference JSON Schemas.
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -30,6 +30,7 @@
 
 ## Specification
 
+* [Overview](specification/overview.md)
 * [External terms](specification/external-terms.md)
 * [Classes and Properties](http://www.essepuntato.it/lode/owlapi/https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/all_vf.TTL)
 


elf-pavlik:
I think we should add both some reasonable disadvantages of Linked Data and some reasonable advantages of just JSON.

bhaugen:
> I think we should add both some reasonable disadvantages of Linked Data and some reasonable advantages of just JSON.

I thought "whole vocab in one place" was a reasonable advantage of json. I'm not sure how to state that as a disadvantage of LD in a general way, except for the example of Agent which is no longer in the LD spec, but is in the json schema spac.

The privacy problems are more about the assumptions of Linked "Open" Data, where privacy advocates want more control over who can access that data, and the LOD solution of ACLs is not reassuring to them. I don't yet know what is reasonable there.

I don't know if the programmer dislike for json-LD is reasonable or not because I have not tried to use json-LD as an input, only as output, where I got a library that handles it fairly well. I think the problems people are complaining about are with interpreting json-LD inputs, and I don't know if they are reasonable or not. It might just be familiarity, or it might be some substantive problems.

I'll ask a json-LD hater I have some contact with...

fosterlynn:
>I think we should add both some reasonable disadvantages of Linked Data and some reasonable advantages of just JSON.

I'm not sure it is a high priority or that we want to get into that role too strongly.  I think we need to support what people want to use for communications that need a standard vocabulary - and they will have way stronger opinions than us and will have done way more research for their situation.  So I think that a machine readable spec in whatever people want is a good thing.  We can explain what they get and don't get out of the vocab in different situations -- maybe that is more what you meant anyhow....?

bhaugen:
The only one I think would be worth adding is that the json-schema spec gives you the whole vocab in one place. @fosterlynn says she doesn't know yet if it will be one page, so maybe wait to add the statement.

elf-pavlik:
> The only one I think would be worth adding is that the json-schema spec gives you the whole vocab in one place.

At the same time not providing a safe way to complement it with terms not in that 'one place' and which will never find their way to that 'one place'.  I think we should explain that such convenience does come as trade-off. Taking it might result in having pains of revert that decision further down the road once you application starts integrating with open heterogeneous ecosystem.

bhaugen:
> At the same time not providing a safe way to complement it with terms not in that 'one place' and which will never find their way to that 'one place'. I think we should explain that such convenience does come as trade-off. Taking it might result in having pains of reverting that decision further down the road once ones application starts integrating with open heterogeneous ecosystem.
> 

I'm good with explaining the tradeoffs, as long as the explanation uses the example of Agent. That would explain clearly to anybody who knows REA.

I don't know what the extensibillity story is with json-schema, but we might run into it in the Holo-REA project...

elf-pavlik:
I would find it very helpful to understand what **graph model** people using plain JSON do rely on. I also wonder if Holochain and SBB use the same graph model.

elf-pavlik:
This issue also seems in a way a parent issue of #306

docs/specification/overview.md
```diff
@@ -0,0 +1,37 @@
...
+* [IRIs](https://tools.ietf.org/html/rfc3987) as global identifiers
+* [HTTP/2](https://tools.ietf.org/html/rfc7540) as universal dereferencing protocol
+* [RDF 1.1](https://www.w3.org/TR/rdf11-concepts/) as well defined (web native) data graph model
+* [SHACL](https://www.w3.org/TR/shacl/) and [ShEx](https://shex.io/) as constraints and validation languages
+* [Web Architecture](https://www.w3.org/standards/webarch/) as omnipresent platform
+
+This leads to certain advantages:
+
+* namespacing helps prevent naming collisions
+* abstract data model which can adopt new serializations, eg. [HDT](http://www.rdfhdt.org/)
```
elf-pavlik: TODO: mention CBOR https://forum.solidproject.org/t/hdt-for-dynamic-and-small-graphs/163/4

fosterlynn:
>This issue also seems in a way a parent issue of #306

Yeah, agree.  I've given up on inverse properties (like schema.org defines) because you seem to feel strongly about it @elf-pavlik .  Will just define them in the graphql treatment and in the json-schemas if they fit into that, don't know yet.

>I would find it very helpful to understand what graph model people using plain JSON do rely on. I also wonder if Holochain and SBB use the same graph model.

I'm not sure I'm understanding your question very well.  So forgive me if I'm not answering the right question. 

So far, people are either taking my word for it because I defined the VF api they are using so they have no choice.  Or are reading VF doc and figuring it out from there.  The effort to create a graphql reference will work for this current group of projects since they are all using graphql as an interface to the UI anyway.

SSB uses json-schemas for its message formats on the backend side.  We are doing that in HoloREA too.  Thus if we can get those fleshed out, again multiple projects can just copy them in, and they will all agree at least on the naming and data structure.

I get that it can be a disadvantage to not have machine-readable context to keep people and software straight.  But people do it anyhow, and just rely on their reading of specs and documentation, and then testing.  So from my perspective of ignorance and not wanting to get into the current religious wars about json-ld, and also wanting devs to be able to choose what they want, I really do want something that can be used in all the different tendencies and experiments.

elf-pavlik:
Another relevant resource comparing various aspects of JSON + JWT vs JSON-LD + JWT vs JSON-LD + LD Proofs
https://w3c.github.io/vc-imp-guide/#benefits-of-json-ld-and-ld-proofs

fosterlynn:
>Another relevant resource comparing various aspects of JSON + JWT vs JSON-LD + JWT vs JSON-LD + LD Proofs

I expect this is important, even though I don't understand a lot of it.  It is probably useful to devs making decisions about what they want to use.  And I think @bhaugen may find it applicable to some work he is doing.  But I still don't think we want to get in the business of recommending any particular technology in VF.  People have "religious wars" on this stuff, and in 2 decades it will all be different anyhow. I think we need to go with what devs want to use now.  @bhaugen keeps reminding us that we need to support SMS and hashtags with the vocabulary.  And csv, Mikey wanted that early on for Holodex.  

I don't mind the spec being LOD at its root, there is a lot of vocabulary history there with w3c, and history even just with the group of people involved in the beginnings of VF as it grew out of open app ecosystem discussions etc.

But I think it is really important to have a spec that can be used however people now and in the future want to, for economic interoperability.  It will be messy I'm sure, but if people care, they will find ways to hook all these different efforts up.

And my sense is that right now our problem is getting wide adoption, not narrowing the options for people, even to what might actually be "better".

My 2c.

bhaugen:
>  my sense is that right now our problem is getting wide adoption, not narrowing the options for people, even to what might actually be "better".

How about "some adoption and actual use" and then bring some of the early adopters into the project to help refine the vocab from early experience?

It's good if some of them are RDF fans. Somebody from the SolidBase team would be an interesting addition if they can take time. But the early adopters will most likely not all be RDF fans.

elf-pavlik:
During chat with @fosterlynn https://gitter.im/valueflows?at=5c9fc64593fb4a7dc2a7eab2 we discussed case where people use any number of additional properties from any number of vocabs on instances they create. For example in LDP which Solid builds upon API can handle data with whatever additional properties client chooses to include. Possibly GraphQL might not have such flexibility and don't allow clients to include in data they submit properties not defined in that GraphQL endpoint. @pospi could you possibly comment on that?

pospi:
> Possibly GraphQL might not have such flexibility and don't allow clients to include in data they submit properties not defined in that GraphQL endpoint. @pospi could you possibly comment on that?

That's not a problem so far as I am aware. The tooling for comparing schemas is pretty open-ended, so it's up to us whether we consider "additional properties" as an error or not. I don't think it factors in to schema stitching, either- if you're targeting an intersection of common functionality between two APIs then your query won't touch missing parts of the related schemas but your query documents will still validate and everything will be fine.

elf-pavlik:
Thanks @pospi 

> The tooling for comparing schemas is pretty open-ended, so it's up to us whether we consider "additional properties" as an error or not.

That sounds useful, I don't think we would ever want to define in VF namespace things like `shoe size` but people can still use tons of things like that from all kinds of domain specific vocabularies.

fosterlynn:
This PR is on hold until we can hold more discussion to refine the explanations.

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

> https://lab.allmende.io/valueflows/valueflows/-/issues/444.

If you have not done so, you are very welcome to register at https://lab.allmende.io and join the ValueFlows organization there.

Issue #443, opened 2019-02-04
Untangling work, use, transfer in scheduling


fosterlynn:
I'm thinking that the scheduling milestone should be done without resolving transfer if possible, as that is already defined for later, and is another big "can of worms".  So part of this is to see if we can separate out the transfer aspects of scheduling for later.  Or if not, then so be it.

Also there are some differences in work and usage, in that work does not involve a material resource (we do not too much like the term "human resources" and would like to work towards human agency as people and democratic groups, therefore don't foresee people as EconomicResources).

Copying from #394 to here, and will delete from that issue:

Summarizing the quantities we may need to think about for usage:
1. Number used of a larger stock resource.
2. Actual time used (a QuantityValue)
3. Duration on a schedule (reserved, available, others can't use it)
4. Duration (larger) where the usage occurred - like (10 hours) used in the last month
5. Duration of the actual usage event, where the resource was used as input to a process. [edit]

Looking at similarities and differences with work:
1. No material resource, always one provider agent
2. Actual time worked (a QuantityValue)
3. Duration on a schedule (reserved, available, others can't schedule the person to work)
4. Duration (larger) where the work occurred - like (10 hours) worked in the last month
5. Duration of the actual work event, where the work was input to a process. [edit]

Work is the same except 1.

[edit] I think that 3. involves transfers.  Although it is also a chunk of time on a schedule, so could be represented as an interval or duration.

fosterlynn:
Further random thoughts:

I think we should reserve the action `use` for actual usage, which contributes to a process.  So availability or possession does not imply usage, as we can see in #220.  (Nor does ownership for that matter.)

Likewise for `work`.

Interestingly, `cite` has a similar aspect to `use` and `work` where the quantity that matters is not the same as the quantity of the cited resource (document, design, etc.).  And I can't think of a situation where a citation could be made for more than one of a "stock" resource.  The quantity that matters still could be '1' in that citations (the events) are counted that way.  For Sensorica's distribution logic, they wanted to have the person doing the citing record a percentage of the process/output that they felt the cited resource contributed to that process, then the contributor(s) to the cited resource would get that percentage of income assigned to that process .  It is different in that citations don't affect availability.  But we could consider `cite` in terms of figuring out what quantity properties are used for what.

[...]

Issue #442, opened 2019-02-04
Multiple aspects of usage (taxi)


fosterlynn:
Use case: Taxi service involves recording
1) miles
2) idling time when not moving

How to record this kind of usage of the taxi?  Or is there a better way of thinking about this use case than usage?

bhaugen:
Taxi fare calculations:

* Seattle:
> Taxi fare is a combination of three separate charges:
Drop (or base) charge.
Distance charge.
Time charge.
You don't simultaneously pay both distance and time charges; you pay one or the other.

Drop charge: Applied as soon as you enter the taxi and accept a ride | $2.60
-- | --
Distance charge: Applied once the taxi is travelling to your destination | $2.70 per mile($0.30 per 1/9 mile)
Time charge: Applied when the taxi stops or travels at slow speeds during the trip (for example, at red lights or in heavy stop-and-go traffic) | $0.50 per minute ($0.30 per 36 seconds)

* Wash DC:
> base charge of $3. Total distance traveled is charged at $2.16 per mile... The traffic wait time rate is $25 per hour, or about $2 every 5 minutes. Riders pay 50 cents for "each piece of luggage the operator places in his or her trunk."

* NYC:
> base fare of $2.50 plus 25 units at $0.50 each, or $12.50, for a total of $15.00. It is customary to tip the driver 20%.

elf-pavlik:
> Drop charge: Applied as soon as you enter the taxi and accept a ride

Sounds like agreement might include *load* event and that specific charge.

> Riders pay 50 cents for "each piece of luggage the operator places in his or her trunk."

Similar just *load* luggage not the person.

> Distance charge: Applied once the taxi is travelling to your destination 	$2.70 per mile($0.30 per 1/9 mile)
> Time charge: Applied when the taxi stops or travels at slow speeds during the trip (for example, at red lights or in heavy stop-and-go traffic)   $0.50 per minute ($0.30 per 36 seconds)
[...]
> base charge of $3. Total distance traveled is charged at $2.16 per mile... The traffic wait time rate is $25 per hour, or about $2 every 5 minutes.

Do you know how does this get calculated? Would having total distance and total duration suffice or one has to have continuous record of speed throughout the whole duration of the ride?

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

[...]

Issue #394, opened 2019-01-11
Example: use of drills (shared tools)


fosterlynn:
From the UC https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#shared-tools

>Local community maintains a shed where they keep shared tools. They maintain each category of tools as one resource with certain number of items in it. People can use those tools, for example:
>    Use 5 electric drills (out of 23 in total) for 3 hours

(Doing this as an issue, as it is a bit obscure for a published example and may have many opinions.)

```yaml
'@context':
  '@vocab': https://w3id.org/valueflows#
  xsd: http://www.w3.org/2001/XMLSchema#
  qudt: http://qudt.org/schema/qudt/
  unit: http://qudt.org/vocab/unit/
  time: http://www.w3.org/2006/time#
  time:inXSDDateTimeStamp:
    '@type': xsd:dateTimeStamp
  alice: https://alce.example/
  bob: https://bob.example/
  carpenters: https://local-carpenters.example/
  community: https://community-tools.example/

'@graph':

  - '@id': community:60f4204e-b8d2-4026-8577-102c3f82c0af
    '@type': EconomicResource
    skos:note: Drills
    currentQuantity:
      qudt:unit: unit:Unitless
      qudt:numericValue: 23

  - '@id': community:6405b8ad-0ac2-4d58-abdb-08089342abcd
    '@type': Transfer
    skos:note: carpenters borrow drills from community tool lending library

  - '@id': community:6405b8ad-0ac2-4d58-abdb-0808903c78ad
    '@type': EconomicEvent
    inputOf: community:6405b8ad-0ac2-4d58-abdb-08089342abcd
    action: give
    provider: community
    receiver: carpenters
    affects:community:60f4204e-b8d2-4026-8577-102c3f82c0af
    observedQuantity:
      qudt:unit: unit:Unitless
      qudt:numericValue: 5
    observedTime:
      time:intervalStarts:
        time:inXSDDateTimeStamp: 2019-01-02T02:10:43.003-5:00
      time:intervalEnds:
        time:inXSDDateTimeStamp: 2019-01-02T02:10:43.003-5:00

  - '@id': carpenters:6453b8ad-0ac2-4d58-abdb-0808903c78ad
    '@type': EconomicEvent
    outputOf: community:6405b8ad-0ac2-4d58-abdb-08089342abcd
    action: receive
    provider: community
    receiver: carpenters
[...]

Issue #231, opened 2017-06-23
EconomicEvent.affects for work


fosterlynn:
EconomicEvent now references its affected resource, which makes sense for inventoried resources.  We have also talked about if a material (or digital) resource is not inventoried, and we only have the category or model of the resource, that would still be a vf:Resource, but it would only have the category or model reference, not other info that exists on an inventoried resource.  (If I remember and understand correctly.)

I am trying to create VF compliant api, and ran into work events.  It is not very clear that an economic event "vf:affects" a category of work.  This is the first time I have seen an actual need to subclass or differentiate EconomicResource in any way.  I'm sure there are other ways too.

Ideas how to best handle this for VF?

For now, I created another property called workCategory.....

elf-pavlik:
Since *work* event directly represents the work, I think you can simply use `vf:classifiedAs` to reference classification. 
Actually this doesn't directly relate to *schedules*, I added it to this milestone to quickly.

fosterlynn:
I think this is too old, we have moved way beyond this, closing.  Please re-open if you think we need it.

PR #446, opened 2019-02-05
Example time-based quantities and times

fosterlynn:
This is just the beginning of this PR, exploring examples for the scheduling milestone.  Focusing on getting the quantity and time properties defined for different use cases, irrespective of transfers or composition.

Started with work, as possibly the simplest, want to make usage be as similar as possible.

Things to note:
* calendar property
* observedQuantity name change
* all the time properties can be = or > than the quantity properties (the quantity properties are what is counted, the time properties are what relates to scheduling)

### DIFF
--- a/examples/use-partial-resource.yaml
+++ b/examples/use-partial-resource.yaml
@@ -17,6 +17,7 @@
     currentQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 10
+    unitOfUse: unit:Hour
 
   - '@id': carpenters:8791913c-0152-4c32-bb3f-f25d61df736b
     '@type': EconomicResource
@@ -25,6 +26,7 @@
     currentQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
+    unitOfUse: unit:Hour
 
   # the carpenters are working on a process
 
--- a/examples/use-partial-resource.yaml
+++ b/examples/use-partial-resource.yaml
@@ -42,7 +42,7 @@
     observedQuantity: # hours of use of the 3 drills
       qudt:unit: unit:Hour
       qudt:numericValue: 3.5
-    affectedQuantity: # portion of the stock resource that is used
+    usedQuantity: # portion of the stock resource that is used
       qudt:unit: unit:Number
       qudt:numericValue: 3
     observedTime: # the 3 drills are unavailable during this time
--- a/examples/use-partial-resource.yaml
+++ b/examples/use-partial-resource.yaml
@@ -1,0 +1,74 @@
+# Example: Use a portion of a stock resource
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - alice: https://alice.example/
+    carpenters: https://carpenters.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/use-partial-resource.yaml
+'@graph':
+
+
+  # tools used by the group of carpenters
+
+  - '@id': carpenters:9f112621-3f13-4f48-a3e2-fd2ca99493ca
+    '@type': EconomicResource
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q58964 # stock drills
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 10
+
+  - '@id': carpenters:8791913c-0152-4c32-bb3f-f25d61df736b
+    '@type': EconomicResource
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q2365675 # a table saw
+    trackingIdentifier: 38922-44
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  # the carpenters are working on a process
+
+  - '@id': carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125
+    '@type': Process
+    name: Make a small shed
+
+  - '@id': carpenters:e72f7d1d-4f0a-4518-a44c-96658cafda94
+    '@type': EconomicEvent
+    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
+    action: use
+    provider: https://carpenters.example/
+    receiver: https://carpenters.example/
+    affects: carpenters:9f112621-3f13-4f48-a3e2-fd2ca99493ca # drills
+    observedQuantity: # hours of use of the 3 drills
+      qudt:unit: unit:Hour
+      qudt:numericValue: 3.5
+    usedQuantity: # portion of the stock resource that is used
+      qudt:unit: unit:Number
+      qudt:numericValue: 3
+    observedTime: # the 3 drills are unavailable during this time
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-20T08:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-20T11:30:00-5:00
+
+  - '@id': carpenters:f325a950-5737-488d-b122-8f21114d0eb9
+    '@type': EconomicEvent
+    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
+    action: use
+    provider: https://carpenters.example/
+    receiver: https://carpenters.example/
+    affects: carpenters:carpenters:8791913c-0152-4c32-bb3f-f25d61df736b # table saw
+    observedQuantity: # hours of use of the table saw (no need for usedQuantity)
+      qudt:unit: unit:Hour
+      qudt:numericValue: 2
+    observedTime: # the table saw is unavailable during this time
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-20T08:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-20T11:30:00-5:00
--- a/examples/use-partial-resource.yaml
+++ b/examples/use-partial-resource.yaml
@@ -42,7 +42,7 @@
     observedQuantity: # hours of use of the 3 drills
       qudt:unit: unit:Hour
       qudt:numericValue: 3.5
-    usedQuantity: # portion of the stock resource that is used
+    affectedQuantity: # portion of the stock resource that is used
       qudt:unit: unit:Number
       qudt:numericValue: 3
     observedTime: # the 3 drills are unavailable during this time
--- a/examples/work.yaml
+++ b/examples/work.yaml
@@ -1,0 +1,134 @@
+# Example: Scheduling - work as a time-based resource
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - alice: https://alice.example/
+    mfg: https://manufacturing.example/
+    fablab: https://fablab.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/work.yaml
+'@graph':
+
+
+  # mfg requests someone to do 8 hours of 3D printing sometime on Oct 14 and/or 15th
+
+  - '@id': mfg:8e5fe80d-a769-4bd5-89e5-2136d33eab9s
+    '@type': Intent
+    action: work
+    receiver: https://manufacturing.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q229367 # 3D printing (as a type of work or skill)
+    intendedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+    intendedTime:
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-15T18:00:00-5:00
+
+  # alice offers that she can provide 3D printing work per her calendar
+
+  - '@id': alice:9bd19194-a36d-4a1f-896b-8082887962ca
+    '@type': Intent
+    action: work
+    provider: https://alice.example/
+    resourceClassifiedAs: mfg:https://www.wikidata.org/wiki/Q229367 # 3D printing (as a type of work or skill)
+    calendar: https://alice.example/calendar-3D/
+
+  # alice finds the mfg request while she is searching for matches to her general offer
+  # she commits to do 3D printing work for the mfg project on Oct 14 sometime between 8h and 18h
+  # alice's calendar then will show her as unavailable during that time
+
+  - '@id': alice:33e8933b-ff73-4a01-964a-ca7a98893082
+    '@type': Commitment
+    action: work
+    provider: https://alice.example/
+    receiver: https://manufacturing.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q229367 # 3D printing (as a type of work or skill)
+    committedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+    committedTime:
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-14T18:00:00-5:00
+
+  - '@id': mfg:6078f30a-b15b-4b1c-8e77-797abae11377
+    '@type': Satisfaction
+    satisfies: mfg:8e5fe80d-a769-4bd5-89e5-2136d33eab9s # mfg's intent (a request)
+    satisfiedBy: alice:33e8933b-ff73-4a01-964a-ca7a98893082 # the commitment
+    satisfiedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+
+  - '@id': mfg:583e83d9-a46d-44ff-bd71-88513a1d83c9
+    '@type': Satisfaction
+    satisfies: alice:9bd19194-a36d-4a1f-896b-8082887962ca # alice's intent (an offer)
+    satisfiedBy: alice:33e8933b-ff73-4a01-964a-ca7a98893082 # the commitment
+    satisfiedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+
+  # alice does the work on Oct 14 from 8h to 16h, completing the sequence
+
+  - '@id': mfg:60f4204e-b8d2-4026-8577-102c3f82c0ad
+    '@type': EconomicEvent
+    action: work
+    provider: https://alice.example/
+    receiver: https://manufacturing.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q229367 # 3D printing (as a type of work or skill)
+    observedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+    observedTime:
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-10-14T16:00:00-5:00
+
+  - '@id': mfg:e56fd654-7b94-4d96-8e60-de39e08329a6
+    '@type': Fulfillment
+    fulfills: alice:33e8933b-ff73-4a01-964a-ca7a98893082 # the commitment
+    fulfilledBy: mfg:60f4204e-b8d2-4026-8577-102c3f82c0ad # the economic event
+    fulfilledQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+
+  # later on fablab finds alice's calendar and commits to make use of her skills during an available slot
+  # alice's calendar then shows her as unavailable during that slot
+
+  - '@id': fablab:86ad6e22-7640-4b93-972a-db7177ab2615
+    '@type': Commitment
+    action: work
+    provider: https://alice.example/
+    receiver: https://fablab.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q229367 # 3D printing (as a type of work or skill)
+    committedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 6
+    committedTime:
+      '@type': time:ProperInterval
+      time:intervalStarts:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-12-07T9:00:00-5:00
+      time:intervalEnds:
+        '@type': time:Instant
+        time:inXSDDateTimeStamp: 2018-12-07T15:00:00-5:00
+
+  - '@id': mfg:d4d2fd71-34f2-41c3-b1c5-19ad5ed2da59
+    '@type': Satisfaction
+    satisfies: alice:9bd19194-a36d-4a1f-896b-8082887962ca # alice's intent (an offer)
+    satisfiedBy: fablab:86ad6e22-7640-4b93-972a-db7177ab26158 # the commitment
+    satisfiedQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 6
--- a/examples/work.yaml
+++ b/examples/work.yaml
@@ -105,10 +105,10 @@
       qudt:unit: unit:Hour
       qudt:numericValue: 8
 
-  # later on fablab finds alice's calendar and commits to make use of her skills during an available slot
-  # alice's calendar then shows her as unavailable during that slot
+  # later on fablab finds alice's calendar and requests to make use of her skills during an available slot
+  # alice commits, and alice's calendar then shows her as unavailable during that slot
 
-  - '@id': fablab:86ad6e22-7640-4b93-972a-db7177ab2615
+  - '@id': alice:86ad6e22-7640-4b93-972a-db7177ab2615
     '@type': Commitment
     action: work
     provider: https://alice.example/
@@ -126,10 +126,10 @@
         '@type': time:Instant
         time:inXSDDateTimeStamp: 2018-12-07T15:00:00-5:00
 
-  - '@id': mfg:d4d2fd71-34f2-41c3-b1c5-19ad5ed2da59
+  - '@id': alice:d4d2fd71-34f2-41c3-b1c5-19ad5ed2da59
     '@type': Satisfaction
     satisfies: alice:9bd19194-a36d-4a1f-896b-8082887962ca # alice's intent (an offer)
-    satisfiedBy: fablab:86ad6e22-7640-4b93-972a-db7177ab26158 # the commitment
+    satisfiedBy: alice:86ad6e22-7640-4b93-972a-db7177ab26158 # the commitment
     satisfiedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 6
--- a/examples/work.yaml
+++ b/examples/work.yaml
@@ -147,10 +147,18 @@
       qudt:unit: unit:Hour
       qudt:numericValue: 21.5
     observedTime:
+      '@type': time:Instant
+      time:inXSDDateTimeStamp: 2018-10-31T00:00:00-5:00
+    accountDuring: org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
+
+  - @id': org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
+    '@type': AccountingPeriod
+    name: October 2018
+    accountingTime:
       '@type': time:ProperInterval
       time:intervalStarts:
         '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-01T00:00:00-5:00
+        time:inXSDDateTimeStamp: 2018-10-01T0:00:00-5:00
       time:intervalEnds:
         '@type': time:Instant
         time:inXSDDateTimeStamp: 2018-10-31T23:59:59-5:00
--- a/examples/work.yaml
+++ b/examples/work.yaml
@@ -5,6 +5,7 @@
   - alice: https://alice.example/
     mfg: https://manufacturing.example/
     fablab: https://fablab.example/
+    org: https://organization.example/
 
 '@id': rgh:valueflows/valueflows/master/examples/work.yaml
 '@graph':
@@ -24,7 +25,7 @@
       '@type': time:ProperInterval
       time:intervalStarts:
         '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+        time:inXSDDateTimeStamp: 2018-10-14T08:00:00-5:00
       time:intervalEnds:
         '@type': time:Instant
         time:inXSDDateTimeStamp: 2018-10-15T18:00:00-5:00
@@ -55,7 +56,7 @@
       '@type': time:ProperInterval
       time:intervalStarts:
         '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+        time:inXSDDateTimeStamp: 2018-10-14T08:00:00-5:00
       time:intervalEnds:
         '@type': time:Instant
         time:inXSDDateTimeStamp: 2018-10-14T18:00:00-5:00
@@ -91,7 +92,7 @@
       '@type': time:ProperInterval
       time:intervalStarts:
         '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-5:00
+        time:inXSDDateTimeStamp: 2018-10-14T08:00:00-5:00
       time:intervalEnds:
         '@type': time:Instant
         time:inXSDDateTimeStamp: 2018-10-14T16:00:00-5:00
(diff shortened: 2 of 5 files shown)


fosterlynn:
Modified property affectedQuantity to observedQuantity in the manufacturing example to make clear I'm proposing that work time-based quantities would be handled with the same property as volume-based or weight-based or count-based or whatever other kind of quantities we have in events.

fosterlynn:
Added a taxi example to explore 2 different ways of recording use for one resource in one process. Ref. https://github.com/valueflows/valueflows/issues/442.

Vocab note: Added an output action of `service`, for discussion, off topic of the PR, but there it was.

fosterlynn:
P.S. This is an exploratory PR, it can be taken apart later as needed if any of it becomes worth merging.  But it is helpful to me to keep all the examples bundled together for now.  Next up: drills!  (I hope.)

fosterlynn:
Added drills example.

Vocab note: new property `usedQuantity`.  Trying to keep the most common measured quantity in the same place (observedQuantity).

fosterlynn:
Added simple `use` example for more basic doc.
 
This round of suggested examples attempts to address https://github.com/valueflows/valueflows/issues/443, https://github.com/valueflows/valueflows/issues/231, https://github.com/valueflows/valueflows/issues/394, https://github.com/valueflows/valueflows/issues/174, https://github.com/valueflows/valueflows/issues/156, https://github.com/valueflows/valueflows/issues/442, https://github.com/valueflows/valueflows/issues/230 -- for the quantity/time fields while keeping transfer and composition out of the picture.

examples/use-partial-resource.yaml
```diff
@@ -0,0 +1,74 @@
...
+        time:inXSDDateTimeStamp: 2018-10-20T11:30:00-5:00
+
+  - '@id': carpenters:f325a950-5737-488d-b122-8f21114d0eb9
+    '@type': EconomicEvent
+    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
+    action: use
+    provider: https://carpenters.example/
+    receiver: https://carpenters.example/
+    affects: carpenters:carpenters:8791913c-0152-4c32-bb3f-f25d61df736b # table saw
+    observedQuantity: # hours of use of the table saw (no need for usedQuantity)
```
elf-pavlik: Keeping in mind [Open-world assumption](https://en.wikipedia.org/wiki/Open-world_assumption) one would need to consider the quantity `unknown`, since we need a way to provide that quantity, we should make it required and mention that one should not make assumptions based on lack of information.

examples/work.yaml
```diff
@@ -0,0 +1,156 @@
...
+    fulfills: alice:33e8933b-ff73-4a01-964a-ca7a98893082 # the commitment
+    fulfilledBy: mfg:60f4204e-b8d2-4026-8577-102c3f82c0ad # the economic event
+    fulfilledQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+
+  # later on fablab finds alice's calendar and commits to make use of her skills during an available slot
+  # alice's calendar then shows her as unavailable during that slot
+
+  - '@id': fablab:86ad6e22-7640-4b93-972a-db7177ab2615
```
elf-pavlik: I think always agent referenced as *provider* has to make the Commitment.

examples/use-partial-resource.yaml
fosterlynn: I think the quantity of the transferable resource (not the time-based use quantity) can be considered to be that of the resource being used.  Doesn't work?

examples/work.yaml
fosterlynn: Well interesting.  What if it is basically a plan?  Not a commitment for someone to give something?

elf-pavlik: one can use `vf:Intent` that have specified both `vf:provider` and `vf:receiver`

examples/use-partial-resource.yaml
elf-pavlik: How one can tell difference between 'missing some information' and 'just use the current quantity of the resource', it also would depend on having access to the information about current quantity of the resource at the time of event. I think event has to capture that information and provide it as part of asserted knowledge rather than leaving it ambiguous.

examples/work.yaml
fosterlynn: In this case, it is not an intent, it is agreed between both agents.  I think for usage where the receiver signed up for a slot on a schedule, they are committing to making use of that offer.  Might be based on offer or request.

examples/use-partial-resource.yaml
fosterlynn: But the observed quantity is the only one that is used for anything in terms of counting / accounting, I'm proposing.  The observed quantity is the one that people need to observe.  That's why I'm trying to get that information into one property.  Other quantities are optional.  So its fine for usedQuantity to be unknown imo.

examples/work.yaml
elf-pavlik: > What if it is basically a plan? Not a commitment for someone to give something?

`vf:Intent`

> In this case, it is not an intent, it is agreed between both agents.

`vf:Commitment` created by agent referenced as `vf:provider`

More *Conversation for Action* details we may need to handle soon, for now let's try to choose if the have already agreed and make *provider* create the Commitment, or they haven't agreed yet and then either one can create Intent.

fosterlynn: If an Intent (offer) is made with a schedule posted online that shows openings, and someone picks out a time duration from the schedule - doesn't that create a Commitment made by the receiver?  In fact there isn't a CfA really, no conversation is needed.

I don't think all plans are Intents.  Most plans include only Commitments because you don't need to seek out a matching agent.

>one can use `vf:Intent` that have specified both `vf:provider` and `vf:receiver`

Then how is it an Intent?  And how do you know if it is an offer or request?

fosterlynn:
>Just as we have `EconomicResource` for which `currentQuantity` should come as computed value based on flows.  We could consider something like `EconomicSchedule` which would in similar way stay computed based on flows. At the same time it would make it simpler to only publish minimum information needed to know availability.

We could, it does seem like the same principle.  But do you think that belongs in VF?  It could be a much more complex data model than EconomicResource, and I'm pretty sure has already a few standards around it (from when I looked around the web).  I'm thinking we need the durations/intervals/times so that we can feed calendars as needed.  But didn't see much else that was needed, just to make sure we can interface.  In the doc for the owl time vocab, it said it can map to iCalendar, for example (if I remember correctly).

fosterlynn:
>> all the time properties can be = or > than the quantity properties (the quantity properties are what is counted, the time properties are what relates to scheduling)

>Not sure about that, let's look at some more examples and compare it to having just one temporal quantity and one lenght/mass or dimensionless unit.

That is one of the advantages of doing it this way, where the quantities observed don't mix with the times.  The times are never used for quantities - just for times, which don't imply any quantity.

examples/work.yaml
elf-pavlik: > Then how is it an Intent? And how do you know if it is an offer or request?

I think we will need http://dublincore.org/documents/dcmi-terms/#terms-creator

> doesn't that create a Commitment made by the receiver? In fact there isn't a CfA really, no conversation is needed.

I'll double check but in CfA only *performer* can make commitment to *customer*, in VF we would have *provider* making commitment to *receiver*. It makes sense to me since *receiver* benefits by receiving something they value, only *provider* can commit to do so since in most cases they will need to or did some efforts.

What problem do you see to at this stage omit CfA details and just have result where Alice makes that commitment?

fosterlynn: >What problem do you see to at this stage omit CfA details and just have result where Alice makes that commitment?

I see no problem at all.

elf-pavlik:
This taxi use case we can also take in *composition* milestone. I can imagine that a driver would make commitment to use the car for the whole shift, possibly also record one *use* for that whole shift and at the same time smaller *use*, *IncludedIn* the all shift one, for each of the rides.
I think `affectedQuantity` should have just `1 unit:Number` since they only use one car and units should match the units affected resource has `currentQuantity`. `observedTime` seems ok with that proper interval. We might simply need `additionalQuantity` which can have any number of other relevant quantities other than quantity of affected resource and duration of the event. Distance doesn't affect `currentQuantity` of the resource or the schedlue, it would fit well in `additionalQuantity`.

examples/use-partial-resource.yaml
elf-pavlik: What available quantity of the resource one would calculate at `2018-10-20T09:00:00-5:00` ?

fosterlynn: 2.5

fosterlynn:
I made the commitments be from the provider.  And added a unitOfUse property on the resource to specify that unit, as opposed to the unit of existence.

Seems like the tradeoff is between needing to specify unitOfUse somewhere, and creating a usage resource.

We did the former in NRP, and I came to not like it that much over the course of VF discussions.  Also there may be multiple units of use, as in the taxi example.  And there may be units of citation... perhaps.  If we say that if there is a resource specified, events need to use the unit of the resource, I think that is cleaner in the long run, see https://github.com/valueflows/valueflows/pull/452.  Although we may get more clarity when we look at transfers in detail.

So right now, I'm tending towards #452  being better, although some of these examples are valid where there isn't a resource at all - so for work, people could go either way, for example.

elf-pavlik:
Another way of capturing taxi exapmle

```yaml
# Example: Use recorded multiple ways

'@context':
  - https://git.io/vf-examples-jsonld-context
  - alice: https://alice.example/
    bob: https://bob.example/
    taxi: https://taxi-coop.example/

'@id': rgh:valueflows/valueflows/master/examples/use-multiple.yaml
'@graph':


  # a taxi resource

  - '@id': taxi:e23799c14-c368-4653-a584-83bf9ae8b82a
    '@type': EconomicResource
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q82650 # taxi
    trackingIdentifier: SDERW432897348294
    currentQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1

  # the taxi transports a person

  - '@id': taxi:e18c808c-929a-450d-9d0d-3b3f531bc125
    '@type': Process
    name: Transport 1 person from Union Station to ORD airport
    plannedTime:
      time:intervalStarts:
        time:inXSDDateTimeStamp: 2018-10-14T08:45:00-5:00
      time:intervalEnds:
        time:inXSDDateTimeStamp: 2018-10-14T09:15:00-5:00

  - '@id': taxi:5d2fe7da-b91b-4dab-91d9-68db80c1feb6
    '@type': EconomicEvent
    inputOf: taxi:e18c808c-929a-450d-9d0d-3b3f531bc125 # the transportation process
    action: use
    provider: https://taxi.example/
    receiver: https://taxi.example/
    affects: taxi:e23799c14-c368-4653-a584-83bf9ae8b82a # the taxi resource
    affectedQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
    additionalQuantity: # distance of the ride
      qudt:unit: unit:Kilometer
      qudt:numericValue: 23
    observedTime: # time of the ride
      time:intervalStarts:
        time:inXSDDateTimeStamp: 2018-10-14T08:45:00-5:00
      time:intervalEnds:
        time:inXSDDateTimeStamp: 2018-10-14T09:10:00-5:00

  - '@id': bob:566ec92d-f6d1-4778-a2e8-6433f804dab8
    '@type': EconomicEvent
    inputOf: taxi:e18c808c-929a-450d-9d0d-3b3f531bc125 # the transportation process
    action: work
    provider: https://bob.example/
    receiver: https://taxi.example/
    observedTime: # time of the ride
      time:intervalStarts:
        time:inXSDDateTimeStamp: 2018-10-14T08:45:00-5:00
      time:intervalEnds:
        time:inXSDDateTimeStamp: 2018-10-14T09:10:00-5:00

  - '@id': taxi:e02e3f6b-420a-4ba8-89b0-909d7363ce06
    '@type': EconomicEvent
    outputOf: taxi:e18c808c-929a-450d-9d0d-3b3f531bc125 # the transportation process
    action: service
    provider: https://taxi.example/
    receiver: https://alice.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q41222493 # taxi service
    additionalQuantity:
      - qudt:unit: unit:Kilometer
        qudt:numericValue: 23
        rdfs:comment: distance of the ride
      - qudt:unit: unit:Minute
        qudt:numericValue: 4.5
        rdfs:comment: time when idling (not moving)
    observedTime: # time of the ride
      time:intervalStarts:
        time:inXSDDateTimeStamp: 2018-10-14T08:45:00-5:00
      time:intervalEnds:
        time:inXSDDateTimeStamp: 2018-10-14T09:10:00-5:00
```

I have impression that your example tries to 'piggy back' on *use* to much to record distance and idle time. I think the *service* flow should stay responsible for including those details. Also *use* a vehicle only means that someone wants to use a vehicle. When would would record such Intent, matching systems can reasonably match it with proposals on car rental, carsharing etc. If one wants a taxi service, one might publish bundled intents to receive *load* and *unload* at specific location on specific time. I understand that you brought this taxi use case as something that involves recording some additional quantities.

I think we would simplify everything if `affectedQuantity` just captures the affected quantity (<= `currentQuantity` of affected resource) and it affects available quantity (`involvesQuantity` for the future). `observedTime` just captures the duration of *use* or *work* and that impacts schedules (`intendedTime` and `commitedTime` for future). In addition we may need to add more open ended `affectedQuantity` for cases where need arises to record quantities not impacting the available quantity (of the resource) and the schedule. Let's take another look at the set of use cases you proposed here and that we have in other examples and see where such approach can work and where it seems to miss something.

fosterlynn:
>I have impression that your example tries to 'piggy back' on _use_ to much to record distance and idle time. I think the _service_ flow should stay responsible for including those details.

Maybe it is a stretch.  But what is the quantity of the service?  I'm thinking 1 Number....  The service basically got a person from A to B.  Then the person has to pay based on a calculation which uses miles and time. 

> Also _use_ a vehicle only means that someone wants to use a vehicle.

Actually I don't see `use` that way.  In an event I think it means actually used.  And I think it should remain as an input action.  I think in the taxi example, the person transported doesn't want to "use" a taxi, they want a service, which involves the taxi and a bunch of other inputs.  The service as what people would request and exchange for.  If you are exchanging your work or use of your equipment, where those are inputs, that is different.

We could do a translation example - that is another place that people sometimes don't use time, for example people have used words as a unit of measure for work in translating, editing, proofreading, etc.

fosterlynn:
>I think we would simplify everything if `affectedQuantity` just captures the affected quantity (<= `currentQuantity` of affected resource) and it affects available quantity (`involvesQuantity` for the future). `observedTime` just captures the duration of _use_ or _work_ and that impacts schedules (`intendedTime` and `commitedTime` for future). In addition we may need to add more open ended `affectedQuantity` for cases where need arises to record quantities not impacting the available quantity (of the resource) and the schedule.

First off, I'd like to never use time attributes to derive quantities.  Time attributes seem best related to availability, scheduling, timelines, accounting periods, etc. - time stuff.  Although some quantities are based on time, quantities are used for different purposes, and this seems a useful and clean way of dividing things.  (This has come to me slowly as we work through it.)  And lack of availabilty != actual usage/work in many cases.  And locating a flow on a timeline != actual usage/work in many cases.

Trying to summarize what we need for quantities (not availability, that is time or current quantity less commitments based on timing, that kind of thing):
* use: the quantity related to the usage (usually time, but not always) that goes into a process, possibly the quantity of the material stock resource being used when < the stock resource quantity
* work: the quantity worked (usually time but not always)
* consume: the quantity decremented from the resource that goes into a process
* cite: some quantitative representation of the value of the citation to a process; also 1 per event/resource
* produce: quantity of resource produced from the process
* service: quantity of service(s) produced from the process
* accept: quantity of the resource accepted into the process
* improve: quantity of the resource coming out of the process (same resource, so the same unless there was a problem)
* load: quantity of material resource loaded into the transportation process
* unload: quantity of material resource unloaded as output of the transportation process (same resource, so the same unless there was a problem)

Missing anything?

Terms we have talked about in terms of quantity: affected, involved, observed, used, additional.  Miss any?

elf-pavlik:
> But what is the quantity of the service? I'm thinking 1 Number....

*service* flow can have processes as both input and output and doesn't *affect* any resource. So it would have no *affectedQuantity*, it can have duration which I don't think 'belongs' to any schedule and any number of `additionalQuantities` that someone finds relevant to describe the service.

> We could do a translation example - that is another place that people sometimes don't use time, for example people have used words as a unit of measure for work in translating, editing, proofreading, etc.

Sounds like another fit for `additionalQuantity` which doesn't refer to quantity of affected resource or duration of the flow itself. I would probably also model it as *service* output by a process which uses 'raw' *work* of a person and possibly some equipment or another *service* inputs from something like management process.

> First off, I'd like to never use time attributes to derive quantities. Time attributes seem best related to availability, scheduling, timelines, accounting periods, etc. - time stuff. Although some quantities are based on time, quantities are used for different purposes, and this seems a useful and clean way of dividing things.

I don't think we can always have some 'main' quantity related to the flow. We can observe its (start and duration | end), Flows that affect some resource can affect <= `currentQuantity` of that resource. And as it seems with services we may have all kinds of additional relevant quantities: words count, distance etc.

Why do you try to always expect just one 'main' quantitative value on a flow?

fosterlynn:
>_service_ flow can have processes as both input and output and doesn't _affect_ any resource. So it would have no _affectedQuantity_

What about an observedQuantity?  But yes, can be both input and output.  Not sure of implications...

>Sounds like another fit for `additionalQuantity` which doesn't refer to quantity of affected resource or duration of the flow itself.

I could go with something like that - but we need some way to understand what the various quantities should be used for, and I'm not sure additionalQuantity will work for that, seems like it could have different meanings?  

> I would probably also model it as _service_ output by a process which uses 'raw' _work_ of a person and possibly some equipment or another _service_ inputs from something like management process

We've done it with accepts/improves before.  In the use cases I'm familiar with it really isn't a service.  In some other use cases I can imagine it is.  But I'll need to do it in yaml to think about it in detail.

>Why do you try to always expect just one 'main' quantitative value on a flow?

Because it is easier to query, report, and apply formulas to, for many functions, in my experience.  I picture a needless and somewhat painful increase in complexity to not have that.  I think that if the same function can look at the same property, that is a very good thing.  But I'm willing to explore your ideas for alternatives.  I'll use this https://github.com/valueflows/valueflows/pull/446#issuecomment-462071159 and play around and see where it leads.....

fosterlynn:
>    additionalQuantity:
      - qudt:unit: unit:Kilometer
        qudt:numericValue: 23
        rdfs:comment: distance of the ride
      - qudt:unit: unit:Minute
        qudt:numericValue: 4.5
        rdfs:comment: time when idling (not moving)

I understand it makes some sense to think of defining the service in terms of the measurements required to figure out the payment.  But I don't see how to see what each additional quantity is for.  Seems like we would need something that looks a lot like a resource classification/specification for each 'piece' of the service?  Especially if 2 of them had the same unit.

Possibly the service of transportation could vf:contain services of transporting a distance, idling time, bags transported, etc.?

elf-pavlik:
first take on carpenters with Intent - satisfaction -> Commitment - fulfillment -> Event

```yaml
# Example: Use a portion of a stock resource

'@context':
  - https://git.io/vf-examples-jsonld-context
  - alice: https://alice.example/
    carpenters: https://carpenters.example/

'@id': rgh:valueflows/valueflows/master/examples/use-partial-resource.yaml
'@graph':


  # tools used by the group of carpenters

  - '@id': carpenters:9f112621-3f13-4f48-a3e2-fd2ca99493ca
    '@type': EconomicResource
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q58964 # stock drills
    currentQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 10

  - '@id': carpenters:8791913c-0152-4c32-bb3f-f25d61df736b
    '@type': EconomicResource
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q2365675 # a table saw
    trackingIdentifier: 38922-44
    currentQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1

  # the carpenters are planing and doing the process

  - '@id': carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125
    '@type': Process
    name: Make a small shed
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-20T08:00:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-20T13:00:00-5:00

  # intents - possibly auto generated from a recipe

  - '@id': carpenters:b5be262f-9eb4-45fd-80e8-7568073de0b9
    '@type': Intent
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    receiver: https://carpenters.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q58964 # stock drills
    involvesQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 3
    time:hasDuration:
      time:unitType: time:unitHour
      time:numericDuration: 4
    time:intervalStarts: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125

  - '@id': carpenters:811a433f-5af5-4425-9e27-f13c45c3ef43
    '@type': Intent
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    receiver: https://carpenters.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q2365675 # a table saw
    involvesQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 1
    time:hasDuration:
      time:unitType: time:unitHour
      time:numericDuration: 2
    time:intervalStarts: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125

  # commitments - self-made or made by another agents

  - '@id': carpenters:4da7a49b-44ea-49d6-81a5-572d92158c2d
    '@type': Commitment
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    provider: https://carpenters.example/
    receiver: https://carpenters.example/
    involves: carpenters:9f112621-3f13-4f48-a3e2-fd2ca99493ca # drills
    involvesQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 3
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-20T08:00:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-20T12:00:00-5:00

  - '@id': carpenters:10173de4-8624-47a8-869e-6157742055ef
    '@type': Satisfaction
    satisfies: carpenters:b5be262f-9eb4-45fd-80e8-7568073de0b9
    satisfiedBy: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    satisfiedQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 3
    satisfiedInterval:
      time:hasBeginning:
        time:inXSDDateTimeStamp: 2018-10-20T08:00:00-5:00
      time:hasEnd:
        time:inXSDDateTimeStamp: 2018-10-20T12:00:00-5:00

  - '@id': carpenters:8a0b6280-8d7c-43ca-a99b-0d90351353e4
    '@type': Commitment
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    provider: https://carpenters.example/
    receiver: https://carpenters.example/
    involves: carpenters:carpenters:8791913c-0152-4c32-bb3f-f25d61df736b # table saw
    involvesQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 1
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-20T08:30:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-20T10:30:00-5:00

  - '@id': carpenters:f5c023bb-defa-4d1a-8d97-1b735e215ca0
    '@type': Satisfaction
    satisfies: carpenters:811a433f-5af5-4425-9e27-f13c45c3ef43
    satisfiedBy: carpenters:8a0b6280-8d7c-43ca-a99b-0d90351353e4
    satisfiedQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
    satisfiedInterval:
      time:hasBeginning:
        time:inXSDDateTimeStamp: 2018-10-20T08:30:00-5:00
      time:hasEnd:
        time:inXSDDateTimeStamp: 2018-10-20T10:30:00-5:00

  # events 

  - '@id': carpenters:e72f7d1d-4f0a-4518-a44c-96658cafda94
    '@type': EconomicEvent
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    provider: https://carpenters.example/
    receiver: https://carpenters.example/
    affects: carpenters:9f112621-3f13-4f48-a3e2-fd2ca99493ca # drills
    affectedQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 3
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-20T08:15:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-20T12:00:00-5:00

  - '@id': carpenters:670da586-cab7-4ae4-99ef-2b641daaa439
    '@type': Fulfillment
    fulfills: carpenters:4da7a49b-44ea-49d6-81a5-572d92158c2d
    fulfilledBy: carpenters:e72f7d1d-4f0a-4518-a44c-96658cafda94
    fulfilledQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 3
    fulfilledInterval:
      time:hasBeginning:
        time:inXSDDateTimeStamp: 2018-10-20T08:15:00-5:00
      time:hasEnd:
        time:inXSDDateTimeStamp: 2018-10-20T12:00:00-5:00

  - '@id': carpenters:f325a950-5737-488d-b122-8f21114d0eb9
    '@type': EconomicEvent
    inputOf: carpenters:e18c808c-929a-450d-9d0d-3b3f531bc125 # the process
    action: use
    provider: https://carpenters.example/
    receiver: https://carpenters.example/
    affects: carpenters:carpenters:8791913c-0152-4c32-bb3f-f25d61df736b # table saw
    affectedQuantity: # portion of the stock resource that is used
      qudt:unit: unit:Number
      qudt:numericValue: 1
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-20T08:45:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-20T10:25:00-5:00

  - '@id': carpenters:fc70badf-e71a-4980-ae53-e807eee67d5b
    '@type': Fulfillment
    fulfills: carpenters:8a0b6280-8d7c-43ca-a99b-0d90351353e4
    fulfilledBy: carpenters:f325a950-5737-488d-b122-8f21114d0eb9
    fulfilledQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
    fulfilledInterval:
      time:hasBeginning:
        time:inXSDDateTimeStamp: 2018-10-20T08:45:00-5:00
      time:hasEnd:
        time:inXSDDateTimeStamp: 2018-10-20T10:25:00-5:00
```

fosterlynn:
I changed the work facilitation event to reference an accounting period.   (This is the FairCoop example.  Also the buying club example for work-at-home people.)  We can work with the names if you like the idea.

examples/work.yaml
```diff
@@ -147,10 +147,18 @@
       qudt:unit: unit:Hour
       qudt:numericValue: 21.5
     observedTime:
+      '@type': time:Instant
+      time:inXSDDateTimeStamp: 2018-10-31T00:00:00-5:00
+    accountDuring: org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
+
+  - @id': org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
+    '@type': AccountingPeriod
```
elf-pavlik: Let's see if we can't just use 'upper' level Process in *composition* milestone. Possibly some work, usage maybe even consumption needs to recorded for that 'accounting period' and representing it as Process would make more sense.

fosterlynn: Hmmm, I don't think of it as a Process, but I'm fine waiting until that milestone to define it.  It does fit with the budgeting stuff from the SolidBase crew.  In addition to this operational usage.  Maybe it is often more of an operational period than an accounting period... ?

Thoughts about how to represent the time property in the FairCoop facilitation UC and the buying club UC?

elf-pavlik: I'll take a look at UC which provides facilitation services based on work and usage of computer, workspace etc. In practice many remote work services require more than just raw work and dependent of usage of various other resources.

fosterlynn: In these UCs, a service wouldn't work, it is basically people, either getting paid by the hour or getting buying club privileges in exchange for worked hours.  We can not worry about it now.  My question was more about how to represent the time if it isn't the duration of the accounting/operational period?

elf-pavlik: The main challenge I see comes with a fact that we don't try to represent an actual *work* interval that occurred at specific time. It seems more like a report based on not recorded events which includes sum from those unrecorded events. In a way a duration of non continuous interval.

>  My question was more about how to represent the time

If we just need some time interval we can always use instances of classes from OWL-Time, for example `time:ProperInterval`. I think VF only needs to focus on what do those time entity mean for the economic environment and what their relations with other entities mean.

fosterlynn: I'm totally fine with using the ProperInterval or whatever and it being a month or whatever, but you didn't like that too much.  It then only relates to the timing that matters to them, and of course doesn't represent the hours actually worked.  Often people also only care about due date, or start date, for example.  I guess the question is do we leave it so people can do what they want for recording time, depending on how much coordination they need?  Or do we define that it must be the actual duration used or worked?  I support the first, based on what people actually do.  I don't see any down side from the network standpoint, or larger planning or anything.  As long as the quantities are there as needed.  It just says this quantity happened sometime(s) during this time.

I do understand your challenge, we had the same first response and tried to recommend entering all actual time.  But it actually doesn't make sense for some kinds of work - sometimes they are meant to be around the right amount, but work more like "salary" than "wages".  Sometimes they do write it down and just record the sum (I do that for the buying club); and sometimes the work is like a few minutes here and there, and doesn't make sense to record exactly.

elf-pavlik:
That FC inspired example could look this way:
```yaml
  - '@id': org:e4783bef-9006-490c-9c03-389272c7444c
    '@type': EconomicEvent
    action: work
    receiver: https://org.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3064143 # facilitation 
    time:hasDuration:
      time:unitType: time:unitHour
      time:numericDuration: 21.5
    time:intervalDuring: org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2

  - '@id': org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
    '@type': time:ProperInterval
    skos:note: October 2018
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-01T0:00:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-31T23:59:59-5:00
```

So the *work* event only has specific duration and position itself during the month of October 2018 instead of giving precise beginning and end. I still need to see if we could use OWL-Time directly for such *non-continous* durration, since we don't represent a continues interval lasting 21.5 hour but more of a total duration (aggregate) of shorter unrecorded intervals.
I will add snippet for someone who actually recorded exact proper intervals and still wants to provide aggregate for month of October, which looks like touching *composition* milestone.

elf-pavlik:
I think we have something like this in this FC work time summaries:

non-continous interval

On the bottom we have ProperIntervals representing actual observable *work* events. On the top we have more of a report / aggregate which sums up all the duration of actual observable events.

fosterlynn:
>So the _work_ event only has specific duration and position itself during the month of October 2018 instead of giving precise beginning and end

I think this goes in the right direction.  Maybe the ProperInterval could actually be the accounting period (or whatever we call it), a little more specifically defined.  Could inherit.

But, I think using quantity instead of duration for the 21.5 hours worked could solve all your other problems with it.

elf-pavlik:
> But, I think using quantity instead of duration for the 21.5 hours worked could solve all your other problems with it.

Let's take similar report / aggregate case but with *use* which involves quantity of a resource. Maybe some agent has agreement with `ouikit` that they can use their resource of 'second grade' kits without scheduling it and just should once a month report how many they used for how much of time. In those reports we will have both
* affected quantity of the resource - 'second grade' kits
* total (non continuous) duration of using them

Still I agree that if `time:` duration properties can't support such sum / aggregate, we could define duration property in `vf:` namespace which doesn't imply Proper Interval (continuous one)

fosterlynn:
Merriam-Webster dictionary

>Definition of duration

>1 : continuance in time gradually increase the duration of your workout
>2 : the time during which something exists or lasts were there for the duration of the concert

I don't think durations can be broken up - that is their point, like calendar durations which are used for scheduling etc., irrespective of actual time used or worked or whatever during the time period

elf-pavlik:
I think we could safely go with

```yaml
  - '@id': org:e4783bef-9006-490c-9c03-389272c7444c
    '@type': EconomicEvent
    action: work
    receiver: https://org.example/
    classifiedAs: https://www.wikidata.org/wiki/Q3064143 # facilitation 
    temporalLength:
      qudt:unit: unit:Hour
      qudt:numericQuantity: 21.5
    time:intervalDuring: org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2

  - '@id': org:9087abc4-6f2b-4f1f-9c56-611b7f7302a2
    '@type': time:ProperInterval
    skos:note: October 2018
    time:hasBeginning:
      time:inXSDDateTimeStamp: 2018-10-01T0:00:00-5:00
    time:hasEnd:
      time:inXSDDateTimeStamp: 2018-10-31T23:59:59-5:00
```

We could require that even if flow has `time:hasBeginning` and `time:hasEnd` defined one has to include `temporalLength` and not expect other applications to compute it. I'll also update ouikit example which also has `involvesQuantity`.

fosterlynn:
>    temporalLength:
      qudt:unit: unit:Hour
      qudt:numericQuantity: 21.5

>I understand that we agree that we can't just have one quantitative value and try to define what it means for each action. For example we can't say quantity for use relates to temporal length of that usage.

So does that mean that instead of the above temporalLength (which is the only quantity in that example in the previous comment here), we have involvedQuantity or flowQuantity, or similar?  Or do I misunderstand?

elf-pavlik:
Please see https://github.com/valueflows/valueflows/pull/458#issuecomment-469484754

Since we don't have any `involves` reference to a resource neither `resourceClassifiedAs`, we have no need for `involvesQuantity`. In this case we only need `temporalLength`. Ouikits example might need both since it `involves` specific resource.

fosterlynn:
I think the way you structured the Oct 2018 period in https://github.com/valueflows/valueflows/pull/446#issuecomment-469471915 works well (some details TBD).  In particular for accounting period kinds of things, where some number of flows will reference the same one.

I'm thinking also about commitments' time needs - as opposed to events, where the exact time is observable.  In commitments, it seems like sometimes you need to commit to an exact time, say to do some work, because that is required to coordinate schedules.  Other times, you may just have a due date.  Or sometimes you may have a start date and a due date both, but they include more time than needed for the work.  But it is not related to an accounting period or proper interval that has some meaning, it is just the way the sequence of commitments works out on the schedule.  Seems excessive to have a separate ProperInterval for that kind of thing, and the event(s) that fulfill the commitment won't have that either.

elf-pavlik:
I think [OWL Time: Topological Temporal Relations](https://www.w3.org/TR/owl-time/#topology) should handle that. Let's just try it out with some use cases.

PR #452, opened 2019-02-07
alt work/usage examples

fosterlynn:
Started with the work example.  Putting here just for discussion!

First question: If a person has a work resource (aka "labor power"), represented on a schedule, when some work gets intended, committed, and done and that resource is referenced, is is "work" or "consume"?  Consume isn't consistent either, because although it is consuming the work capability time on the schedule, it doesn't consume it by subtracting a quantity - the work capability doesn't really have a quantity in any useful way.

Anyhow, continuing to think.


fosterlynn:
>I like possibility of modeling each skill of a person as a resource

Yes seems like it has possibilities.

> later process may require _use_ of those skills and still _work_ of a person.

I think we ought to be careful about using the same action for people and equipment, would prefer `use` stays with non-people at this point.

>I wouldn't see _consume_ on skills.

It wouldn't consume skills, it would consume the person's "labor power" or "work capacity" or something like that.  But, I actually like it less and less myself anyhow.  

At this point I like just using `work` whether there is a resource or not.  But have to try it out with the use case about availability to work vs working.

>As discussed in #446 I don't think fablab can make commitment that alice will provide something for them, they can create an Intent that they would like alice to provide something.

How do you see it working in e-commerce?  Like a person orders something online, based on the offer published.  Is it only a commitment when the provider comes back with order confirmation?  And is the namespace always the provider's?

Online, I can see it can work either way and there is communication back to the customer.  But if this is not a shopping cart, and an order is just placed with no response (like the buying club we are in), it seems like there is some commitment from the purchaser to receive those veggies - so that the seller can count on those being committed in a calculation of available quantity.  And can go ahead and send them.

elf-pavlik:
> Online, I can see it can work either way and there is communication back to the customer. But if this is not a shopping cart, and an order is just placed with no response (like the buying club we are in), it seems like there is some commitment from the purchaser to receive those veggies - so that the seller can count on those being committed in a calculation of available quantity. And can go ahead and send them.

We could take this use case once we dive into Conversation for Action more deeply. We haven't included intents into agreements yet. Possibly you could have intent to receive specific list of items provided by specific agent. Together in the same agreement you would very likely have a Commitment to provide something in exchange. I think in the buying club you mention, they deliver food because you made commitment to pay for it and not only because you 'commit' to receive the food.

Can we at this stage keep it one a safer side and have always *provider* making the Commitment? Keeping in mind that when we work on Conversation for Action we may need to expand or adjust vocab to provide all needed steps.

fosterlynn:
>Can we at this stage keep it one a safer side and have always _provider_ making the Commitment?

OK will work with that.

fosterlynn:
Added a simple usage example.

fosterlynn:
Added the drill example.  Decided to leave it with the same technique used on the earlier examples.  Otherwise, we are in the business of creating a new usage resource before starting, using issue/give.  Which might be done in a more controlled situation, but then they would have serialized resources probably.  (In fact I still strongly suspect the drills is not a real-life use case, and would be fine with requiring usage to reference either a serialized resource or a whole stock resource.)

fosterlynn:
Added the taxi example. 

Working on formulating a set of rules that gives some flexibility but not too many choices, because of the added complexity.  Something like this, in order:
1. If there is a ResourceSpecification with a unit specified, that is the unit that resources and events must use.
2. Else, if there is a EconomicResource with a unit specified, that is the unit that events must use.
3. Else, events can use what they want.

fosterlynn:
>I have impression that your example tries to 'piggy back' on _use_ to much to record distance and idle time. I think the _service_ flow should stay responsible for including those details.

Modified the taxi scenario, based on discussion in #446 .

Moved the distance and idling time to output contained services.  It also occurred to me that `use` should be about the process from more of an internal perspective, so I now completely agree with your comments.  To illustrate, I made the input distance recorded on `use` of the taxi larger, to include some distance driving around before the ride.  This will record total distance of the taxi over time, which can be used for maintenance scheduling, etc. - all related to taxi processes, not the service provided and probably exchanged in some way.

Also split out the pieces of the service that are quantified so you can tell what the quantities refer to.

fosterlynn:
Cleaned up the work example, especially the namespaces.  Eventually do signatures may come into effect somewhere here?  And I think that provider and receiver (and scope if applicable) will have copies of the records.... would they have a copy with their id in their namespace?  Hopefully not....

fosterlynn:
In working on the drill-type example, I think I need transfers to make it work with one resource/one unit solution.  Also, I want to see if transfers will want the same pattern of underlyingResource as I am experimenting with here.  If so, it seems to me this might be more compelling.

Otherwise, seems like there are options: 
* usage and work separate resources (usage having underlying material resource) - keeping one unit per resource, one resource per event - like this PR
* defining alternate units for usage (and maybe citation, and maybe giving rights, lending, etc.) in material resources that are used (or cited or lent etc.) - and just letting work not have any resource
    * use observedQuantity for the counted quantity, adding others as needed (maybe affected for the drills assuming no transfer, etc.)
    * use affected when there is a material resource that is incremented/decremented, others for usage, work, etc. - define a set of quantity fields that match what is actually happening in the action, then using action to parse what quantity to use

Did I miss anything and did I describe them correctly?  Perhaps something is still missing on how you want to handle availability?  Please correct me, as I'm trying to not be biased but maybe I am anyhow.

PR #453, opened 2019-02-09
improve integration with OWL-Time

elf-pavlik:
This reduces one level of indirection and duplication in the vocab. We can simply consider any `Intent`, `Commitment` and `EconomicEvent` as more specific `time:TemporalEntity` and use properties defined in [Time Ontology in OWL](https://www.w3.org/TR/owl-time/#namespaces) directly one them. Since we already have distinct `Intent`, `Commitment` and `EconomicEvent` we don't need in addition `observed*`, `commited*` and `intended*` prefixes. One can simply check `rdf:type` (`@type`) of flow to have that knowledge.

Our examples already incorrectly use `time:intervalEnds` and `time:intervalStarts` (which have `time:ProperInterval` as range) we have to update them to `time:hasBeggining` and `time:hasEnd` (which have `time:Instant` as range). If we accept this PR, we can use those properties directly on any flow or Process.

### DIFF
--- a/docs/specification/external-terms.md
+++ b/docs/specification/external-terms.md
@@ -44,8 +44,6 @@ https://www.w3.org/TR/owl-time
 * [`time:before`](https://www.w3.org/TR/owl-time/#time:before)
 * [`time:after`](https://www.w3.org/TR/owl-time/#time:after)
 * [`time:intervalDuring`](https://www.w3.org/TR/owl-time/#time:intervalDuring)
-* [`time:intervalStarts'](https://www.w3.org/TR/owl-time/#time:intervalStarts)
-* [`time:intervalFinishes'](https://www.w3.org/TR/owl-time/#time:intervalFinishes)
 
 `vf` defines two property chain axioms `vf:hasBeginning` and `vf:hasEnd` as slight variant
 of [Alignment of PROV-O with OWL-Time(https://www.w3.org/TR/owl-time/#time-prov)]
--- a/docs/specification/external-terms.md
+++ b/docs/specification/external-terms.md
@@ -20,3 +20,32 @@ http://xmlns.com/foaf/spec/
 https://www.w3.org/TR/vocab-org/
 
 * [`org:Organization`](https://www.w3.org/TR/vocab-org/#org:Organization)
+
+## QUDT
+
+http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html
+
+* [`qudt:QuantityValue`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Classes)
+* [`qudt:unit`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Properties)
+* [`qudt:numericValue`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Properties)
+
+### UNIT
+
+http://qudt.org/doc/2017/DOC_VOCAB-UNITS-BASE.html
+
+* [`unit:Number`](http://qudt.org/doc/2017/DOC_VOCAB-UNITS-BASE.html#Instances)
+
+## TIME
+
+https://www.w3.org/TR/owl-time
+
+* [`time:inXSDDateTimeStamp`](https://www.w3.org/TR/owl-time/#time:inXSDDateTimeStamp)
+* [`time:hasDuration`](https://www.w3.org/TR/owl-time/#time:hasDuration)
+* [`time:before`](https://www.w3.org/TR/owl-time/#time:before)
+* [`time:after`](https://www.w3.org/TR/owl-time/#time:after)
+* [`time:intervalDuring`](https://www.w3.org/TR/owl-time/#time:intervalDuring)
+* [`time:intervalStarts'](https://www.w3.org/TR/owl-time/#time:intervalStarts)
+* [`time:intervalFinishes'](https://www.w3.org/TR/owl-time/#time:intervalFinishes)
+
+`vf` defines two property chain axioms `vf:hasBeginning` and `vf:hasEnd` as slight variant
+of [Alignment of PROV-O with OWL-Time(https://www.w3.org/TR/owl-time/#time-prov)]
--- a/examples/fulfill-satisfy.yaml
+++ b/examples/fulfill-satisfy.yaml
@@ -68,14 +68,8 @@
     affectedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
-    observedTime:
-      '@type': time:ProperInterval
-      time:intervalStarts:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-0:00
-      time:intervalEnds:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T12:00:00-0:00
+    hasStart: 2018-10-14T8:00:00-0:00
+    hasEnd: 2018-10-14T12:00:00-0:00
 
   - '@id': mfg:6f438393-7f87-4914-806c-e23a4fd15e89
     '@type': Fulfillment
@@ -94,14 +88,8 @@
     affectedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
-    observedTime:
-      '@type': time:ProperInterval
-      time:intervalStarts:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T13:00:00-0:00
-      time:intervalEnds:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T17:00:00-0:00
+    hasStart: 2018-10-14T13:00:00-0:00
+    hasEnd: 2018-10-14T17:00:00-0:00
 
   - '@id': mfg:0f563083-8da4-46fe-adc3-68b05ba06320
     '@type': Fulfillment
--- a/examples/fulfill-satisfy.yaml
+++ b/examples/fulfill-satisfy.yaml
@@ -68,7 +68,7 @@
     affectedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
-    hasStart: 2018-10-14T8:00:00-0:00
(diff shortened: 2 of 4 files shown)


fosterlynn:
I agree we should fix the technicalities with our use of the owl:time spec.  But I don't agree at all to make Flow and Process subclasses of anything.  I'm looking at the semantic side, not the rdf technicalities.  I don't even think that times are all that critical on some flows - like say commitments that are part of a process and could just as easily imply their times from the process plannedTime. But in any case, Flows and Processes are not times, and are not that similar to each other.  They have times.  But the times mean different things.

fosterlynn:
>Since we already have distinct `Intent`, `Commitment` and `EconomicEvent` we don't need in addition `observed*`, `commited*` and `intended*` prefixes

I am however OK with trying to think of a name we can use for all of them for time (and quantities too).

elf-pavlik:
> like say commitments that are part of a process and could just as easily imply their times from the process plannedTime.

Time Ontology actually makes it quite straight forward:

```yaml
'@id': ex:bb1388fa-57f9-4152-be33-31ac267cca4f
'@type': Process

'@id': ex:a35d58b1-5839-4d1b-a325-1c6d3f5d3d9d
'@type': Commitment
time:intervalDuring: ex:bb1388fa-57f9-4152-be33-31ac267cca4f
```

One can also state that one process needs to start after another one finished, or that process finishes when *create* event finishes etc.

time-owl

fosterlynn:
But fundamentally, subclass pretty universally means "is a" or like "is a type of".  And Process and flow types are not TemporalEntities, that is such a small part of what they are semantically.  We might even decide at some point there is a better time representation we want to use.

So to me, time is much better represented as properties of the classes we have, just like quantity, etc.

elf-pavlik:
Relevant section from the spec which describes alignment with PROV: https://www.w3.org/TR/owl-time/#time-prov which has a lot in common with VF.

I find considering flows and processes as temporal entities very natural. This way one can also directly represent them on timeline (eg. schedule). Using Time Ontology we can express flows like:

```yaml
'@id': ex:eb6b2521-8d46-4ef9-85fa-087585a56b73
'@type': Commitment
action: use
involves: ex:c9c8c652-f1a1-479d-a50b-05beaa9825ac
involvesQuantity:
  qudt:unit: unit:Number
  qudt:numericValue: 8
time:hasDuration:
  time:numericDuration: 2
  time:unitType: time:unitHour
time:intervalDuring:
  time:hasBeggining:
    time:inXSDDateTimeStamp: 2018-10-15T8:00:00-0:00
  time:hasEnd:
    time:inXSDDateTimeStamp: 2018-10-15T18:00:00-0:00
```

use 8 pcs for 2 hours on Oct 15 2018, between 8:00 and 18:00 UTC

elf-pavlik:
also valid:
```yaml
time:hasDurationDescription:
  time:hours: 2
```

fosterlynn:
> This way one can also directly represent them on timeline (eg. schedule).

We can do that with time properties too.  Can do everything we need to with time properties.

What would you do for non-rdf implementations?  Still create a super-class?  Seems very owl LOD specific.  Seems OK to me to borrow LOD definitions for simple properties of VF classes, but not so much to inherit from non-VF classes, too fundamental to the model.

>time:intervalDuring:

I leave to you to pick the best for durations (and instants, not every flow time is an actual duration).

elf-pavlik:
> What would you do for non-rdf implementations? Still create a super-class? Seems very owl LOD specific.

I don't think `time:TemporalEntity` has to appear directly in any implementation. 

> Seems OK to me to borrow LOD definitions for simple properties of VF classes, but not so much to inherit from non-VF classes, too fundamental to the model.

We don't talk about any OOP inheritance, actually due to `rdfs:domain` and `rdfs:range` definitions as soon as someone uses `time:intervalBefore` with an instance of a flow, the reasoner can infer from that flow also has type `time:ProperInterval`. 

Until recently we had directly `start` and `duration` property directly on flows: https://github.com/valueflows/valueflows/pull/291/files#diff-a5832671e1376b69e3b7f7fb50526009L342

I support reusing OWL-Time but I think we can do it in a way that we had before and just use properties from `time:` directly on flows as my snippets in comments above illustrate.

fosterlynn:
>I support reusing OWL-Time but I think we can do it in a way that we had before and just use properties from `time:` directly on flows as my snippets in comments above illustrate.

OK, so no subclassing I understand. :+1:   

I'll switch to your
```
time:intervalDuring:
  prov:endedAtTime: 2018-10-15T8:00:00-0:00
  prov:startedAtTime: 2018-10-15T18:00:00-0:00
```
That replaces this, right?
```
      '@type': time:ProperInterval
      time:intervalStarts:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2018-12-07T9:00:00-5:00
      time:intervalEnds:
        '@type': time:Instant
        time:inXSDDateTimeStamp: 2018-12-07T15:00:00-5:00
```

fosterlynn:
```
time:intervalDuring:
  prov:endedAtTime: 2018-10-15T8:00:00-0:00
  prov:startedAtTime: 2018-10-15T18:00:00-0:00
```

Question: does this work also for intents and commitments?  Or is prov assuming observed times?  (The names imply observed times.)

elf-pavlik:
Please wait with updating your examples, I think we can actually for this case define properties in `vf:` namespace and include OWL property chain axioms as in https://www.w3.org/TR/owl-time/#time-prov

release-doc-in-process/all_vf.TTL
```diff
@@ -46,6 +46,7 @@ vf:RecipeProcess a          owl:Class ;
 
 vf:RecipeFlow a             owl:Class ;
         rdfs:label          "vf:RecipeFlow" ;
+        rdfs:subClassOf     time:TemporalEntity ;
```
fosterlynn: I still am not comfortable with the subclassing - the broad understanding of that term is like "is a", "is a type of", etc. And I think it is conceptually misleading.

examples/fulfill-satisfy.yaml
```diff
@@ -68,14 +68,8 @@
...
       qudt:numericValue: 4
-    observedTime:
-      '@type': time:ProperInterval
-      time:intervalStarts:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T8:00:00-0:00
-      time:intervalEnds:
-        '@type': time:Instant
-        time:inXSDDateTimeStamp: 2018-10-14T12:00:00-0:00
+    hasStart: 2018-10-14T8:00:00-0:00
```
fosterlynn: hasBeginning?

release-doc-in-process/all_vf.TTL
```diff
@@ -270,13 +275,25 @@ vf:under  a                 owl:ObjectProperty ;
...
         rdfs:domain         vf:EconomicResource ;
         rdfs:range          vf:EconomicResource ;
         vs:term_status      "unstable" ;
         rdfs:comment        "Used when a stock economic resource contains units also defined as economic resources." .
 
+vf:hasBeginning a           owl:DatatypeProperty ;
+        rdfs:domain         time:TemporalEntity ;
+        rdfs:range          xsd:dateTimeStamp ;
+        owl:propertyChainAxiom (time:hasBeginning time:inXSDDateTimeStamp) ;
+        rdfs:comment        "Specific time marking the beginning of the flow".
```
fosterlynn: ....flow or process.

docs/specification/external-terms.md
```diff
@@ -20,3 +20,32 @@ http://xmlns.com/foaf/spec/
...
+
+https://www.w3.org/TR/owl-time
+
+* [`time:inXSDDateTimeStamp`](https://www.w3.org/TR/owl-time/#time:inXSDDateTimeStamp)
+* [`time:hasDuration`](https://www.w3.org/TR/owl-time/#time:hasDuration)
+* [`time:before`](https://www.w3.org/TR/owl-time/#time:before)
+* [`time:after`](https://www.w3.org/TR/owl-time/#time:after)
+* [`time:intervalDuring`](https://www.w3.org/TR/owl-time/#time:intervalDuring)
+* [`time:intervalStarts'](https://www.w3.org/TR/owl-time/#time:intervalStarts)
+* [`time:intervalFinishes'](https://www.w3.org/TR/owl-time/#time:intervalFinishes)
```
fosterlynn: That's a lot to expect apps to support.  I think we should see what is used by known real use cases, and include them, then add as needed.  After all, anyone can use any of them if they want, and have coordinated with their communication partners.

docs/specification/external-terms.md
```diff
@@ -20,3 +20,32 @@ http://xmlns.com/foaf/spec/
...
+
+* [`qudt:QuantityValue`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Classes)
+* [`qudt:unit`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Properties)
+* [`qudt:numericValue`](http://qudt.org/doc/2016/DOC_SCHEMA-QUDT-v2.0.html#Properties)
+
+### UNIT
+
+http://qudt.org/doc/2017/DOC_VOCAB-UNITS-BASE.html
+
+* [`unit:Number`](http://qudt.org/doc/2017/DOC_VOCAB-UNITS-BASE.html#Instances)
```
fosterlynn: Not clear why you are calling out this particular unit?  Even if true that we use it a lot.

release-doc-in-process/all_vf.TTL
elf-pavlik: It has very precise definition explaining exactly what it causes: https://www.w3.org/TR/rdf-schema/#ch_subclassof

I removed it for now since no need to argue about it in this PR

docs/specification/external-terms.md
elf-pavlik: Any app only will support what functionality of that app requires. Here I don't consider server hosting some dataset as an app since it can't never know what various apps reading and writing from it will need or will not need and can't assume how many different properties apps will use. Still I only left those which we already encountered in our recent discussions.

docs/specification/external-terms.md
elf-pavlik: Since we use it a lot people should know when to find it.

PR #461, opened 2019-02-28
metered electricity example

fosterlynn:
Wanted to make sure there weren't any weird things with this use case.  Seems pretty straight forward.

### DIFF
--- a/examples/stream-resource.yaml
+++ b/examples/stream-resource.yaml
@@ -1,0 +1,37 @@
+# Example: Streaming resource (electricity)
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - bob: https://bob.example/
+    elec: https://electricity-coop.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/fstream-resource.yaml
+'@graph':
+
+  # electricity is metered
+
+  - '@id': elec:da7bfa38-a1b2-4f6f-a896-b9bf1ec8807a
+    '@type': EconomicEvent
+    action: transfer
+    provider: https://electricity-coop.example/
+    receiver: https://bob.example/
+    resourceClassifiedAs: wd:Q12725 # electricity
+    under: https://electricity-coop.example/agreement-vernon-county-2019
+    observedQuantity:
+      qudt:unit: unit:KilowattHours # KiloA?
+      qudt:numericValue: 435
+    observedTime:
+      hasEnd: 2019-01-31T10:00:00-5:00
+
+  - '@id': elec:da7bfa38-a1b2-4f6f-a896-b9bf1ec8807q
+    '@type': EconomicEvent
+    action: transfer
+    provider: https://electricity-coop.example/
+    receiver: https://bob.example/
+    resourceClassifiedAs: wd:Q12725 # electricity
+    under: https://electricity-coop.example/agreement-vernon-county-2019
+    observedQuantity:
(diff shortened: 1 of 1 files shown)


elf-pavlik:
How about:

```yaml
  - '@id': elec:da7bfa38-a1b2-4f6f-a896-b9bf1ec8807a
    '@type': EconomicEvent
    action: consume
    provider: https://electricity-coop.example/
    receiver: https://bob.example/
    resourceClassifiedAs: wd:Q12725 # electricity
    under: https://electricity-coop.example/agreement-vernon-county-2019
    affects: elec:087dbbc6-0e4a-48e5-b15d-566977040d7c # electric grid bob stays connected to
    affectedQuantity:
      qudt:unit: unit:KilowattHours # KiloA?
      qudt:numericValue: 435
    observedTime:
      hasBeginning: 2019-01-01T10:00:00-5:00
      hasEnd: 2019-01-31T10:00:00-5:00
```

Bob actually *consume* the electricity from specific grid. I guess if Bob has some sub-grid we could use *issue* on the `elec` grid and then Bob would do *consume* on his sub-grid.

fosterlynn:
>`affects: elec:087dbbc6-0e4a-48e5-b15d-566977040d7c # electric grid bob stays connected to`

The electric coop could do it that way if they like.

>Bob actually _consume_ the electricity from specific grid. I guess if Bob has some sub-grid we could use _issue_ on the `elec` grid and then Bob would do _receive_ & _consume_ on his sub-grid

True Bob is consuming (or wasting) the electricity, but there is no measurement until the meter gets read, at that point in time.  There would need to be very different measurements going on internal to Bob's processes and consumption of the electricity to know how it fits into those processes I guess.  And let's not even go to the IoT ! :)  Anyhow, that would be different, whether it is transfer or consumption, not the reading that is recorded once a month that is taken by the coop.  I see consume as related just as much to the process where you are consuming as it is to the resource.  It's a bit meaningless otherwise imo.

```
    observedTime:
      hasBeginning: 2019-01-01T10:00:00-5:00
      hasEnd: 2019-01-31T10:00:00-5:00
```
Sure if they want to.  I suppose they need to look up the last event anyhow to subtract.

In any case, seems like it can be done, however they want to do it, without too many complications.  Anything wrong with how I did it?  There are options.

>Last but not least we should also have variant with day and night rates. Which seems like another case of non-continous intervals.

I don't know how they do this, I suppose just take a reading twice a day at the same time.  So it would look the same, just more of them.

fosterlynn:
BTW, I know you have thought about this, and I do think it will be good when people can start including personal/family processes in REA graphs, and thus go full circle, and consider the environment too.  I don't see thinking about it as part of getting this next round of the vocab out there though....

elf-pavlik:
Since the meter shows already consumed / issued amount of kWh I think event should reflect that fact. We could have use case for more of a 'transfer of rights' if we have utilities which give agents certain quota. I don't know of electric grids working this way but for example mobile bandwidth can have certain amount of GB of quota assigned. This could get represented as 'transfer' of rights to *consume* specific amount of data transfer in specific time period (interval).

fosterlynn:
added beginning date to the observedTime, per suggestion

fosterlynn:
>I don't know of electric grids working this way but for example mobile bandwidth can have certain amount of GB of quota assigned.

This is a good use case for when we get to "reserved" kind of thing, with a bit of a different flavor.

PR #462, opened 2019-03-01
Core attributes (was: affects > involves, affectedQuantity > observedQuantity)

fosterlynn:
working with core attributes for a minimal PR, started with 'affects' and 'affectedQuantity'

question: On the time properties, if we think they will be either an Instant or a TemporalEntity (Instant being a subclass of TemporalEntity), should we explicitly define the rdfs:range to be both of those, or leave it at TemporalEntity and let the examples speak?

### DIFF
--- a/examples/fulfill-satisfy.yaml
+++ b/examples/fulfill-satisfy.yaml
@@ -65,7 +65,7 @@
     provider: https://bob.example/
     receiver: https://manufacturing.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
-    affectedQuantity:
+    observedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
     observedTime:
@@ -91,7 +91,7 @@
     provider: https://bob.example/
     receiver: https://manufacturing.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
-    affectedQuantity:
+    observedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
     observedTime:
--- a/examples/process-manufacturing.yaml
+++ b/examples/process-manufacturing.yaml
@@ -52,8 +52,8 @@
     action: consume
     provider: https://manufacturing.example/
     receiver: https://manufacturing.example/
-    affects: mfg:3129ca8b-fcda-45be-bbda-294dc924d3b9 # plywood sheets
-    affectedQuantity:
+    involves: mfg:3129ca8b-fcda-45be-bbda-294dc924d3b9 # plywood sheets
+    observedQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 3
 
@@ -63,8 +63,8 @@
     action: use
     provider: https://manufacturing.example/
     receiver: https://manufacturing.example/
-    affects: mfg:52f0e212-3c4f-4d27-b345-5e964c135824 # CNC machine
-    affectedQuantity:
+    involves: mfg:52f0e212-3c4f-4d27-b345-5e964c135824 # CNC machine
+    observedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 3.5
 
@@ -74,8 +74,8 @@
     action: cite
     provider: https://makezine.com/
     receiver: https://manufacturing.example/
-    affects: mfg:6b97b1be-8e07-44ac-82e5-214f1b2aaf33 # desk design specification
-    affectedQuantity:
+    involves: mfg:6b97b1be-8e07-44ac-82e5-214f1b2aaf33 # desk design specification
+    observedQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
 
@@ -86,7 +86,7 @@
     provider: https://alice.example/
     receiver: https://manufacturing.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
-    affectedQuantity:
+    observedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 7
 
@@ -96,8 +96,8 @@
     action: produce
     provider: https://manufacturing.example/
     receiver: https://manufacturing.example/
-    affects: mfg:e1721a61-cd47-4556-84b9-8b1b81da15bf # desk
-    affectedQuantity:
+    involves: mfg:e1721a61-cd47-4556-84b9-8b1b81da15bf # desk
+    observedQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
 
--- a/examples/process-workflow.yaml
+++ b/examples/process-workflow.yaml
@@ -31,8 +31,8 @@
     action: accept
     provider: https://alice.example/
     receiver: https://auto-repair.example/
-    affects: alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
-    affectedQuantity:
+    involves: alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    observedQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
 
@@ -42,8 +42,8 @@
     action: improve
     provider: https://auto-repair.example/
     receiver: https://alice.example/
-    affects: alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
-    affectedQuantity:
+    involves: alice:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    observedQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -376,6 +376,13 @@ vf:intendedTime a           owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The intended temporal instant or interval." .
 
+vf:satisfiedTime a          owl:ObjectProperty ;
+        rdfs:label          "satisfied time" ;
+        rdfs:domain         vf:Satsifaction ;
+        rdfs:range          time:TemporalEntity ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "For satisfaction involving scheduling, the satisfied portion of the time." .
+
 vf:plannedTime a            owl:ObjectProperty ;
         rdfs:label          "planned time" ;
         rdfs:domain         vf:Process ;
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -292,26 +292,33 @@ vf:currentQuantity a        owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The current amount and unit of the economic resource. This can be stored or derived from economic events affecting the resource" .
 
-vf:observedQuantity a       owl:ObjectProperty ;
-        rdfs:label          "observed quantity" ;
-        rdfs:domain         vf:EconomicEvent ;
+vf:flowQuantity a           owl:ObjectProperty ;
+        rdfs:label          "flow quantity" ;
+        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent) ] ;
         rdfs:range          qudt:QuantityValue ;
         vs:term_status      "unstable" ;
-        rdfs:comment        "The amount and unit of the event. This is the quantity that could be used to increment or decrement a resource, depending on the type of resource and type of event.  It is also used in counting or accounting functions, and exchanges." .
+        rdfs:comment        "The numeric value and unit of the event, commitment, or intent." .
 
-vf:committedQuantity a      owl:ObjectProperty ;
-        rdfs:label          "committed quantity" ;
-        rdfs:domain         vf:Commitment ;
-        rdfs:range          qudt:QuantityValue ;
-        vs:term_status      "unstable" ;
-        rdfs:comment        "The amount and unit of the commitment." .
+#vf:observedQuantity a       owl:ObjectProperty ;
+#        rdfs:label          "observed quantity" ;
+#        rdfs:domain         vf:EconomicEvent ;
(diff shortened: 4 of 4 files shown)


fosterlynn:
>question: On the time properties, if we think they will be either an Instant or a TemporalEntity (Instant being a subclass of TemporalEntity), should we explicitly define the rdfs:range to be both of those, or leave it at TemporalEntity and let the examples speak?

Actually, I think I would feel more comfortable with both as the range, since we're using a property of Instant.  Does that make sense in rdf-land?

fosterlynn:
added satisfiedTime to support partial satisfaction of intended schedule

fosterlynn:
>Actually, I think I would feel more comfortable with both as the range, since we're using a property of Instant.

pushed that, for comment of course :)

fosterlynn:
>I also don't won't to eliminate at this moment possibility of creating one intent for _work_ with number of people expected to provide it.

I feel like Intents will have more quantities needed for various expression of intent, and options for expression at different levels.  I think commitments and events are more constrained.  I think people need to have their individual agency respected then.  So I don't see 50 workers on one commitment or event.  If they are a group and the group has agency, then that is still one agent making the commitment.

fosterlynn:
Another experiment - flowQuantity and flowTime.  I'd like to think of something better than involves for the resource too, but haven't thought of anything yet.

I think at this point, we need generic names and definitions that can be experimented with.  And of course we can continue to discuss.  But if we have a quantity property (say flowQuantity) with a fairly stable name, you and I can disagree on the definition and usage for now, as in this case for time-based resources.  And if we can't figure it out, it can get resolved over time in actual implementations and use.  And if we need another quantity field, say involvedQuantity or whatever, it can get easily added when someone needs it.

As soon as we agree on some naming, I can update the existing examples again.

fosterlynn:
Another idea.

Right now in flows we have:
* resourceClassifiedAs
* resourceConformsTo
* involves

I think it might be helpful to get "resource" into the vf:involves name so the above 3 properties hang together better for people.  Could be:
* involvesResource
* resourceInvolved
* hasResource
* resource

If we get the term "involves" out of the property that references the resource (last 2 options), we could use it for the main quantity field (vf:involvesQuantity), and it wouldn't be attached to the resource particularly. It would be more like the flow quantity, involved in the flow, which could affect a resource or not, definitely is be part of the process, etc.

Looking for compromise, and not sure if you don't like "flowQuantity" because of the name or the definition.  Or both, of course.  But also looking for a more explicit name for the resource reference than just "involves".

fosterlynn:
A note of explanation for why I would really like to stabilize these core attributes of all the flows.  At least the names.

Most of our current projects are reaching a point now or soon where there will be a new implementation of the vocabulary:
* holochain we are finishing a demo written for their prototype Go code, and just getting started on a version written for their new production Rust code, with several user groups interested in writing apps on top of that
* activity pub we are taking over the commonspub generic social media code development from MoodleNet (they forked it because they need to focus on their own requirements), and will be starting soon on extending it for VF to create combined social and economic networking 
* a couple UI projects, that are currently using VF api referencing the NRP/OCP code as backend, want to go p2p, probably using the activity pub mechanism - so this would be the time for them to switch to more current VF definitions
* a european group we are connected to through one of those UI projects got a grant for a project, which included VF in the proposal, and that is just starting up this week, I'm unclear on the overall architecture right now

So, the holochain and activity pub projects (plus Luandro's ssb project not mentioned above because I'm not sure of the current status and timing) will be frameworks or mechanisms in those technologies, so that version of VF used will end up being at the core and spreading in that way by default.  There will be some level of de-facto vocabulary being created.  So I'd like to get it as close as possible to something we all can be happy with or at least live with.

fosterlynn:
Another small note: I'm fine with going back to `intendedQuantity`, `committedQuantity` and `observedQuantity`.  As well as `intendedTime`, `committedTime`, `observedTime`.  I don't care a lot if we use the same names for all the ICE flows, I'm good either way.

fosterlynn:
>plus Luandro's ssb project not mentioned above because I'm not sure of the current status and timing

Hey @luandro has been moving forward on this stuff, see https://github.com/valueflows/vf-graphql/issues/8.  Very cool!  But yes, another to add to the list of projects who need us to make some decisions...

elf-pavlik:
> Another small note: I'm fine with going back to `intendedQuantity`, `committedQuantity` and `observedQuantity`.  As well as `intendedTime`, `committedTime`, `observedTime`.  I don't care a lot if we use the same names for all the ICE flows, I'm good either way.

I currently don't see need for that, the `@type` of flow makes it clear and currently having specific properties for specific types of flows sound like trying to optimize for queries while not knowing if we really need such optimization. If we do we could look at adding more specific sub-properties. 

> If we get the term "involves" out of the property that references the resource (last 2 options), we could use it for the main quantity field (`vf:involvedQuantity`), and it wouldn't be attached to the resource particularly.

Actually in `involvesQuantity` I only see beneficial that it makes it clear that it relates to quantity of a resource which flow `invovles` or only specifies its classification. 

> Could be:
> * hasResource

I recall long conversations in early days of VF where you argued that flows don't have resources. The resources stand on their own and various flows can affect / involve some quantity of those resources.


> * resource

Property name needs to capture **the meaning of the relationship** it represents. Here it just echoes one of the types of the statement's object.

fosterlynn:
>I currently don't see need for that, the @type of flow makes it clear

OK with me

>Actually in involvesQuantity I only see beneficial that it makes it clear that it relates to quantity of a resource which flow invovles or only specifies that resource classification.

So you are OK with involvesQuantity referring to a referenced resource or only a classification or specification?  Are you OK with involvesQuantity referring to a type of work / skill?  Or a service?  And anything not actually inventoried?  Not needing to have the property vf:involves be present in the flow?

Or did you mean the opposite?  involvesQuantity only is valid if there is an involved resource specified?

elf-pavlik:
> Or did you mean the opposite? involvesQuantity only is valid if there is an involved resource specified?

A specific resource referenced with `involves` or just category or specification referenced with `resourceClassifiedAs` or `resourceConformsTo`.

So in example of ouikits the number of ouikits but never the temporal length of usage.

fosterlynn:
>A specific resource referenced with involves or just category or specification referenced with resourceClassifiedAs or resourceConformsTo.

So, would this be a valid use of involvedQuantity?
```
action: work
involvedQuantity: 
      qudt:unit: unit:Hour
      qudt:numericQuantity: 10
resourceClassifiedAs: wd:Q3064143 #facilitation
```

(no involved inventoried resource)

fosterlynn:
>So, would this be a valid use of involvesQuantity?

Assuming not, how about this as two valid examples?

main quantity not derived
```
action: work
flowQuantity: 
      qudt:unit: unit:Hour
      qudt:numericQuantity: 4
resourceClassifiedAs: wd:Q3064143 #facilitation
```

main quantity derived
```
action: use
flowQuantity: 
      qudt:unit: unit:Number-per-Hour
      qudt:numericQuantity: 72
involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericQuantity: 12
temporalLength:
      qudt:unit: unit:Hour
      qudt:numericQuantity: 6
resourceClassifiedAs: wd:Q3064143 #facilitation
```

elf-pavlik:
> `resourceClassifiedAs: wd:Q3064143 #facilitation`

I don't see this one, since we have no resource involved (other than possibly considering person doing the work as 'human resource'. Since *facilitation* directly classifies the work I think in this case we can simply use `classifiedAs: wd:Q3064143 #facilitation`, so we classify the flow itself which represents the work, not some involved resource.

```yaml
action: use
flowQuantity: 
      qudt:unit: unit:Use-Hour #not sure right unit, but it is like kit-hours
      qudt:numericQuantity: 72
involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericQuantity: 12
temporalLength:
      qudt:unit: unit:Hour
      qudt:numericQuantity: 6
resourceClassifiedAs: wd:Q3064143 #facilitation
```

This one looks ok, except minor things
* not 'facilitation' but some other wikidata reference
* `flowQuantity` - i think this naming doesn't have any notion of how this quantitative value relates to this flow. it just in a way echoes type of the subject 'flow' and type of the object 'quantitative value' but nothing about the relation itself. In most cases such quantities relate to some way of social accounting for services, we could start here with `accuntingQuantity` and see if we find some name that captures the meaning of this relationship even clearer.

elf-pavlik:
> main quantity not derived

One can always just simply use the same value for `accountingQuantity` as one of the other quantities

```yaml
action: work
accountingQuantity: 
      qudt:unit: unit:Hour
      qudt:numericQuantity: 4
temporalLength: 
      qudt:unit: unit:Hour
      qudt:numericQuantity: 4
classifiedAs: wd:Q3064143 #facilitation
```

fosterlynn:
>flowQuantity - i think this naming doesn't have any notion of how this quantitative value relates to this flow. it just in a way echoes type of the subject 'flow' and type of the object 'quantitative value' but nothing about the relation itself. In most cases such quantities relate to some way of social accounting for services, we could start here with accuntingQuantity and see if we find some name that captures the meaning of this relationship even clearer.

Well, we have nothing named "flow" in the vocabulary.  It gives the idea of the quantity being the one in motion.  But I'm not in love with "flow" and am open to other ideas of course.  Except not "accountingQuantity" because many people will see that as a narrow term, like debits and credits.  It is also the quantity in and from recipes, it is the quantity that goes into a process to be part of the transformation, it is the quantity used in calculating distributions, etc.  It is anything you need a count or quantity for.  Last of all it is used for accounting, analysis, etc.  I also personally don't really care about the "rules" for naming that don't like just naming it what it is.  Sometimes that gets too fancy imo.  This is the event quantity, commitment quantity, just the quantity.  But if you can think of something that is more expressive but still general, that would be great.  

Also, I don't see why you don't like time-based units of measure as a base or fundamental quantity?  As far as I can tell, all the measurement ontologies have them.  Why is that not as valid as duration based terms?

Thinking....
actionQuantity? that may make it more clear that it relates more to the action than just affecting the resource....

elf-pavlik:
> It is also the quantity in and from recipes, it is the quantity that goes into a process to be part of the transformation

I think here `involvesQuantity` will get used since often recipe describing how to transform some kinds resources into other kinds resources. I think to clarify that we should take some examples of recipes which involve all those types of quantitative values.
* quantity of involved resources
* duration / temporal length
* service / accounting quantity - product of the two factors above and possible other factors

Tomorrow I'll write down YAML for that recipe for harvesting which we have already discussed on various occasions. Where plan can get created by scaling both quantities of involved resources as well as duration of using them.

fosterlynn:
>I think here involvesQuantity will get used since often recipe describing how to transform some kinds resources into other kinds resources.

Repeating, what does that name give you that flowQuantity (or whatever name for the one/main quantity) does not give you, in conjunction with action increment/decrement, etc.?  What is the benefit that outweighs using the same quantity field name for all events, or in this case all recipe inputs?

elf-pavlik:
With just single quantitative value we can not describe something like: *use 4 clamps for 30min* which also recipe for making some furniture might require.

fosterlynn:
Per our chat, here are some ways that come to mind for how the flow quantity field is used.
* Recipe quantities are transferred to the plan, and multiplied if the plan involves multiples of the recipe quantities.
* Event quantities are used to increment or decrement the resource if there is one and if the action calls for that.
* Event quantities are used in formulas defined in value equations to distribute income.
* When planning, committed quantities are looked at to see if there are enough already or if more need to be produced or obtained (in addition to the resource currentQuantity).
* When fulfilling a commitment, the quantity fulfilled is all or part of the commitment quantity.  Fulfilled quantities for a commitment are added up.
* When satisfying an intent, the quantity satisfied is all or part of the intent quantity.  Satisfied quantities for an intent are added up.
* An event of some quantity can be exchanged for an event of some other quantity.
* "Unit pricing" or exchange ratios can be expressed with some event quantity for some other event quantity.
* Reporting of events for analysis, accounting uses the quantities for summing within a period, comparing to previous, with various filters (including on quantity), etc.

Feel free to add to the list.  And to consider if any of these would be easier or better with different quantity fields for different actions etc.

fosterlynn:
And a list for time (instant or duration).
* Timelines / plans
* Calendar scheduling of an agent or resource
* Recipe durations (in days, hours, etc.) can be multiplied if the planned output quantity is a multiple of the recipe output quantity. Right now we have that duration on the RecipeProcess only.  It could possibly be added to the RecipeFlow if there is a real world need for that, with some information for when to schedule within the process duration.... but I suspect that could mean they need a series of processes to accommodate that level of detail.  Recipe durations are translated into real calendar durations during planning.
* Analysis or accounting reporting can be within a time period, and would use the time on events as a filter for that.
* Comparison of promised or planned time to actual time, to improve processes or for accountability

Probably are others....

fosterlynn:
Regarding time and referring to the list above for times:  Do we agree on the need for one time property in flows, that can be an instant or a duration (start/end or whatever is appropriate)?  If so, let's get a name.  And I'd like to know your preferred way to use the owl time for that.  And try to keep it as simple as possible for people to use either instant or duration.

flowTime? calendarTime? 

Or we could treat flows as inferred times if that means that it translates easily to non-rdf implementations.  Then we may not need another name.

Or we could keep what we have...

fosterlynn:
Reading the apple harvesting use case [here](https://github.com/valueflows/valueflows/pull/468), you have `involvesQuantity` and `temporalLength`.  In order to get a more complete picture, can you please do a brief example of:

a citation - what quantities are used and how are they named?  (say 1 document that contributes 20% of the process's value)

a work event that measures the work of a person in words (for translations)

Both of those are real examples from the wild.

Also, does `involvesQuantity` only work for inventoried resources?  Or would you use it for a material resource that the agent chooses not to inventory (so no resource instance exists in the software)?

elf-pavlik:
#467 already went it with some of the changes, now I also see #469 and #470 
do you still see need to keep this PR open?

fosterlynn:
Let's keep it open until we resolve core properties, just for the discussion, not the code changes, then close it.  It has open questions in the comments I'd like to keep track of.

PR #463, opened 2019-03-02
explain qty vs time

fosterlynn:
Not to merge, as I know we still disagree @elf-pavlik .  I did this as an exercise to try to coherently explain how I'm seeing time-related quantities.  If you want to write an alternative textual explanation, it might help us to define the differences.  Did this as a separate PR to specifically leave #462 as a generic naming without textual definitions, to support different directions.


fosterlynn:
I added a sentence to account for compound quantities and supporting quantity properties.  Also a sentence about scaling quantity plus time when generating a plan from a recipe.

PR #465, opened 2019-03-19
clarify resource classification and specification

fosterlynn:
(has been asked by 2 people in the last week)

### DIFF
--- a/docs/introduction/resources.md
+++ b/docs/introduction/resources.md
@@ -26,11 +26,11 @@ So, for example, you may want an apple. Or you may want a green apple. Or you ma
 
 Or, Herb is the parent classification of Anise Hyssop, Goldenrod, Nettles, Red Clover, etc.  Besides its usefulness in understanding taxonomies of resource types, this can be useful when one can define a general recipe that will work for many more specific kinds of resources.
 
-People can use the multitude of existing taxonomies for resources, or can also create their own as needed.  The references to resource classifications are uri's, and not covered inside ValueFlows.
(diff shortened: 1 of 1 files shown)


docs/introduction/resources.md
```diff
@@ -26,7 +26,11 @@ So, for example, you may want an apple. Or you may want a green apple. Or you ma
 
 Or, Herb is the parent classification of Anise Hyssop, Goldenrod, Nettles, Red Clover, etc.  Besides its usefulness in understanding taxonomies of resource types, this can be useful when one can define a general recipe that will work for many more specific kinds of resources.
 
-People can use the multitude of existing taxonomies for resources, or can also create their own as needed.
+People can use the multitude of existing taxonomies for resources, or can also create their own as needed.  The references to resource classifications are uri's, and not covered inside ValueFlows.
+
+The ResourceSpecification is part of ValueFlows, and part of the recipe structure. 
+It is used when you need to have more definition and configurability than you can 
+find in a simple taxonomy, whether there is a recipe or not, but is required in a recipe.
```
elf-pavlik: This sounds like one can not use classifications in recipe. Recipe to produce some product conforming to specific specification needs to reference that specification, still the same recipe for some or even all inputs may reference just classifications.

fosterlynn: Yes, I was mostly thinking of the output.  We'd have to see in detail if a recipe would work with just a name for a resource classification of an input, probably would if one assumes substitutability.  I'll just remove that last piece.

fosterlynn: Done, and one more clarification.

PR #466, opened 2019-03-19
service example

fosterlynn:
including resources, resource specifications, resource classification

### DIFF
--- a/examples/process-service.yaml
+++ b/examples/process-service.yaml
@@ -1,0 +1,114 @@
+# Example: Simple service process
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - alice: https://alice.example/
+    bob: https://bob.example/
+    carol: https://carol.example/
+    training: https://training.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/process-service.yaml
+'@graph':
+
+  # Economic resources and specifications
+
+  - '@id': training:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': EconomicResource
+    classifiedAs: https://www.wikidata.org/wiki/Q53953506 # 3D printer
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': training:6b97b1be-8e07-44ac-82e5-214f1b2aaf33
+    '@type': EconomicResource
+    conformsTo: training:52f0e212-3c4f-4d27-b345-5e964c135824 # 3D printing class handout
+    currentQuantity:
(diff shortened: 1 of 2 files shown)


examples/process-service.yaml
```diff
@@ -0,0 +1,114 @@
...
+    receiver: https://training.example/
+    resourceConformsTo: training:f438393-7f87-4914-806c-e23a4fd15e89 # 3D printer training
+    actionQuantity:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 8
+
+  - '@id': training:d4d2fd71-34f2-41c3-b1c5-19ad5ed2da59
+    '@type': EconomicEvent
+    outputOf: training:02b39a30-3e04-4305-9656-7f261aa63c84
+    action: provide
```
elf-pavlik: I think we can simply use *service* here, since it gets both provided and received at the same time.

fosterlynn: OK will do.  Not great, but can't think of anything better.  And I agree `deliver` isn't very good either. :(

fosterlynn: done, and added to the .ttl file

examples/process-service.yaml
```diff
@@ -0,0 +1,114 @@
...
+  # Service delivery
+
+  - '@id': training:02b39a30-3e04-4305-9656-7f261aa63c84
+    '@type': Process
+    name: Give the 3D printing workshop
+
+  - '@id': training:b90b0b77-09a2-42e2-8bd4-e9ae2c1c617
+    '@type': EconomicEvent
+    inputOf: training:02b39a30-3e04-4305-9656-7f261aa63c84
+    action: consume
```
elf-pavlik: I understand that in your scenario handouts will 'change existence' during this process. In case where they do not change existence (eg. just a manual) one could use *issue* (*give*) as well?

fosterlynn: I am thinking of the handouts (and a bunch of other stuff) being part of the service delivered - they would get passed out to each student, who would take it home.  Maybe could `give` them instead (at least once we know how to do that), but I don't see any need to do that separately really.  People will be able to do that if they want, presumably, but for me it over-complicates it.  They become just part of the overall service.

Another part of this use case not included in this example, but was part of the real use case at Sensorica, was consuming polymer for people to practice making something with the 3d printer.  Again, I would just see all of that as consumed and becoming part of the workshop service, even though each student can hopefully take home some trinket they made.  This also too unimportant to log as some production, since it is really just for training.

fosterlynn:
closing, doing updated copy of this example

PR #467, opened 2019-03-19
change affects to involves for events

fosterlynn:
I think we agree on this, so I thought I'd get that pushed.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -178,10 +178,10 @@ vf:recipeOutputOf
 vf:involves
         a                   owl:ObjectProperty ;
         rdfs:label          "involves" ;
-        rdfs:domain         [ owl:unionOf (vf:Commitment vf:Intent) ] ; 
+        rdfs:domain         [ owl:unionOf (vf:Commitment vf:Intent vf:EconomicEvent) ] ; 
         rdfs:range          vf:EconomicResource ;
         vs:term_status      "unstable" ;
-        rdfs:comment        "Economic resource the commitment commits or the intent intends to affect. Could be an actual resource or a classification / specification of resource." .
+        rdfs:comment        "Economic resource referenced by the flow." .
 
 vf:underlyingResource
         a                   owl:ObjectProperty ;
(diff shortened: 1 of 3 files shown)


release-doc-in-process/all_vf.TTL
```diff
@@ -178,10 +178,10 @@ vf:recipeOutputOf
 vf:involves
         a                   owl:ObjectProperty ;
         rdfs:label          "involves" ;
-        rdfs:domain         [ owl:unionOf (vf:Commitment vf:Intent) ] ; 
+        rdfs:domain         [ owl:unionOf (vf:Commitment vf:Intent vf:EconomicEvent) ] ; 
         rdfs:range          vf:EconomicResource ;
         vs:term_status      "unstable" ;
-        rdfs:comment        "Economic resource the commitment commits or the intent intends to affect. Could be an actual resource or a classification / specification of resource." .
+        rdfs:comment        "Economic resource referenced by the flow." .
```
elf-pavlik: What do you think about: *Economic resource involved in the flow*

release-doc-in-process/all_vf.TTL
```diff
@@ -198,13 +198,13 @@ vf:relationship  a          owl:ObjectProperty ;
...
-        rdfs:range          vf:EconomicResource ;
-        vs:term_status      "unstable" ;
-        rdfs:comment        "The economic resource that is decremented or incremented by the economic event, or just what the economic event is about." .
+#vf:affects
+#        a                   owl:ObjectProperty ;
+#        rdfs:label          "affects" ;
+#        rdfs:domain         vf:EconomicEvent ;
+#        rdfs:range          vf:EconomicResource ;
+#        vs:term_status      "unstable" ;
+#        rdfs:comment        "The economic resource that is decremented or incremented by the economic event, or just what the economic event is about." .
```
elf-pavlik: Let's just delete it, we can always find it in git history so we have 'undo' button there if needed.

release-doc-in-process/all_vf.TTL
fosterlynn: I try not to repeat the word, but no problem with me - changed.

release-doc-in-process/all_vf.TTL
fosterlynn: done

PR #468, opened 2019-03-20
initial harvest recipe with process example

elf-pavlik:
First take on harvesting recipe, it scales
* input trees, people and resources by factor of 5
* temporal lengths by factor of 6
* produced apples by factor 5 * 6 = 30

At this moment we have not defined how to describe how things should scale.

### DIFF
--- a/examples/harvest-recipe.yaml
+++ b/examples/harvest-recipe.yaml
@@ -1,0 +1,140 @@
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - bumblebee: https://bumblebee.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/harvest-recipe.yaml
+'@graph':
+
+  - '@id': bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    '@type': RecipeProcess
+    skos:note: two people harvesting an apple tree in 15 min
+
+  - '@id': bumblebee:fac5022d-463c-4a07-a601-8f54d12293bc
+    '@type': RecipeFlow
+    skos:note: tree bearing apples
+    action: consume
+    recipeInputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': bumblebee:8ffdef1c-2825-4ae1-9127-24cdacf09fde
+    '@type': RecipeFlow
+    skos:note: harvested apple tree
+    action: produce
+    recipeOutputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': bumblebee:efdf5c80-1a31-45bc-bb14-0e3746d1a81a
+    '@type': RecipeFlow
+    skos:note: apples
+    action: produce
+    recipeOutputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 20
+
+  - '@id': bumblebee:8e664596-30df-428e-b6fd-a784c2acf38b
+    '@type': RecipeFlow
+    skos:note: willow basket
+    action: use
+    recipeInputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 2
+    temporalLength:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 0.25
+
+  - '@id': bumblebee:76694411-781c-41b9-a265-5fe2ee456d79
+    '@type': RecipeFlow
+    skos:note: ladder
+    action: use
+    recipeInputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    temporalLength:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 0.25
+
+  - '@id': bumblebee:1178138b-21dd-42e1-8f43-c8462d14eaab
+    '@type': RecipeFlow
+    skos:note: person harvesting
+    action: work
+    recipeInputOf: bumblebee:53b44c35-d514-4945-b257-9f66c7ec0d67
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 2
+    temporalLength:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 0.25
+
+  - '@id': bumblebee:11a2caee-a44d-4099-aa1b-56504684127f
+    '@type': Process
+    skos:note: ten people harvesting 30 apple trees in 3 hours
+
+  - '@id': bumblebee:60ac2d7c-e5cd-4de4-b981-0ad64ea716a2
+    '@type': Intent
+    skos:note: tree bearing apples
+    action: consume
+    inputOf: bumblebee:11a2caee-a44d-4099-aa1b-56504684127f
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 30
+
+  - '@id': bumblebee:0cdd34f8-e688-4d96-934e-b12b27d2c51b
+    '@type': Intent
+    skos:note: harvested apple tree
+    action: produce
+    outputOf: bumblebee:11a2caee-a44d-4099-aa1b-56504684127f
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 30
+
+  - '@id': bumblebee:b32fdc8f-dc39-4b9e-92e9-991aa5bac2ec
+    '@type': Intent
+    skos:note: apples
+    action: produce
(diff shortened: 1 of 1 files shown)


examples/harvest-recipe.yaml
```diff
@@ -0,0 +1,140 @@
...
+
+  - '@id': bumblebee:11fdbf89-90d4-47c8-a59d-37fdf2c3fb47
+    '@type': Intent
+    skos:note: person harvesting
+    action: work
+    inputOf: bumblebee:11a2caee-a44d-4099-aa1b-56504684127f
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 10
+    temporalLength:
```
fosterlynn: Does this mean that the range of involves would be expanded to include Agent? Or would there be some kind of "work availability" or "labor power" kind of EconomicResource? Or neither, just using that property for number of people not related to resources?

elf-pavlik: Currently I don't see need for defining `Person` and `Resource` as disjoint. If we can in some situations consider dog, cat or a cow as a `Resource` I don't see why sometimes we can't act as both `Resource` and `Agent`. It may also come into play when modeling passenger transportation where process would *load* me in one place and *unload* me in another place, which differs from providing *load* of something unknown / undefined for me and providing *unload* of something unknown / undefined for me.

elf-pavlik: As for the range we don't have to 'extend' it at all. All it does it implies that referenced thing also has this class. So one can use it with anything one doesn't problem with getting class `vf:EconomicResource` added by RDFS reasoner.

fosterlynn: One concern I have about this is that non-rdf implementations can't do any of that. So we would have to figure out how to translate that. Besides the concern I already expressed about "human resources" in corporate capitalism.... Not necessarily totally opposed, but seeing difficulties. And would need to see it carried through to commitments. But we don't need to do that now, since that is not the purpose of your use case....

fosterlynn:
One thought about Intents.  I think Intents will need more quantities than commitments and events, where things are getting tightened up, and are basically at an operational level.

So I think it makes sense to request 10 people for 3 hours to pick apples, especially where the rest of the plan won't be visible.  And probably we will need more than that, like availableQuantity for offers.  And perhaps others.

In this example, once we get to commitments and events, I don't so much see the need.  Seems like the process will be defined as taking 3 hours.  Each event will have the commttedTime and observedTime for seeing the duration as relates scheduling and recording the facts in terms of time spent.

What I would intuitively record on the commitments and events is:
work: number of hours (assuming people are being paid by the hour) with one commitment for each person, and 1 (or more as needed) event for each person
baskets: number of baskets
ladders: number of ladders

This assumes the baskets and ladders are resources of the orchard.  (If they are renting them or whatever, then the quantity would reflect how that is done by the rental agent.  And probably wouldn't be part of the process in terms of actual use.  The actual use would still probably be just number of ladders for the process.)

Sorry, this is how I see it.  Not sure how to resolve.  Still willing to have a second and third quantity, as long as we can have one first quantity that can be used across the board.

I wonder if it would help to also show the time on these, since they will make that part clear?  And to try out a commitment or two coming from the intents?

fosterlynn:
The more examples I look at, the more I become convinced that in general for usage, people will log time if the resource is individually identified, expensive, require maintenance, those kinds of things; and the event will be for just one of the resource.  People will log physical measurements like count or weight or volume if the resource is not individually identified, relatively cheap, doesn't need individual attention paid to it, etc.; and the time won't matter so much except for scheduling. 

And there are probably examples where people will log using a compound unit, which I am not familiar with at the operational level.  (I have seen it at the analytical level, like the Nova Scotia fisheries analysis where the resource flows are tons per year of a certain species of fish caught with a certain type of gear.)

As an exercise to be more concrete, looking at the list [here](https://github.com/valueflows/valueflows/pull/462#issuecomment-474604363) of uses of quantities, and checking if everything on these usage examples will work with these:

> Recipe quantities are transferred to the plan, and multiplied if the plan involves multiples of the recipe quantities.

Quantity of ladders is multiplied for the plan quantity.  Durations affect the scheduled times.

>Event quantities are used to increment or decrement the resource if there is one and if the action calls for that.

Use doesn't increment or decrement a material resource quantity.  Does change availability on the schedule.

>Event quantities are used in formulas defined in value equations to distribute income.

For usage, I think only resources specifically contributed by someone would distribute income, and then the resources would have to be specifically identified, even if it is a stock of more than one ladder - triggering more than one commitment and event.

> When planning, committed quantities are looked at to see if there are enough already or if more need to be produced or obtained (in addition to the resource currentQuantity).

For usage, seems like this is the quantity of ladders related to schedule and availability.

>When fulfilling a commitment, the quantity fulfilled is all or part of the commitment quantity. Fulfilled quantities for a commitment are added up.

For ladders, this seems basically like the number of ladders within the scheduled time.  Would there ever be a need to swap in ladders from the same stock during the scheduled time?  Doesn't seem like it... but if there were, then it wouldn't affect the 10 ladders for some duration of time?  If there were a need to swap in ladders from another stock for part of the time, that would be a separate event.

>When satisfying an intent, the quantity satisfied is all or part of the intent quantity. Satisfied quantities for an intent are added up.

Same as fulfillment I think.  If there were multiple commitments, that would be fine.

>An event of some quantity can be exchanged for an event of some other quantity.

Since the whole event is part of the exchange, this seems fine.  (Seems fine no matter how we do it.)

>"Unit pricing" or exchange ratios can be expressed with some action quantity for some other action quantity.

This one seems like it would involve both the number of ladders and the time used.  Might look like "ladder hours" kind of thing.  Seems like there would need to be a compound quantity of some sort.  It seems less likely to apply to usage than possession/availability, but if it did apply to usage, then seems like the usage would need to be recorded with the compound quantity to match the pricing.  (Other side of the exchange doesn't matter if it is currency or something else as far as I can tell.)

>Reporting of events for analysis, accounting uses the quantities for summing within a period, comparing to previous, with various filters (including on quantity), etc.

This one seems like a compound quantity would need to be created from the number of ladders and the time as a quantity (like your `temporalLength`) to be a meaningful number, for standard reports that show one number across types of resources.  Also you might also want more individual comparisons, like average length of usage, average number of ladders used, etc.  So this one would benefit from having separate quantities on the event.  Seems like that would work as supporting quantities, or as the way you have done it.

elf-pavlik:
> The more examples I look at, the more I become convinced that in general for usage, people will log time if the resource is individually identified, expensive, require maintenance, those kinds of things; and the event will be for just one of the resource. People will log physical measurements like count or weight or volume if the resource is not individually identified, relatively cheap, doesn't need individual attention paid to it, etc.; and the time won't matter so much except for scheduling.
And there are probably examples where people will log using a compound unit, which I am not familiar with at the operational level.

I need to do this clear breakdown of requirements and what quantitative values each requirement depends on. 

> So this one would benefit from having separate quantities on the event. Seems like that would work as supporting quantities, or as the way you have done it.

I think we need to make sure, that understanding what quantitative value different requirements depend on, people can make a clear choice which values they will record and which they will omit.

I don't think we can make any of the requirements mutually exclusive, people should not find themselves in position that they need to choose one requirement or the other.

fosterlynn:
>I think we need to make sure, that understanding what quantitative value different requirements depend on, people can make a clear choice which values they will record and which they will omit.

>I don't think we can make any of the requirements mutually exclusive, people should not find themselves in position that they need to choose one requirement or the other.

Not sure I understand.  Example?

fosterlynn:
>I need to do this clear breakdown of requirements and what quantitative values each requirement depends on.

That would be very helpful.

elf-pavlik:
#471

fosterlynn:
>First take on harvesting recipe, it scales
>   *  input trees, people and resources by factor of 5
>    * temporal lengths by factor of 6
>    * produced apples by factor 5 * 6 = 30

From discussion with @bhaugen yesterday:  Planning either works as demand-driven or supply-driven.  Agriculture tends to be dominantly supply driven (once things are planted), but at some point there is a demand-driven part that kicks in.

I suspect most orchards just plan to pick what is ripe within that window of time, maybe spreading it out a bit for their own process management, or for demand, or not.  Then they trade it while it lasts.  And they can plan ahead for that in terms of harvesters for the season.

Or if you wanted in this use case, someone could have ordered some apples, and harvesters pick for that demand.

The planning algorithm can work demand-driven working backwards or supply-driven working forwards.  Bob will write up some much more detailed use cases from his experience.

elf-pavlik:
I would see this recipe useful in scenario where agent taking care of an orchard wants to plan:
* we need to harvest 30 trees

Then application used for planning could take that '30 trees constant' and based on recipe assist person using it to choose specific ratio between people and equipment involved and time duration. In my example I used 5 x 6 but they should get the same result using 10 x 3 or 3 x 10

elf-pavlik:
Variant of this UC mentioned in https://github.com/valueflows/valueflows/pull/471#issuecomment-478383324

> threading (or orchard, @elf-pavlik pick) >1 person, >1 equipment, demand-driven? 

While in harvesting UC we start with little more 'supply driven':

* we need to harvest 30 trees with ripe apples

the threading UC can start in more 'demand driven' way:

* we need to produce 1000 threaded pieces this week

Other info in that UC

* current quantity of not yet threaded pieces: 2500
* current quantity of drill presses used for threading 4 https://www.wikidata.org/wiki/Q2211295
* current quantity of threading adapters for drill presses 6 
* current quantity of threading bits (specific size) 20 (some new some already used)
* current quantity of people working full time in the workshop 5 (not every day all available to work on this specific order)
* based on past analysis - average rate at which a person can thread pieces with this equipment: 40 pieces an hour

fosterlynn:

PR #469, opened 2019-03-24
affectedQuantity > observedQuantity

fosterlynn:
Changed this, as we agreed at the end of the "cleanup" milestone.  I'd like to have one name for the (main) quantity in I-C-E, as we agreed more recently, but I think it will take too long to resolve even just the name.  So I'd like to have something I think is OK for now, while we continue to discuss, since it might get embedded in several systems.

### DIFF
--- a/examples/fulfill-satisfy.yaml
+++ b/examples/fulfill-satisfy.yaml
@@ -65,7 +65,7 @@
     provider: https://bob.example/
     receiver: https://manufacturing.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
-    affectedQuantity:
+    observedQuantity:
       qudt:unit: unit:Hour
       qudt:numericValue: 4
     observedTime:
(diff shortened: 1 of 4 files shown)


fosterlynn:
>-1 I think we can change it to involvesQuantity following #467

>If people start using it for temporal length it possibly could change. Either way I think we definitely needs something that describes amount of the resource if flow involves a resource (or just specifies resourceClassifiedAs or resourceConformsTo

Do you agree that it can be used for any quantity?  Including work hours?  Including services?  What if people do start using it for what you are using temporalLength for (like use hours)?  I need something that will make sense for all of those while we discuss.

I don't like the name, it implies it needs to be related to an inventoried resource referenced by `involves`.  Involves doesn't reference classifications or specifications. And that is like where we disagreed and didn't know it for quite some time with `affectedQuantity`.

I need something to use in the meantime that can be a clear name for any quantity.  While we continue to discuss.  And we did have agreement on observedQuantity.  I need something that will change only when we do finally come to agreement, if we can do that.

I'm also good with picking one name for I-C-E, like flowQuantity, actionQuantity, or a better name that gives more of a verb in rdf style, if we can decide in the next couple days.

fosterlynn:
Closing in favor of new PR for quantifiedAs.

Issue #464, opened 2019-03-05
refining plans (including allocation step)


elf-pavlik:
I don't think we have looked into refining plans. Both independent plans which might only include Intents as well as shared plans which require Commitments. I think common example might come from having a specific allocation step.

For example agent can make commitment to provide usage of 3 cargo bikes and make commitment based on
```yaml
'@type': Commitment
resourceClassifiedAs: wd:Q573863 # cargo bike
involvesQuantity:
  qudt:unit: unit:Number
  qudt:numericValue: 3
```

Later agent might want to schedule specific bikes, and supersede the previous commitment with 3 new ones:

```yaml
'@type': Commitment
involves: ex:96af8b82-cde2-4148-bd63-18419921a00f # specific cargo bike
involvesQuantity:
  qudt:unit: unit:Number
  qudt:numericValue: 1

'@type': Commitment
involves: ex:83b2d80e-9a66-4d31-b05a-36bfb0775922 # specific cargo bike
involvesQuantity:
  qudt:unit: unit:Number
  qudt:numericValue: 1

'@type': Commitment
involves: ex:b4ab48b7-fe1f-4c9e-b67d-048315bc35fc # specific cargo bike
involvesQuantity:
  qudt:unit: unit:Number
  qudt:numericValue: 1
```

Similar someone may want to commit to work 20 hours in specific month, later refining planning and creating more specific commitments to work at specific times.

I think composition and *includedIn* could address that, where more specific (refined) commitments stay *includedIn* the less specific ones.

fosterlynn:
Can we do this by changing the commitments?  I think we need to move towards sort of functional style, some kind of "activity" creating, updating, deleting the VF entities, event flows.  So then that kind of history is kept, but there is only one current plan.

In our ActivityPub experiments, we think we can just use existing types of Activity, and reference the VF entities using as:object in the activity. Warning: Not implemented yet!

Maybe in VF doc, we can just recommend people use AS vocabulary for this.... but maybe this is OT, don't know....

elf-pavlik:
I see it more like having events fulfilling commitments or commitments fulfilling intents. We don't have situation when we just update single Commitment but as in example above based on one less specific Commitment (eg. no *involves*) we create multiple more specific Commitments. Similar some 'monthly quota' Commitment would result in adding more specific scheduled Commitments throughout the month, in a variant of your FC use case, Alice would commit to work 20h in next month and then make more specific commitments to for example chair specific scheduled teleconference.

[...]

Issue #456, opened 2019-02-12
One EconomicEvent quantity field for counting, accounting, and base for valuation


bhaugen:
It's difficult for me to understand all the backs and forths in issues like #446 , #452 , and #455 so I will cut to the chase.

I would like to have one and only one quantity field in EconomicEvents for the purposes stated in the issue title. Those will often be used in large aggregate computations where having possible exceptions for each item on which field to use will be slower and more error-prone and in general cause programmers to curse those who made those design choices.

The field has had different names: affectedQuantity, observedQuantity, ? I don't care what it is called, as long as it is one field and consistently named. People can add other quantity fields for other purposes if they want.

bhaugen:
Oh, and y'all can call them properties if you want, too. I'm easy. I just want one and only one of them.

elf-pavlik:
@bhaugen one of the use cases we'll look at *OuiKit*:
* https://github.com/valueflows/valueflows/blob/b63f16d8cf0d9fd920aa31ab7daf62fb2166233a/use-cases/ouikit.md
* https://www.matelo-evenements.org/membre/67

People can borrow/rent cups, bowls, cutlery for events.

If we do one *use* event it needs two dimensions:
* count of items
* duration of usage

If we have *check-in* and *check-out* events. They could have just the count of items but to calculate agreement one has two consider them as a pair and based on two events calculate duration of usage (borrowing or renting).

In the end to do *valuation* you need both dimensions *count of items* + *duration*, no matter if you capture it as single flow (event, commitment, intent) or split it in two and have to always consider a pair and calculate duration based on data from both flows in a pair.

bhaugen:
I would do that as a resource which represents a collection of items and a quantity representing the duration of usage of that collection of items.

bhaugen:
Lynn says I might not have understood the use case correctly. Regardless, I think I would check in and out a resource that represents a collection. Possibly more later.

bhaugen:
> In the end to do _valuation_ you need both dimensions _count of items_ + _duration_, no matter if you capture it as single flow (event, commitment, intent) or split it in two and have to always consider a pair and calculate duration based on data from both flows in that pair.

Valuation in our experience is often a complex calculation, but then we put the computed value on the event, so after the value of the event has been computed, it's now one field again. Don't know if VF wants to add a value field to EconomicEvent or not, but in implementations, I would do that.

elf-pavlik:
[...]

Issue #449, opened 2019-02-06
Citations - how to record quantity?


fosterlynn:
Ref. #443 

>>>Interestingly, `cite` has a similar aspect to `use` and `work` where the quantity that matters is not the same as the quantity of the cited resource (document, design, etc.). And I can't think of a situation where a citation could be made for more than one of a "stock" resource. The quantity that matters still could be '1' in that citations (the events) are counted that way. For Sensorica's distribution logic, they wanted to have the person doing the citing record a percentage of the process/output that they felt the cited resource contributed to that process, then the contributor(s) to the cited resource would get that percentage of income assigned to that process . It is different in that citations don't affect availability. But we could consider `cite` in terms of figuring out what quantity properties are used for what.

>> For Sensorica's distribution logic, they wanted to have the person doing the citing record a percentage of the process/output that they felt the cited resource contributed to that process, then the contributor(s) to the cited resource would get that percentage of income assigned to that process .

>I think we should deal with this use case as part of _agreements_ milestone. It sounds to me like something related to _Reciprocity_ removed in #430 I hope we can agree that we should not put any reciprocity related responsibility directly on flows and that Sensorica experiment of using quantity in _cite_ event for that reason will have different way to model it in VF.

It is no more related to the removed vf:Reciprocity than recording work or usage.

I agree with you that we should not be recording agreements in events.  But somewhere you need to record what the receiver thinks the use value was of the cited resource for their process (if you are Sensorica), since it is specific to that citation, not an agreement or general rule.  What that means in terms of reciprocity is recorded in the agreement, agree.  There are probably other creative ways to record citations' use value too, but this was the best they and we came up with.

elf-pavlik:
Could you document here that Sensorica use case in detail and how NRP handles it?

fosterlynn:
Could you document here that Sensorica use case in detail and how NRP handles it?

https://github.com/valueflows/valueflows/wiki/Use-Cases-and-Requirements#citations-and-value-flows

Sensorica records the percentage number in the numeric value field, and the unit as Percent.  When the trace of the value flow happens during calculation of distributions of income, if NRP encounters a citation event with a Percent unit, it allocates to the citation that percent of the amount that the process has been allocated.

fosterlynn:
Yaml example to clarify.  Not making a PR, wanting to keep the discussion all here until we come to some understanding.

```
# Example: Citation

'@context':
  - https://git.io/vf-examples-jsonld-context
  - bob: https://bob.example/
[...]

Issue #375, opened 2016-08-15
timeline based stocks (availability?) for Usage and Services 


elf-pavlik:
in https://github.com/valueflows/valueflows/issues/148#issuecomment-239818476 @bhaugen writes: 

> The [Quick Response Engine](https://en.wikipedia.org/wiki/Quick_Response_Engine) (QRE)  was a skunkworks project for an ERP software company I worked for in the 1990's, and was also my first implementation of the set of ideas that fed into VF.  
> 
> We did shift scheduling for people and equipment in that project. The inventory of such time-based resources was a shift calendar composed of 15-minute slots. It was implemented as sparse array composed only of slots where the resource was scheduled to be available: for a person, the shifts they were scheduled to work. Each time-based resource had a shift calendar. A process could reserve as many of these 15-minute slots as it needed, assuming they were not already taken by some other process.

For _work_ and _use_ #29  type events we need a way to describe a resource/stock #41. I would consider describing timeline of availability as stock in which we can create 'logical resources' which we can reference in allocations #39

elf-pavlik:
Looking at https://valueflo.ws/introduction/flows.html#timeline-plans-and-observations

![vf-timeline](https://raw.githubusercontent.com/valueflows/valueflows/master/assets/flows.png)

I don't really see how 'time based resources' would fit there. Intents and commitments seem capable of representing planned (intended) availability.

![flows schedule](https://user-images.githubusercontent.com/876431/54950304-9b92d480-4f06-11e9-87bb-9f5a46237817.png)

*(on both diagrams blue intent, orange commitments)*

For personal schedule one could have similar diagram just use quantity 1 (just me).

Reading https://valueflo.ws/introduction/resources.html#the-difference-between-a-resource-and-its-specification-or-classification

> An economic resource is observable. Its specification or classification defines what type of thing the economic resource is.

So far any 'future resource' to my understanding has to have `vf:currentQuantity: 0` which actually one can observe. While I can see how we can express future plans with intents (independent plans) and commitments (agreed plans), at this moment I don't see need for instances of `time based` `vf:EconomicResource`.

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

> https://lab.allmende.io/valueflows/valueflows/-/issues/375.

If you have not done so, you are very welcome to register at https://lab.allmende.io and join the ValueFlows organization there.

PR #471, opened 2019-03-25
requirements for quantitative values

elf-pavlik:
Initial draft, needs more iterations

### DIFF
--- a/docs/introduction/requirements.md
+++ b/docs/introduction/requirements.md
@@ -2,11 +2,11 @@
 
 ## Quantitative values
 
-### Real time tracking of resources
+### Real time tracking of resources, and planning
 
-*present*
+*present*, *future*
 
-It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s).
+It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s). Similar for planning at any time in the future projected `vf:currentQuantity` and *available quantity* might differ based on existing commitments.
 
 Use cases:
 
--- a/docs/introduction/requirements.md
+++ b/docs/introduction/requirements.md
@@ -8,17 +8,36 @@
 
 It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s).
 
+Use cases:
+
+* Ouikit
+* Transportation
+
 ### Planning schedules
 
 *future*
 
 It mostly relate to flows with *use* and *work* action, but it may also come into play for other actions. For example *load* and *unload* might require very specific scheduling. Last but not least it might play role in *service quotas*, for example grid utilities providers might need to play different volumes of *consume* in different time periods. For *consume* and *produce* precise scheduling might play greater role for resources with limited buffering. Mentioned grid utilities or perishable products give two examples. Also [*just in time manufacturing*](https://en.wikipedia.org/wiki/Just-in-time_manufacturing) would have stricter scheduling requirements.
 
+Use cases:
+
+* Ouikit
+* Harvesting
+* Metered electricity
+* Shared studios
+
 ### Social accounting
 
 *past*, *future*
 
 It plays role when agents make agreements which involve reciprocity. For *use* it might combine quantity of used resource and duration of the usage. For *service* it might depend on various factors and specific Terms of Service. In some cases it might involve quantities like *number of words*, *distance* etc.
+
+Use cases:
+
+* Ouikit
+* Translation service
+* Transportation (taxi)
+* Accommodation (hoter, BnB)
 
 ### Reports and analysis
 
--- a/docs/introduction/requirements.md
+++ b/docs/introduction/requirements.md
@@ -1,0 +1,27 @@
+# Requirements
+
+## Quantitative values
+
+### Real time tracking of resources
+
+*present*
+
+It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s).
+
+### Planning schedules
+
+*future*
+
+It mostly relate to flows with *use* and *work* action, but it may also come into play for other actions. For example *load* and *unload* might require very specific scheduling. Last but not least it might play role in *service quotas*, for example grid utilities providers might need to play different volumes of *consume* in different time periods. For *consume* and *produce* precise scheduling might play greater role for resources with limited buffering. Mentioned grid utilities or perishable products give two examples. Also [*just in time manufacturing*](https://en.wikipedia.org/wiki/Just-in-time_manufacturing) would have stricter scheduling requirements.
+
+### Social accounting
+
+*past*, *future*
+
+It plays role when agents make agreements which involve reciprocity. For *use* it might combine quantity of used resource and duration of the usage. For *service* it might depend on various factors and specific Terms of Service. In some cases it might involve quantities like *number of words*, *distance* etc.
+
+### Reports and analysis
+
+*past*
+
+Agents might want to track various metrics to analyze trends, efficiency etc. In some cases it might involve quantities like *number of words*, *distance* etc.
--- a/docs/introduction/requirements.md
+++ b/docs/introduction/requirements.md
@@ -8,16 +8,21 @@
 
 It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s). Similar for planning at any time in the future projected `vf:currentQuantity` and *available quantity* might differ based on existing commitments.
 
+Recipes should allow to specify in scalable way both the quantity of the resource as well as duration of action in case of *use*.
+
 Use cases:
 
 * Ouikit
 * Transportation
+* Harvesting
 
 ### Planning schedules
 
 *future*
 
 It mostly relate to flows with *use* and *work* action, but it may also come into play for other actions. For example *load* and *unload* might require very specific scheduling. Last but not least it might play role in *service quotas*, for example grid utilities providers might need to play different volumes of *consume* in different time periods. For *consume* and *produce* precise scheduling might play greater role for resources with limited buffering. Mentioned grid utilities or perishable products give two examples. Also [*just in time manufacturing*](https://en.wikipedia.org/wiki/Just-in-time_manufacturing) would have stricter scheduling requirements.
+
+Recipes should allow to specify durations in scalable way whenever plan can get created by scaling recipe on temporal axis.
 
 Use cases:
 
@@ -31,6 +36,8 @@ Use cases:
 *past*, *future*
 
 It plays role when agents make agreements which involve reciprocity. For *use* it might combine quantity of used resource and duration of the usage. For *service* it might depend on various factors and specific Terms of Service. In some cases it might involve quantities like *number of words*, *distance* etc.
+
+While VF still doesn't clearly define Proposals, one of options to define them may involve *flow templates* which should allow to specify quantitative values used for accounting in scalable way.
 
 Use cases:
 
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -27,6 +27,7 @@
 * [Recipes](introduction/recipes.md)
 * [Conversation for Action](introduction/cfa.md)
 * [Accounting](introduction/accounting.md)
+* [Requirements](introduction/requirements.md)
 
 ## Specification
 


elf-pavlik:
Properties we discussed so far which I see play role for each requirement

* Real time tracking of resources
  * `currentQuantity`
  * `involvesQuantity`
  * `hasBegining`
  * `hasEnd`
  * `time:inXSDDateTimeStamp
* Planning schedules
  * `currentQuantity`
  * `involvesQuantity`
  * `hasBegining`
  * `hasEnd`
  * `time:inXSDDateTimeStamp
  * `temporalLength` (broader than `time:hasDuration`)
* Social accounting
  * `accountingQuantity` as requested in #456 
  * `perceivedRelativeAddedValueFraction` recently suggested in https://github.com/valueflows/valueflows/issues/449#issuecomment-476318288
* Reports and analysis
  * all above and more

elf-pavlik:
This little experiment I did to visualize planned schedules based on intents and commitments only relies on:
* `involvesQuantity`
* `hasBeginning`
* `hasEnd`

vf-schedules-vis

I'll do one more iteration on it to get it to the point where one can point it to YAML file which has intents and commitments and it will create schedules visualization for them.

docs/introduction/requirements.md
```diff
@@ -0,0 +1,46 @@
+# Requirements
+
+## Quantitative values
+
+### Real time tracking of resources
+
+*present*
+
+It involves `vf:currentQuantity` as well as *available quantity*. They might differ for example when ongoing *use* event(s) involves certain quantity of resource. *Available quantity* would equal `vf:currentQuantity` minus all the quantities currently involved in *use* event(s).
```
fosterlynn: Also possibly accept-improve processes, transportation processes... maybe others?

docs/introduction/requirements.md
```diff
@@ -0,0 +1,46 @@
...
+* Ouikit
+* Harvesting
+* Metered electricity
+* Shared studios
+
+### Social accounting
+
+*past*, *future*
+
+It plays role when agents make agreements which involve reciprocity. For *use* it might combine quantity of used resource and duration of the usage. For *service* it might depend on various factors and specific Terms of Service. In some cases it might involve quantities like *number of words*, *distance* etc.
```
fosterlynn: The people we know who wanted to count words for translations wanted to do it on a `work` event.

docs/introduction/requirements.md
elf-pavlik: https://valueflo.ws/introduction/flows.html#actions
* *load* and *unload* already decrement and increment they will get reflected on `currentQuantity`
* *accept* and *improve* at this moment don't increment or increment which to stay honest seems inconsistent to me

All four above differ from *use* in respect that acting as pair one can decrement and other one can increment the resource. Since we have only one flow with *use* it can't both increment and decrement the resource.

docs/introduction/requirements.md
elf-pavlik: To clarify that specific use case, do they count words from the original document or words in the translated document?

elf-pavlik: I guess in that translations UC to meet *Social accounting* requirement people doing translation work would use either *word count* or *work hours* for *accountingQuantity*. On the other hand to meet *Reports and analysis* use case they might need both factors, for example to have some averages of words per hours for different languages and domains, based on which they could do time estimates and plan their working schedules.

docs/introduction/requirements.md
fosterlynn: I see what you are saying with load/unload.

accept / improve were created for moving the same resource through one or more stages.  Originally they were for translating, which involved initial translation, proofreading, editing, etc. of a document - I forget the exact stages and sequence.  In any case, it is the same resource in and out, which requires an additional piece of identification to not create cycles - we used process type.  The other use case we had was Sensorica QT (quality testing) of a part that had been produced.

docs/introduction/requirements.md
fosterlynn: I don't know, could ask if it is important.  I'm sure there is an industry standard for that.  I don't tend to think it matters, because this is words on the flow.

elf-pavlik: I moved some citations specific conversation to related issue https://github.com/valueflows/valueflows/issues/449#issuecomment-476400049

fosterlynn: >On the other hand to meet Reports and analysis use case they might need both factors, for example to have some averages of words per hours for different languages and domains, based on which they could do time estimates and plan their working schedules.

Nobody in our experience has needed both.  We could run into it in the future of course, but I'd prefer to stick with known use cases as much as possible.

elf-pavlik:
https://github.com/valueflows/valueflows/pull/458#issuecomment-476253960

>> At this moment we don't even have any example of how one would know vf:currentQuantity of the resource at that specific time in the past. For commitments, I don't think we even have concept of future vf:currentQuantity that resource might (or might not) have at some point in the future.

> I'm not sure this applies to what you are talking about, but in NRP when making plans from a recipe, the logic does calculate when planned resources will be available and when not, based on current quantity and committed quantities, at least for consumed resources, not sure about used resources, but that would make sense too. (Bob wrote it, so he could supply details if needed.) The "future" quantity wouldn't show up in the vocabulary, just used in the calculations.

Based on comments above I added planning aspect to *tracking of resources* requirement.

fosterlynn:
Another piece of requirements for quantitative values is where do the units get defined?  For inventoried items, it is clear because the flow units need to match the EconomicResource.currentQuantity.unit.  The unit can also be defined in the ResourceSpecification, for non-inventoried items or non-items.

But this doesn't cover that items that are both produced and used, for example, need different units.  Or produced and cited.  Seems like we ought to handle the definition of allowed units for usage and citation as part of these requirements.  There are probably other examples in our broadening transfer discussions.  Since we are leaning away from having separate resources to handle all of that, we need a place(s) in the current model.

elf-pavlik:
Could you please clarify what do you mean by 'units get defined'?

Each instance of `qudt:QuantityValue` has explicit unit and numeric value. So one can use whatever units one wants on any flow. When flow involves resource and either increments it or decrements it, the unit should match the unit used for `vf:currentQuantity`, otherwise application may not know how to do the decrement or increment.

For flows that don't involve a resource but only reference classification, one has to choose unit every time unless a `RecipeFlow` gets used which has units already specified. `ResourceSecification` seems mostly differ from classifications that it can also include unit, so flows referencing it with `resourceConformsTo` could by default use the same unit.

fosterlynn:
>So one can use whatever units one wants on any flow. When flow involves resource and either increments it or decrements it, the unit should match the unit used for vf:currentQuantity

> flows referencing it with resourceConformsTo could by default use the same unit.

Yes exactly.  Except that without using underlyingResource, we need to define different units for different quantities (your view) or different units for the quantity matching the action (my view).  In either case, we need that on both resource and specification in some way.  

Just seems like part of the discussion, because it will be easier or harder, depending on our other choices, so if we detail that out along with the quantities, we'll know.

elf-pavlik:
> or different units for the quantity matching the action (my view)

in cases like ouikit action *use* and *reserve* for planing schedules needs
* `involvesQuantity` - number of ouikits
* beginning and end (or sometimes just duration of reservation + topological temporal relations in earlier stage)

and for social accounting needs for action like *reserve*
* `involvesQuantity` - number of ouikits
* duration of the reservation

so even for single action we can't find just one quantitative value

fosterlynn:
>in cases like ouikit action use and reserve for planing schedules needs

>    involvesQuantity - number of ouikits
>    beginning and end (or sometimes just duration of reservation + topological temporal relations in earlier stage)

So this is one quantity and also time (however we structure the time)

>and for social accounting needs for action like reserve

>    involvesQuantity - number of ouikits
>    duration of the reservation

So this is for exchange, right?  How would you say that OuiShare would express it's "price" for renting the ouikits?  Say if I want to rent some for my event?  How can you do it with 2 quantities?

fosterlynn:
>So this is for exchange, right? How would you say that OuiShare would express it's "price" for renting the ouikits? Say if I want to rent some for my event? How can you do it with 2 quantities?

Actually you could do it that way, just the "ratio" or "unit price" would need to express both together, maybe something like kit-hours.  The event could have both quantities.

fosterlynn:
So let's continue, and include at least one use case that contains each action or pair of actions.  More than one if they are handled differently.

I'd also like to not be using our "favorite" quantity name, it is distracting.  Let's say quantity1, quantity2, etc. - and include what it actually is, like you did above.  Otherwise, the assumption creeps in that the quantity is about what all you will do with it, not what it is.

fosterlynn:
Re-stating your use case then:

in cases like ouikit action `use` and `reserve` for planing schedules needs

  *  quantity1 - number of ouikits
  *  time - beginning and end (or sometimes just duration of reservation + topological temporal relations in earlier stage)

and for social accounting needs for action like `reserve`

  *  quantity1 - involvesQuantity - number of ouikits
  *  quantity2 - duration of the reservation

elf-pavlik:
Just to clarify, we need to have possibility to construct a single flow which can meet multiple (sometimes all) of the requirements. We don't talk about creating different flows for different requirements?

fosterlynn:
In the case of using a serialized piece of equipment, say a 3d printer:

 `use` for planning schedules needs

  *  quantity1 - number of hours
  *  time - beginning and end

and for lending for action like `reserve`

  *  quantity1 - number of hours
  *  time - beginning and end

to give or trade away the equipment - decrements
  *  quantity1 - number of 3d printers (1 in this case)
  *  time - one time

fosterlynn:
In the case of working on a process by the hour:

`work`

  *  quantity1 - number of hours
  *  time - beginning and end OR one time

In the case of working on a process by the word:

`work`

  *  quantity1 - number of words
  *  time - beginning and end OR one time

elf-pavlik:
> `use` for planning schedules needs

I don't think we should make
```yaml
qudt:unit: unit:Number
qudt:numericValue: 1
```
some special 'implicit value' and sometimes use specific quantitative value for quantity of a resource and sometimes for something else.

> work
> * quantity1 - number of words
> * time - beginning and end OR one time

what if for *Reports and analysis* agent needs to record both *number of words* and *time*

also intents might specify *2 people* doing translation work

fosterlynn:
Suggestion: making your list a bit more specific, from our discussion, are these a good list of requirements to match to what quantity to use in different use cases?
* planning (from recipe)
* managing schedule of agent or resource
* Incrementing / decrementing resources
* "social accounting" (exchange, distribution)
* analysis/reporting (I think this one might require random other quantities - and suspect we can probably skip it?)

fosterlynn:
Refinement of the above:
* planning (from recipe)
* managing calendar of agent or resource
* "social accounting" == "reciprocity" (exchange, distribution)

fosterlynn:
An attempt to put this into something a bit easier to summarize and compare.  Using the 3 above, availability == calendaring, reciprocity == social accounting.  Leaving out the "1 Number" assumed quantity for people and big equipment for now so it is easier to look at, not taking a position on if we should have it or not.  And understanding that production planning will want to know availability also for usage and work, but availability can exist without a production plan.

use case: ouikits usage (exact)
* planning - number kits, begin time, end time
* availability - number kits, begin time, end time
* reciprocity - number kits, number hours (unit for reciprocity: number of kit-hours)

use case: 3D printer usage
* planning - number hours, begin time, end time (exact)
* availability - begin time, end time
* reciprocity - number hours

use case: work in hours
* planning - number hours, begin time, end time (not necessarily exact)
* availability - begin time, end time
* reciprocity - number hours

use case: work in words
* planning - number words, begin time, end time (not necessarily exact)
* availability - begin time, end time (or possibly not applicable)
* reciprocity - number words

use case: consume items
* planning - number items, one time or begin time/end time of process
* availability - number items, one time probably
* reciprocity - number items

use case: cite a document
* planning - 1 document, one time or begin time/end time of process
* availability - N/A
* reciprocity - number citations (events)

use case: cite by % of process usefulness
* planning - 1 document, one time or begin time/end time of process
* availability - N/A
* reciprocity - number percent

use case: produce items
* planning - number items, one time
* availability - just resource quantity
* reciprocity - number items

use case: ship some items (load and unload)
* planning - number items, one time or begin time/end time
* availability - number items, begin time/end time of process or between decrement and increment
* reciprocity - number items (or usually a service also generated as an output?)

use case:  edit a translation (accept and improve)
* planning - 1 translation, one time
* availability - N/A
* reciprocity - 1 translation

use case: wash the dishes (accept and improve)
* planning - number dishes, one time
* availability - number dishes, begin time, end time
* reciprocity - number dishes

use case: 3D printing class (service)
* planning - 1 service (one event for each service), begin time, end time
* availability - N/A
* reciprocity - 1 service

First draft!  Is this a useful format?  Can't remember the more complex use cases, can add later.

elf-pavlik:
> First draft! Is this a useful format? More complex ones to come....

I think you've made an excellent list @fosterlynn :exclamation: 

> use case: 3D printer usage
>* planning - number hours, begin time, end time
>* availability - begin time, end time
>* reciprocity - number hours

I can see variant where commitment just gets made based on classification / specification and can have number of printers in it. Later it can get refined #464 and replaced with multiple commitments per printer or not plan can stay unrefined and multiple events for specific printers can fulfill that commitment. So we would have:

* planning - number printers, number hours, begin time, end time
* availability - number printers, begin time, end time
* reciprocity - number printers, number hours

>use case: work in hours
>*  planning - number hours, begin time, end time
>*  availability - begin time, end time
>*  reciprocity - number hours

We already discussed that in some cases it could make sense to create at least an intent for more than one person. We don't need to agree on that but I think in some cases coops (eg. doing harvests) may even agree to make commitments for specific number of people which in similar way as printers, can get later refined with commitments by specific people or fulfilled by events. In various scenarios people may just prefer to delegate more to the coop and not negotiate directly and independently with each of the agents which they provide work services to. If commitment with agent managing orchard gets made for 60 people working on specific time, the coop stays responsible that 60 skilled people showing up on time, maybe even arranging transportation for all of them. This way orchard's agent doesn't need to do 60 independent coordination. Some people in some scenarios may simply prefer to delegate some of the responsibilities related to managing their working time, we should support all the A2A scenarios P2P, P2O, O2P and O2O (A - agent, P - person, O - organization)

* planning - number of people working, number hours, begin time, end time
* availability - number of people working, begin time, end time
* reciprocity - number of people working, number hours

> use case: work in hours
> use case: work in words

I see that you separate those by accounting requirement, I can see situation where the same person does work which includes both of accounting ways and while planing schedule, availability needs to accommodate for all the work no matter how it gets accounted. I think we might want to look into *distance* based accounting again which will in similar way make both working time and usage timer more flexible if service gets delivered based on distance. 

> use case: consume items
> * planning - number items, one time or begin time/end time of process

I don't think all the scenarios can depend solely on process interval, for more complex processes it may require specifying that consume happens somewhere around beginning of the process and produce somewhere around end of the process. #453 pretty much recommends using topological temporal relations between flows and processes, besides just having specific beginning and end on both.

> use case: cite by % of process usefulness
>* planning - 1 document, one time or begin time/end time of process
>* availability - N/A
>* reciprocity - number percent

I have impression that we still stay unclear on this UC, in #449 you said that it doesn't need to imply any reciprocity but people could possibly use it as one of the factors when handling reciprocity. I understand that we can assume that this percentage will not get used for anything else than social accounting.

As I mentioned in issue for that specific UC exemplifying this requirement, we should not limit it in any way to *cite* action and make it available for any flow no matter of action. So *work*, *use*, *consume* etc. could carry such perceived % as well. Which gives us for example

use case: work in hours with perceived %

* planning - number hours, begin time, end time
* availability - begin time, end time
* reciprocity - number perceived percent

use case: 3D printer usage with perceived %

* planning - number hours, begin time, end time
* availability - begin time, end time
* reciprocity - number perceived percent

-----

I'll do another round on it and add other use cases like the harvesting one

fosterlynn:
>I have impression that I still stay unclear on this UC, in #449 you said that it doesn't need to imply any reciprocity but people could possibly use it as one of the factors when handling reciprocity. I understand that we can assume that this percentage will not get used for anything else than social accounting.

Maybe the format is lacking?  The "reciprocity" feature is more like logging so it can be used later for exchange of some sort, not necessarily the exchange itself.  In other words, the event quantity defined here will be the one logged in the event which can then be used as-is for part of an exchange. 

Picturing NRP, the plan just has the resource type (like "part design" or often the more specific specification of the design file), and then when you log, you see a list of specific resources, pick one, and enter the percent.  Can't enter any percent in planning, it is completely unknown.

fosterlynn:
>I can see variant where

You can feel free to add into that list if you like with variants or new cases.  Then we can look at it all at once to see what emerges.  I hope. :)

fosterlynn:
>We already discussed that in some cases it could make sense to create at least an intent for more than one person. We don't need to agree on that

I have no problem with that.  I'm focusing on commitments and events because I think that is where we will get our minimum quantities needed.  Intents will have more to them.  And because I have a lot of direct experience with commitments and events, less with intents.

fosterlynn:
>I can see variant where commitment just gets made based on classification / specification and can have number of printers in it. Later it can get refined #464 and replaced with multiple commitments per printer or not, plan can stay unrefined and multiple events for specific printers can fulfill that commitment. So we would have:

>    planning - number printers, number hours, begin time, end time
>    availability - number printers, begin time, end time
>    reciprocity - number printers, number hours

Is this the exact same scenario as ouikits then?

fosterlynn:
> *    planning - number items, one time or begin time/end time of process

>I don't think all the scenarios can depend solely on process interval, for more complex processes it may require specifying that consume happens somewhere around beginning of the process and produce somewhere around end of the process.

That would be covered by the "one time" I think.  The planner would set the commitment time as desired.  I'll get rid of process times though, probably don't want to include thinking about those here, as the flows need time anyhow.

fosterlynn:
I added all your suggested use cases from the comments, and a few more of my own.

>variant where commitment just gets made based on classification / specification and can have number of printers in it - plan can stay unrefined (from @elf-pavlik )

    >planning - number printers, number hours, begin time, end time

I don't think number hours is needed for planning on this one?

fosterlynn:
Some initial impressions and thoughts from the [list](https://github.com/valueflows/valueflows/pull/471#issuecomment-476896380):

Planning: I think we can move all the use cases to one quantity for planning.  That way the planning both from recipe and without recipe becomes much simpler.  I think that the examples we have for `use` action that have both a quantity of a resource and a quantity of time used don't really need the quantity of time used.  
- Even if they are a batch of 3D printers, if they are thought of as a batch for planning, there isn't a great need to specify the time of use, except for availability and scheduling, which use the times, not an additional quantity.  In any case I think the benefit of one quantity outweighs this kind of requirement, which we have no real example of, and affects a relatively small proportion of flows.
- If we are talking about number of people, like 10 people for 15 days, I think that can be handled either with a commitment for each person (in the case of the orchard coordinating the contributions of the people) or by this being a service (in the case of the 10 people being a team (a group agent) that coordinates things like distribution among themselves.

Availability: It is clear that for calculating available of internal used resources, we would need different methods of creating a calendar if we have only the one quantity field that is used in planning.  Basically an app would need to know whether to use the quantity (it is physical amount of a resource) or not (it has a time unit).  If there were 2 named quantities, the app could just use the one related to the involved resource.  For resources that are used externally to the scope agent, then the transfers or reservations or that kind of non-use event would be used to determine availability.

Reciprocity: Sometimes needs an additional quantity or a compound quantity.  Sometimes needs a different quantity.  Although more often it is the same as the planning quantity.  There also might be other examples that show even more variety than the above use cases. There sometimes might be a need for a calculation to figure out the reciprocity quantity from existing quantities.

OK, can't take this any further tonight, but please consider it just thinking in progress, everything will have to be considered together for any conclusions.

elf-pavlik:
I think we shouldn't that *keeping track on resources* related requirement as *planning*. I see availability and scheduling also as part of planning which in many ways relies on simpler *keeping track on resources*. Also *reciprocity* plays role in planning, only *reporting and analysis* doesn't directly relate to planning while it can get used to improve planning. I tried tagging requirements with *past*, *present*, *future* and it breaks down more or less:

|                                             | past | present | future |
-------------------------------------|-------|-----------|---------|
| keeping track on resources |         |      ✓     |     ✓   |
| availability and schedules    |         |      ✓     |     ✓   |
| reciprocity                            |   ✓   |              |     ✓   |
| reporting and analysis         |    ✓  |      ✓     |     ✓   |

fosterlynn:
More random (and possibly rambling) thoughts that I had started earlier:

One is that using the underlyingResource structure would solve a lot of these problems of different quantities, since there would be only one unit associated with a resource or resource specification, and only one quantity needed in flows.  But I also understand not wanting to use it, in some scenarios it is pretty clunky.  On the other hand, some of the RDF academics think that is the way to go for things like apartment leasing.  So I suspect we will want to support it, in addition to whatever we can come up with for quantities to support the different events, actions, and units related to the same resource.  I think it won't really affect this discussion, but probably I should do some detail to make sure.  Lower priority for the moment possibly.

Another: Thinking about planning requirements, and the code involved, I think we could create a plan from a recipe with multiple quantity fields in the flow in terms of scaling them to create the plan, and carrying them down thru I-C-E.  But I think that puts a lot of complexity onto the user who creates recipes, and also on the data entry of I-C-E for all of them, when only a fraction are `use` of stock-type resources with multiple substitutable items.  And worse if there is no recipe and people are creating plans from scratch.

A thought about some different ways of splitting into different fields:  You have done it by kind of a specific definition of different "kinds" of quantities.  So involvesQuantity is for quantities that affect resources, temporalLength is about any quantity that has a time-based unit (I think), the subjective percentage quantity is for valuing a flow in that specific way.  Another possible way to think about it is in terms of the requirements we are tinkering with, like a reciprocityQuantity or something.  Anyhow, it might be useful to try to figure out the best way to segment things, is it "what is it" or "what is it for" or possibly something else.

fosterlynn:
>I think we shouldn't think of that keeping track on resources related requirement as planning. I see availability and scheduling also as part of planning which in many ways relies on simpler keeping track on resources

I was thinking of the planning requirement as creating a plan from a recipe and creating a plan without a recipe.  What do you mean by "keeping track on resources"?

I agree availability and scheduling also play a part in planning, so maybe those are hard to separate.   It might be the same separation as "materials planning" and "capacity planning", which were separated in software for some decades in manufacturing.

>Also reciprocity plays role in planning,

Don't understand that one exactly.  I can see that whatever quantities are needed for later reciprocity are needed at least in the events.  Not sure about planning.

>I tried tagging requirements with past, present, future and it breaks down more or less:

That might help, I had some trouble in my format when it seemed like I-C-E had different requirements.  That might at least partially map to your chart?  Can you expand on what are the implications of your chart?  Or if you think we should use something like that for each use case?

fosterlynn:
>I tried tagging requirements with past, present, future and it breaks down more or less:

Taking off from that thought process, here is another riff.

1. planning: creating a plan from recipe or without recipe (creates intents, commitments)
1.1 scaling the recipe
1.2 availability of resources (checks currentQuantity and/or offers, and commitments)
1.3 commitments to intents (creates commitments)
2. recording observations: against the plan, without a plan (creates events)
3. claims of reciprocity triggered by the observations (creates commitments, checks agreements or reciprocal intents)

(On 3. I think we ought to explore creating claims instead of commitments.)

fosterlynn:
Thoughts at this stage of what we might need for quantities. Using generic names atm.  Putting the time properties there too, so we remember we have them and can use them for scheduling and coordinating.

Intents: Probably depends on how defined the intent is, and I think will need some quantities that we don't need in the commitment-event stages.  But taking intents that are part of an operational plan, which I think means only requests, I see:
* XQuantity (the main quantity, usually the only one)
* hasBeginning
* hasEnd
* theInstantProperty

Commitments:
* XQuantity
* hasBeginning
* hasEnd
* theInstantProperty

Events:
* XQuantity
* YQuantity (other observed quantity needed for calculating reciprocity)
* hasBeginning
* hasEnd
* theInstantProperty

Fulfillment:
* XQuantity
* hasBeginning
* hasEnd

I think the YQuantity could possibly be driven by what is needed to calculate reciprocity, which will need some analysis for how to represent it. (XQuantity may likely be also needed to calculate reciprocity, but not necessarily.) It is possible we will need more than one other quantity for calculating reciprocity, but it hasn't shown up yet in our use cases.  I suspect we need to do an exercise in how reciprocity can be calculated from reciprocal intents and various agreements, and we might find some things.

I'm still basically ignoring the reporting and analysis, assuming it will use what it has.  We might find at some point a specific requirement for additional quantities needed for that, but I propose we wait for a requirement.  And even then, I suspect it would only be needed on events.

fosterlynn:
I'm going to give a try at defining QuantityX.

>I don't see it a matter of 'what unit' but what does that quantitative value mean - how does it relate to that flow, for different meanings we should provide different properties.

(That's from the [citation issue](https://github.com/valueflows/valueflows/issues/449#issuecomment-476725905) but I think that is where some core of our differences are with the quantities. And I think this is a helpful statement to define how to think about it.)  

So, what does the XQuantity mean and how does it relate to the flow?

It is the "edge" quantity, the "weight", the quantity of what is flowing.  The flow is about an action that relates a resource to a process, how that resource contributes to this process.  (Ignoring "pure" transfers for the moment.  And thinking about inputs primarily.)  I'm looking at "resource" as loosely defined to include resource classification and/or resource specification for non-inventoried resources.  So for me besides inventoried items, that includes items that could be inventoried but are not (perhaps not worth it), people's skills and the ability to contribute those skills, also rights to use some item(s), actually using some item(s) as part of contributing to the process.  So it is about the motion more than the thing.

Looking more specifically at "actually using some item(s) as part of contributing to the process", we have found that sometimes the quantity that is used in planning is the quantity of items (especially when the items are substitutable and it is not important to document time details about the usage) and sometimes it is the time used (especially when the item is important to plan, observe, and maintain as a uniquely identified item).

Looking more specifically at the example of defining work in terms of a time quantity (like hours) vs defining work in terms of a non-time quantity (words for translations, probably other forms of "piecework"), these are different kinds of things, but they are the same kind of thing in relation to the flow and as the flow relates to the skills and contribution of the provider as well as the process.  A planning algorithm would look at the RecipeFlow and make use of this one quantity in the same way for hours or words, putting it into the Intent or Commitment, possibly scaled.  

One other note: QUDT and all other quantity ontologies I have looked at include time-based units.  And owl time defines ways to look at durations using a number and a unit, just like quantity ontologies. So it is not a matter of using a "quantity" ontology vs a "time" ontology, and does not have anything to do with rdf preferences as far as I can tell.  It is a semantic choice.  And another note about quantity ontologies: they all seem to provide a way to categorize kinds of quantities, like time, weight, count, etc.

So, first attempt, will need work:

XQuantity: The quantity of a flow relating an economic resource (inventoried or not) to how it acts in a process.  The unit of measure of the quantity is chosen by the users to represent what is most important to them in planning and recording the flow.

bhaugen:
I have not studied everything in this thread yet, so this might duplicate or contradict something, but:
> use case: cite a document
planning - 1 document, one time
availability - N/A
reciprocity - number citations (events) - although this is not really reciprocity

> use case: cite by % of process usefulness
planning - 1 document (or number percent if planned), one time
availability - N/A
reciprocity - number percent

Sensorica had one more variation:

use case: cite by exact value in terms of some quantity of some resource (usually money)
planning - 1 document, one time
availability - N/A
reciprocity - quantity of resource

This is also the approach of https://github.com/ktorn/vdp

bhaugen:
@elf-pavlik @fosterlynn I love the concept and name "social reciprocity" and will start to re-use it immediately, with credit to VF.

elf-pavlik:
> Intent:
>*   XQuantity
>*    hasBeginning
>*    hasEnd
>*    theInstantProperty

> Commitments:
>*    XQuantity
>*    hasBeginning
>*    hasEnd
>*    theInstantProperty

> Events:
>*    XQuantity
>*    YQuantity (other observed quantity needed for calculating reciprocity)
>*    hasBeginning
>*    hasEnd
 >*   theInstantProperty

I've already mentioned uses cases where Intents and Commitments will also need duration aka YQuantity. Also it doesn't always play role in calculating reciprocity but sometimes might. I'll add clear use cases exemplifying it above but quick example: *Commitment to use 4 3d printers for 2 hours (sometimes next week), reciprocity based on % of what will get produced in the process*. Commitment requires both x - quantity of resource, y - duration of use and non of them matter directly for social reciprocity. Still both x and y need to get fulfilled before one can think in terms 'claim' #473 counted based on that r (reciprocity) quantity.

fosterlynn:
>I've already mentioned uses cases where Intents and Commitments will also need duration aka YQuantity. 

I don't see it though, at least for Commitments.  I'd like to write some yaml for anything that seems like it might need also a YQuantity or ZQuantity.  I can start with the bulk 3d printer usage one. Let me know of any more you think fit that requirement, and i'll try them out.

And just to make sure I understand, this is only for where there are 2 quantities listed, not for when there is one time-based quantity, right?

fosterlynn:
>I don't see it though, at least for Commitments.

And I want to say I'm still aware that having just one quantity causes the need for some extra logic in a calendar program such as the one your wrote if it will encounter both counts and hours in the same quantity field in one data feed.

elf-pavlik:
> And just to make sure I understand, this is only for where there are 2 quantities listed, not for when there is one time-based quantity, right?

Currently it seems to me that we may start coming closer that we need dedicated property for quantity meant for social reciprocity. Using printers (or ouikits) in bulk for some time and having reciprocity based on % of perceived added value illustrates that well. In general one can choose to use such % on any flow and such dedicated social reciprocity property seems to serve it perfectly well.

For the number of items (printers, ouikits, drills whatever) and the intended or commited duration of use (before knowing exact beginning and end), we seem to need another two properties to relate to those quantitative values.

So far it seems to me that we can handle everything (any flow that needs to meet all of discussed requirements) with:
* quantity of involved resource (either specific resource or classification or specification) - needed for satisfaction and fulfillment
* duration - when satisfaction and fulfillment also depend on it *use*, *work*, *reserve* possibly other ones
* beginning, end and specific instant - for exact scheduling, doesn't affect satisfaction and fulfillment in quantitative way but may still affect evaluation

Tomorrow we could take exercise of writing YAML with various flows (especially ones you gathered in https://github.com/valueflows/valueflows/pull/471#issuecomment-476896380) in a way that every single flow meets all the requirements. If ensure that any flow can meet all the requirements then everyone can choose not to care about some of those requirements but still those who care will have a clear way to create flows.

fosterlynn:
>variant where commitment just gets made based on classification / specification and can have number of printers in it - plan can stay unrefined (from @elf-pavlik )

>    planning - number printers, number hours, begin time, end time
 >   availability - number printers, begin time, end time
  >  reciprocity - number printers, number hours

But I will do this to see if it is acceptable to not have the extra quantity on planning

```
  - '@id': printing:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeFlow
    action: use
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 10
    time:hasDuration:
      time:unitType: time:unitHour
      time:numericDuration: 7

  - '@id': printing:a25500e0-0106-43cd-8cbb-e74779488835
    '@type': Intent
    action: use
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    receiver: http://www.printing.example/
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 10
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': printing:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 10
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Satisfaction
    satisfies: printing:a25500e0-0106-43cd-8cbb-e74779488835 # the offer (intent)
    satisfiedBy: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    satisfiedQuantity:  # naming?
      qudt:unit: unit:Number
      qudt:numericValue: 10

  - '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 10
    YQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 7
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    fulfilledQuantity: # naming?
      qudt:unit: unit:Number
      qudt:numericValue: 10
```

fosterlynn:
Next up, citation of a document by % of process payment.

```
  - '@id': printing:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeProcess
    skos:note: Manufacture one batch of widgets
    time:hasDuration:
      time:unitType: time:unitDay
      time:numericDuration: 4

  - '@id': mfg:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeFlow
    action: cite
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1

  - '@id': mfg:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: cite
    provider: http://www.mfg.example/
    receiver: http://www.mfg.example/
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
    # the duration is derived from the RecipeProcess in this case
    hasBeginning: 2018-12-07T00:00:00-5:00
    hasEnd: 2018-12-107T00:00:00-5:00

  - '@id': mfg:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: cite
    provider: http://www.mfg.example/
    receiver: http://www.mfg.example/
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
    YQuantity:
      qudt:unit: unit:Percent
      qudt:numericValue: 5
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': mfg:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: mfg:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: mfg:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 1
```

OR, if the percent is determined ahead of time as a policy, or as part of that particular recipe:
```
  - '@id': printing:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeProcess
    skos:note: Manufacture one batch of widgets
    time:hasDuration:
      time:unitType: time:unitDay
      time:numericDuration: 4

  - '@id': mfg:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeFlow
    action: cite
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Percent
      qudt:numericValue: 5

  - '@id': mfg:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: cite
    provider: http://www.mfg.example/
    receiver: http://www.mfg.example/
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Percent
      qudt:numericValue: 5
    # the duration is derived from the RecipeProcess in this case
    hasBeginning: 2018-12-07T00:00:00-5:00
    hasEnd: 2018-12-107T00:00:00-5:00

  - '@id': mfg:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: cite
    provider: http://www.mfg.example/
    receiver: http://www.mfg.example/
    resourceConformsTo: https://www.mfg.example/890823094
    XQuantity:
      qudt:unit: unit:Percent
      qudt:numericValue: 5
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': mfg:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: mfg:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: mfg:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    XQuantity:
      qudt:unit: unit:Percent
      qudt:numericValue: 5
```

elf-pavlik:
> Should flows assume a quantity is the quantity of the involved resource if there is one, else it is 1, if no quantity exists in the flow?

I strongly prefer to always have all the know values explicit. I see implying something tricky and also against [Open-world assumption](https://en.wikipedia.org/wiki/Open-world_assumption) where one doesn't assume operating with absolutely complete knowledge and has to handle *unknown* things.
Whenever piece of information stays available at creation (or update) of any resource, one should assert it in explicit way. Having implied values in practice makes it impossible to record flow where we *don't know* exact quantity or *intentionally don't want to record it* (omitting such quantity would result in some default by implication). I see no reason not to record quantities whenever
* agent creating the record knows them
* agent creating the record intends to record that known quantity

``` yaml
    '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 9
    YQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T14:30:00-5:00
```

I see no reason why not to use `time:hasDuration` as well for the event. If we consider variant where neither number of printers nor duration of action matter for reciprocity, for example it uses that *percived %*, we still need 3rd property to record reciprocity related value. I would see it

``` yaml
    '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 9
    time:hasDuration:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5
    reciprocityQuantity:
      qudt:unit: unit:PERCENT
      qudt:numericValue: 20
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T14:30:00-5:00
```

For now I think we should require explicit duration even if flow has precise beginning and end, based on implementation experience we could consider relaxing it and allow specifying just beginning and end and defining duration as the temporal difference between them.

elf-pavlik:
As for Fulfillment and Satisfaction we might need to define more flexible construct for example:

```yaml
  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    fulfillmentFactor:
      - factor: XQuantity
        factorQuantity:
          qudt:unit: unit:Number
          qudt:numericValue: 9
      - factor: YQuantity
        factorQuantity:
          qudt:unit: unit:Hour
          qudt:numericValue: 6.5
```

So for bulk 3d printers usage with % based reciprocity we would have something like

```yaml
  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    fulfillmentFactor:
      - acceptedCompleteFulfillment: true
        factor: involvesQuantity
        factorQuantity:
          qudt:unit: unit:Number
          qudt:numericValue: 9
      - acceptedCompleteFulfillment: true
        factor: time:duration
        factorQuantity:
          qudt:unit: unit:Hour
          qudt:numericValue: 6.5
      - acceptedCompleteFulfillment: true
        factor: reciprocityQuantity
        factorQuantity:
          qudt:unit: unit:PERCENT
          qudt:numericValue: 20
```

I think each of the Fulfillment (maybe also Satisfaction) factors should have maybe boolean *acceptedCompleteFulfillment: true* or *satisfied*. So in case where something took less time or less equipment than planned, one can mark it still as *acceptedCompleteFulfillment*.

So if flow has other factors, for example `lingo:wordCount`, recorded for analysis and fulfilment but not affecting reciprocity calculations, one can still include it in Fulfillment:

```yaml
      - acceptedCompleteFulfillment: true
        factor: lingo:wordCount
        factorQuantity:
          qudt:unit: unit:Numer
          qudt:numericValue: 1257
```

fosterlynn:
>I see no reason why not to use time:hasDuration as well for the event.

Because it's not about time, it's about quantity, conceptually.  And there are too many subtle options.  I'll do a snippet for the use case of one piece of equipment where it is always logged for the one resource.
```yaml
  - '@id': printing:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeFlow
    action: use
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 7
    time:hasDuration:
      time:unitType: time:unitHour
      time:numericDuration: 7

  - '@id': printing:a25500e0-0106-43cd-8cbb-e74779488835
    '@type': Intent
    action: use
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    receiver: http://www.printing.example/
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 7
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': printing:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 7
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Satisfaction
    satisfies: printing:a25500e0-0106-43cd-8cbb-e74779488835 # the offer (intent)
    satisfiedBy: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 7

  - '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T14:30:00-5:00

  - '@id': printing:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: printing:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    XQuantity:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5

    # and the commitment would get marked finished
```

fosterlynn:
Here is the work by words use case, because that has a quantity that isn't a involves resource kind of thing and isn't a time kind of thing.  The printing impressions one is that kind too.

```yaml
  - '@id': printing:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeProcess
    skos:note: Translate a document
    time:hasDuration:
      time:unitType: time:unitDay
      time:numericDuration: 5

  - '@id': translators:c404cff5-19c7-453d-b0d8-d8e96055bd0b
    '@type': RecipeFlow
    action: work 
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    XQuantity:
      qudt:unit: unit:Word
      qudt:numericValue: 1000

  - '@id': translators:a25500e0-0106-43cd-8cbb-e74779488835
    '@type': Intent
    skos:note: English to Spanish
    action: work
    resourceClassifiedAs:  https://www.wikidata.org/wiki/Q7553 #translation
    receiver: http://www.translators.example/
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    hasBeginning: 2018-12-07T00:00:00-5:00
    hasEnd: 2018-12-12T00:00:00-5:00

  - '@id': translators:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: work
    provider: http://www.alice.example/
    receiver: http://www.translators.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    hasBeginning: 2018-12-07T00:00:00-5:00
    hasEnd: 2018-12-12T00:00:00-5:00

  - '@id': translators:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Satisfaction
    satisfies: translators:a25500e0-0106-43cd-8cbb-e74779488835 # the offer (intent)
    satisfiedBy: translators:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025

  - '@id': translators:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: work
    provider: http://www.alice.example/
    receiver: http://www.translators.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    hasBeginning: 2018-12-07T10:00:00-5:00
    hasEnd: 2018-12-07T15:00:00-5:00

  - '@id': translators:9cb4944b-d26d-4774-a530-d18f2747c0d7
    '@type': Fulfillment
    fulfills: translators:856c43b1-0a63-445f-a56f-707b257f086d # the commitment
    fulfilledBy: translators:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc # the economic event
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
```

fosterlynn:
>As for Fulfillment and Satisfaction we might need to define more flexible construct

I don't think that is necessary.  Either it is finished or not, we don't need that complication.  People know the facts on the ground when they are working on it.  Finished implies that there are different facts on the ground than in the computer, otherwise you wouldn't need it, you could just add up the commitments or events quantities and see if it were finished.

fosterlynn:
>> As for Fulfillment and Satisfaction we might need to define more flexible construct

>I don't think that is necessary

Also, I would like to lean towards defining things that we know are needed because we saw it in a real use case, otherwise we end up with unnecessary complexity which we made up for no reason.  The beauty of REA is its simplicity and universality.

fosterlynn:
>If we consider variant where neither number of printers nor duration of action matter for reciprocity, for example it uses that percived %, we still need 3rd property to record reciprocity related value. I would see it

``` yaml
    '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': EconomicEvent
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 9
    time:hasDuration:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5
    reciprocityQuantity:
      qudt:unit: unit:PERCENT
      qudt:numericValue: 20
    hasBeginning: 2018-12-07T08:00:00-5:00
    hasEnd: 2018-12-07T14:30:00-5:00
```
>For now I think we should require explicit duration even if flow has precise beginning and end, based on implementation experience we could consider relaxing it and allow specifying just beginning and end and defining duration as the temporal difference between them.

I think this illustrates the difference we have had for months and still apparently have.  If the quantity that matters for planning is the involvesQuantity in your example, and the quantity that matters for reciprocity is reciprocityQuantity in your example, then the hasDuration doesn't need to be recorded, there is no reason for it.  Coordination and scheduling/availability concerns can be handled by the hasBeginning and hasEnd throughout the series of flows, it doesn't need a quantity or a duration.  Any accounting analysis or reporting won't need that because it isn't important operationally.

And as I indicated above, I think we should aim at keeping the vocabulary as simple as possible, with due diligence for known use cases, of which we have quite a few, including a handful in production using this basic model.  I think we should add things as needed, not relax as needed.  We'd never find out what can be relaxed.  And in the meantime, we have added confusion and difficulty in implementing the vocabulary.  I think this is an important principal.

I'll also say that in my experience, it is like pulling teeth to get people to even log their time unless their pay depends on it.  If they see a real reason to log something besides personal reasons, like they want the data for improving their methods or whatever, *maybe* they'll do it, big maybe.

bhaugen:
The genius of REA is how it simplified the domains of accounting, economics, planning, and analysis.

Unifying people and agentified groups into EconomicAgent. Unifying material goods, immaterial ideas, and the use of the time of people and machines, into EconomicResources. Unifying sales, payments, work, production, movement, and transfer, into EconomicEvents.

When I worked on ERP systems and then found REA it was clear to me that I could do most of what the ERP system did with a small percentage of the moving parts, and I could do things the ERP system could not do, like plan materials and time-based resources in the same pass through the data, and plan through multiple companies in a supply chain at the same time.

elf-pavlik:
I'll make dedicated issue for that *acceptedCompleteFulfillment*, I don't think it directly impacts what we have discussed.

elf-pavlik:
> I'll do a snippet for the use case of one piece of equipment where it is always logged for the one resource.

I don't think number of resource should matter, IMO quantities shouldn't jump between properties depending on how specific agent chooses to keep inventory of their resources. 

> If the quantity that matters for planning is the involvesQuantity in your example, and the quantity that matters for reciprocity is reciprocityQuantity in your example, then the hasDuration doesn't need to be recorded, there is no reason for it.

Ok, let's change it to Commitment which has duration and topological temporal relation but not specific beginning and end:

``` yaml
    '@id': printing:4cb1f323-4390-4e4a-ac9e-ff64d42f08bc
    '@type': Commitment
    action: use
    provider: http://www.equipshare.example/
    receiver: http://www.printing.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q3834994 #3d printer
    involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 9
    time:hasDuration:
      qudt:unit: unit:Hour
      qudt:numericValue: 6.5
    reciprocityQuantity:
      qudt:unit: unit:PERCENT
      qudt:numericValue: 20
    time:intervalDuring:
      hasBeginning: 2018-12-07T00:00:00-5:00
      hasEnd: 2018-12-16T00:00:00-5:00
```

> I'll also say that in my experience, it is like pulling teeth to get people to even log their time unless their pay depends on it. 

I guess in that case duration gets taken into account for *social reciprocity* requirement and they can use dedicated property for that eg. `reciprocityQuantity`. If they don't care about scheduling and analysis and didn't plan based on duration, they would just omit properties related to those requirements. 


> Unifying material goods, immaterial ideas, and the use of the time of people and machines, into EconomicResources.

Currently I don't see us doing it in VF, use of the time of people and machines gets represented with flows not resources.

> Unifying sales, payments, work, production, movement, and transfer, into EconomicEvents.

How does  *use of the time of people* differ from *work*  ? I understood that REA represent firs with economic resource and second with event or commitment (flow in general)

elf-pavlik:
> Here is the work by words use case, because that has a quantity that isn't a involves resource kind of thing and isn't a time kind of thing.

I understand that word count matters for social reciprocity here, the way you use beginning and end suggests that person doing this work will block full 5 days to work on it

```yaml
  - '@id': translators:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: work
    provider: http://www.alice.example/
    receiver: http://www.translators.example/
    resourceClassifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    XQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    hasBeginning: 2018-12-07T00:00:00-5:00
    hasEnd: 2018-12-12T00:00:00-5:00
```

Besides using reciprocity specific property, I'll also just give deadline and duration estimate which person can use when doing more precise scheduling of one's own working time:

```yaml
  - '@id': translators:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: work
    provider: http://www.alice.example/
    receiver: http://www.translators.example/
    classifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    reciprocityQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    time:hasDuration:
      skos:note: based on 350 words per hour average
      qudt:unit: unit:Hour
      qudt:numericValue: 3
    time:before:
      time:inXSDDateTimeStamp: 2018-12-12T00:00:00-5:00
```
I also chose to use *classifiedAs* instead of *resourceClassified* as but we have #466 to further discuss it.

Going a little bit towards *composition* milestone. Agent providing translation service could have dedicated *work* flows, *use* flows etc. and *service* flow as output which would matter for the social accounting:

```yaml
  - '@id': translators:856c43b1-0a63-445f-a56f-707b257f086d
    '@type': Commitment
    action: service
    provider: http://www.alice.example/
    receiver: http://www.translators.example/
    classifiedAs: https://www.wikidata.org/wiki/Q7553 #translation
    reciprocityQuantity:
      qudt:unit: unit:Words
      qudt:numericValue: 1025
    time:before:
      time:inXSDDateTimeStamp: 2018-12-12T00:00:00-5:00
```

I think application could just consider *reciprocityQuantity* and it still work no matter if one would create *work* flow or *service* flow in this scenario.

bhaugen:
@elf-pavlik 
>> Unifying material goods, immaterial ideas, and the use of the time of people and machines, into EconomicResources.

> Currently I don't see us doing it in VF, use of the time of people and machines gets represented with flows not resources.

>> Unifying sales, payments, work, production, movement, and transfer, into EconomicEvents.

> How does use of the time of people differ from work ? I understood that REA represent firs with economic resource and second with event or commitment (flow in general)

We've discussed resources which can be identified and resources that can only be inferred.

I've seen work (and use of machines) represented two different ways:
1. As a Use event for the machine, and a Work event for the person. In both cases the quantity is usually hours. The machine or the person's skill is the resource. For the machine, a use event requires a different unit than a production event (where you are producing a machine).
2. As an interval on a timeline representing the machine or person's time, where the event consumes the interval. The machine or person's time is the resource.

The Use and Work events are somewhat simpler to code. But the machine use event has is own complication of requiring a different unit than the unit to count the number of machines in inventory.

The timeline is more complex in software construction if you want to create a calendar-per-resource, but allows for reserving and scheduling calendar intervals. That model is almost always used for scheduling something like a meeting room. 

So a work event is a shortcut for the use of the person's time with a particular skill. 

@fosterlynn says you think of the event quantity as we have currently defined it as affecting a particular resource, and that might have factored into some of your questions to me. In this discussion, the representation of use as consuming a resource which represents an interval of the timeline of a machine or agent would be like that (quantity consuming a resource). In general, I think it is a cleaner model. But then we would need to figure out how the interval on a timeline is related to the machine or agent's skill or type of work.

elf-pavlik:
Currently `vf:EconomicResource` pretty much represents a stock and so far we only used it for what we can call products. I try to stick here what our definitions include and YAML examples we've written. Not conceptual ideas of what someone could mean when using english term *resource*.

> As an interval on a timeline representing the machine or person's time, where the event consumes the interval.

In the simple prototype I did https://github.com/valueflows/valueflows/pull/471#issuecomment-476355884 draws a blue square based on `vf:Intent` where person wants to provide work (even when receives it oneself counts) or steward of equipment intends to make it available. So in this case an instance of `vf:EconomicResource` **does not** represent and interval on the timeline, an instance of `vf:Intent` does, or `vf:Commitment`, or `vf:EconomicEvent`. 

> The machine or person's time is the resource

Again we have snippets where we represent the machine with an instance of `vf:EconomicResource`, I don't think we have snippets where we represent *usefulness* of it as an instance of `vf:EconomicResource`. If we did, when we move machine from one place to another, does it have to move any *usefulness* resource between those places as well? How do you *produce* and *consume* those *usefulness* resources? What `vf:currentQuantity` that *usefullness* resource would have?
If we want to explore representing *usefullness* as an instance of `vf:EconomicResource` let's create dedicated PR (or issue) for it and try to address all those questions there. For now we seem to practically use instances of `vf:EconomicResource` to represent sock what we can call products.

elf-pavlik:
Currently `vf:EconomicResource` pretty much represents a stock and so far we only used it for what we can call products. I try to stick here what our definitions include and YAML examples we've written. Not conceptual ideas of what someone could mean when using english term *resource*.

> As an interval on a timeline representing the machine or person's time, where the event consumes the interval.

In the simple prototype I did https://github.com/valueflows/valueflows/pull/471#issuecomment-476355884 draws a blue square based on `vf:Intent` where person wants to provide work (even when receives it oneself counts) or steward of equipment intends to make it available. So in this case an instance of `vf:EconomicResource` **does not** represent and interval on the timeline, an instance of `vf:Intent` does, or `vf:Commitment`, or `vf:EconomicEvent`. 

> The machine or person's time is the resource

Again we have snippets where we represent the machine with an instance of `vf:EconomicResource`, I don't think we have snippets where we represent *usefulness* of it as an instance of `vf:EconomicResource`. If we did, when we move machine from one place to another, does it have to move any *usefulness* resource between those places as well? How do you *produce* and *consume* those *usefulness* resources? What `vf:currentQuantity` that *usefullness* resource would have?
If we want to explore representing *usefullness* as an instance of `vf:EconomicResource` let's create dedicated PR (or issue) for it and try to address all those questions there. For now we seem to practically use instances of `vf:EconomicResource` to represent stock of what we can call products.

fosterlynn:
>Currently vf:EconomicResource pretty much represents a stock and so far we only used it for what we can call products

Agree, currently we don't have an instance of vf:EconomicResource unless there is an inventoried resource in some way, with some possible looseness of definition there, for example apartment leases - or not, depending on which sets of conversations influenced what understanding we each came away with. 

But in any case, I think it is worth taking a look at the broader "resource" concepts in terms of conceptualizing REA, no matter how it was expressed.

I also think that we just misunderstood each other in terms of the affectedQuantity.  I always understood that to be the event quantity, which could apply to any kind of broader "resource" concept, which in the data could be a vf:affects, one or more vf:resourceClassifiedAs, and/or a vf:resourceConformsTo (on EconomicEvent).  Just shows how important the name and definition really are.  Which is also why I understood the name in an EconomicEvent to be vf:resourceConformsTo for example, not vf:conformsTo - it is about defining the resource-y part of the event, how the event relates to the little-r "resource" involved.

Anyhow, from today's chat, we will move ahead by:
* listing a handful of use cases to look at that cover broad enough requirements
* each of us (@elf-pavlik and me) will do a PR to represent our point of view, including
     * textual explanation for the doc to explain the quantities and how to use them
     * ttl file updates for the data structure
     * full yaml examples for the use cases
* if we still don't agree, defining the differences in the VF doc and asking there for help from people who want to use it
* and trying out our perspectives with the real user groups we do work with

Anything to add or change on my understanding from the chat, Pavlik?

fosterlynn:
Listing requirements we want to cover:
* different ratios of people + equipment to duration of work and usage can handle the same quantity (@elf-pavlik from chat)  
* various "kinds" of quantities: inventoried resource impacting, time based, neither (for example, words, miles, printing impressions) - note there are different intersecting angles to look at "kind", like what it is used for, or what kind of unit it is (like durations can be represented as a qudt quantity or a time duration)
* what else is needed for reciprocity definitions or calculations
     * do we have an example where there are other "extra" quantities needed besides for reciprocity? 
     * do we have an example where there is more than one "extra" quantity needed for reciprocity?
* cover every action at least once in the whole set of use cases
* others????

fosterlynn:
Suggested  use cases:
* threading (or orchard, @elf-pavlik pick) >1 person, >1 equipment, demand-driven?
     * use
     * work
     * consume?
     * ?
* edit a document using words as work qty, supply-driven
     * work
     * accept
     * improve
* taxi ride - lots of potential quantities as part of a service, 
     * service
     * ?

Please continue to fill this in..... without needlessly duplicating features....

elf-pavlik:
> edit a document using words as work qty, supply-driven

I'd like to include variant where both work and usage get recorded on the process and edit gets delivered as *service*, I think sometimes also *improve* could have that word count. 

As for requirements let's for every flow meet all the discussed requirements. So no matter if
* *work* flow in addition records word count
* *work* flow goes as input to process which provides *service* with word count
* *work* flow goes s input to process which provides *service* with % for reciprocity but for analysis we also want to record the word count

Agent doing the *work* might still need to meet the *scheduling* requirement and **start with work commitment that has only estimated duration** and later as refining the plan #464 updated/replace with commitment which has specific beginning and end.

elf-pavlik:
> But in any case, I think it is worth taking a look at the broader "resource" concepts in terms of conceptualizing REA, no matter how it was expressed.

I re-opened #375 we could possibly move that part of conversation there.

fosterlynn:
>> But in any case, I think it is worth taking a look at the broader "resource" concepts in terms of conceptualizing REA, no matter how it was expressed.

>I re-opened #375 we could possibly move that part of conversation there.

OK, but my point was more that we should leave open the possibility to define vf properties at a sweet spot of specificity that maintains REA's simplicity and essential definitions, but makes it operationally useful.

fosterlynn:
>* work flow goes as input to process which provides service with word count
>* work flow goes as input to process which provides service with % for reciprocity but for analysis we also want to record the word count (possibly same count on both work and service)

This isn't how they thought about it. As clarification, this example process would be one in a series of processes in the recipe: initial translation, editing, proofreading, formatting, that kind of thing - can't remember the exact process types or sequence.  So after the first output of initial translation, it is the same economic resource being accepted and improved.  A series where it wouldn't make much sense to have services along the way.

I'll start a placeholder issue on "workflow" recipes and plans, which is what we called this sequence of accept/improve on the same resource.  Will need to find old documentation.

fosterlynn:

PR #475, opened 2019-03-31
TemporalDuration > Duration

fosterlynn:
made the Process.duration more specific, limiting the fields that people would use a bit - unitType, numericDuration

ref. https://www.w3.org/TR/owl-time/#duration


elf-pavlik:
Oh, I missed this one in #453 

We don't need `vf:duration` people should use [`time:hasDuration`](https://www.w3.org/TR/owl-time/#time:hasDuration) as suggested in https://valueflo.ws/specification/external-terms.html#time

Could you change this PR to simply remove `vf:duration` all thogether ?

fosterlynn:
>Could you change this PR to simply remove vf:duration all thogether ?

done

PR #478, opened 2019-04-02
Lynn quantities option

fosterlynn:
Creating PR in case you want to look at where I got to @elf-pavlik .  Not sure there is much worth looking at yet though.

Status:
* work in words, complete (may want to add more requirements, but this shows the simple version which is the actual use case in the wild)
* threading - first draft of recipe done, nothing beyond that (simplified, not sure about some of the machining inputs)
* taxi - nothing done except some thoughts from yesterday before we talked on how the recipe can indicate what extra reciprocal quantities should be entered on the event, line 31

### DIFF
--- a/examples/machining.yaml
+++ b/examples/machining.yaml
@@ -1,70 +1,111 @@
-# Example: Part of planning and doing a translation - one step, proofreading
+# Example: Machine shop, one step of threading a part
 
 '@context':
   - https://git.io/vf-examples-jsonld-context
   - alice: https://alice.example/
-    translators: https://translators.example/
+    machine: https://machine-shop.example/
 
-'@id': rgh:valueflows/valueflows/master/examples/translate.yaml
+'@id': rgh:valueflows/valueflows/master/examples/machining.yaml
 '@graph':
 
-  # recipe (fragment: proofreading process)
+  # recipe
 
-  - '@id': translators:c404cff5-19c7-453d-b0d8-d8e96055bd0b
+  - '@id': machine:a7e94dde-2997-451e-a7e1-8d364d5e6dee
+    '@type': ResourceSpecification
+    name: Threaded part 87
+    unit: unit:Number
+    substitutable: True
+
+  - '@id': machine:2d350055-acba-4e17-92f8-8b126272ee40
+    '@type': ResourceSpecification
+    name: Unthreaded part 45
+    unit: unit:Number
+    substitutable: True
+
+  - '@id': machine:07819789-dd51-44c3-b35c-9210165bc832
+    '@type': ResourceSpecification
+    classifiedAs: https://www.wikidata.org/wiki/Q2211295 # drilling machine
+    name: Drill Press
+    unit: unit:Number
+    unitOfUse: unit:Hour # want a better way to do this!
+    substitutable: True
+
+  - '@id': machine:c404cff5-19c7-453d-b0d8-d8e96055bd0b
     '@type': RecipeProcess
-    skos:note: Proofread a translated blog post
+    skos:note: Thread a part
     time:hasDuration:
-      time:unitType: time:unitDay
+      time:unitType: time:unitHour
       time:numericDuration: 2
-    scalingFormula: D
+    scalingFormula: D*M
 
-  - '@id': translators:13cf44e5-eaa1-4ad0-88d5-e2910a73f377
+  - '@id': machine:c7897c39-7f05-4a5d-a487-80e130a2414b
     '@type': RecipeFlow
-    action: work 
-    resourceClassifiedAs: https://www.wikidata.org/wiki/Q834191 # proofreading
+    action: produce 
+    resourceConformsTo: machine:a7e94dde-2997-451e-a7e1-8d364d5e6dee # threaded part
     quantifiedAs:
-      qudt:unit: unit:Word
-      qudt:numericValue: 1
-    scalingFormula: M*Q
+      qudt:unit: unit:Number
+      qudt:numericValue: 80 # minimum batch size to be worth it for setup
+    scalingFormula: Q*M
     scalesPlan: True
 
-  - '@id': translators:9f112621-3f13-4f48-a3e2-fd2ca99493cd
+  - '@id': machine:13cf44e5-eaa1-4ad0-88d5-e2910a73f377
     '@type': RecipeFlow
-    action: accept
-    resourceClassifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    action: work 
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
+    quantifiedAs:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 2
+    scalingFormula: Q*M
+
+  - '@id': machine:9f112621-3f13-4f48-a3e2-fd2ca99493cd
+    '@type': RecipeFlow
+    action: consume
+    resourceConformsTo: machine:2d350055-acba-4e17-92f8-8b126272ee40 # unthreaded part
     quantifiedAs:
       qudt:unit: unit:Number
-      qudt:numericValue: 1
-    scalingFormula: Q
+      qudt:numericValue: 80
+    scalingFormula: Q*M
 
-  - '@id': translators:c7897c39-7f05-4a5d-a487-80e130a2414b
+  - '@id': machine:ff208521-9fec-4bb4-a6d3-197f6735cb26
     '@type': RecipeFlow
-    action: improve 
-    resourceClassifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    action: use
+    resourceConformsTo: machine:07819789-dd51-44c3-b35c-9210165bc832 # drill press
     quantifiedAs:
-      qudt:unit: unit:Number
-      qudt:numericValue: 1
-    scalingFormula: Q
+      qudt:unit: unit:Hour
+      qudt:numericValue: 2
+    scalingFormula: Q*M #?
 
-  # resources
+  # resources before
 
-  - '@id': translators:e1721a61-cd47-4556-84b9-8b1b81da15bf
+  - '@id': machine:e1721a61-cd47-4556-84b9-8b1b81da15bf
     '@type': EconomicResource
-    name: Summer of Commoning: A Tour of French Communities and Projects
-    classifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    name: Drill Press
+    conformsTo: machine:07819789-dd51-44c3-b35c-9210165bc832
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: 1
+      qudt:numericValue: 4
 
-  # plan
+  - '@id': machine:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': EconomicResource
+    name: Unthreaded Part 45
+    conformsTo: machine:2d350055-acba-4e17-92f8-8b126272ee40
+    currentQuantity:
+      qudt:unit: unit:Number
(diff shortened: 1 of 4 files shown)


fosterlynn:
Tried the drill press example, tried to make it a number-hours example but it just wasn't.  Will try another one to get that requirement in - maybe back to the kits....

examples/machining.yaml
```diff
@@ -0,0 +1,430 @@
...
+    name: Unthreaded part 45
+    unit: unit:Number
+    substitutable: True
+
+  - '@id': machine:07819789-dd51-44c3-b35c-9210165bc832
+    '@type': ResourceSpecification
+    classifiedAs: https://www.wikidata.org/wiki/Q2211295 # drilling machine
+    name: Drill Press
+    unit: unit:Number
+    unitOfUse: unit:Hour # want a better way to do this!
```
elf-pavlik: Would you ever want to put not a time unit here? Also what 'unit' means here, can one use machine for fractions of unit, for example .5 hour?

examples/machining.yaml
```diff
@@ -0,0 +1,430 @@
...
+    time:hasDuration:
+      time:unitType: time:unitHour
+      time:numericDuration: 2
+    scalingFormula: D*M
+
+  - '@id': machine:c7897c39-7f05-4a5d-a487-80e130a2414b
+    '@type': RecipeFlow
+    action: produce 
+    resourceConformsTo: machine:a7e94dde-2997-451e-a7e1-8d364d5e6dee # threaded part
+    quantifiedAs:
```
elf-pavlik: ok, in some older example i experimented with something similar, i think i called it `atomicQuantity` following the original meaning of *atom*. this also seems to answer my question above about `unitOfUse` in a way that if one wants to put some *atomic* / *quantized* constrains some other construct would take responsibility of that.

fosterlynn:
Also pushed a taxi example, a bit rough and experimental.

examples/translate.yaml
```diff
@@ -0,0 +1,182 @@
...
+    '@type': Commitment
+    action: work
+    provider: http://www.alice.example/
+    receiver: http://www.translators.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q834191 # proofreading
+    quantifiedAs:
+      qudt:unit: unit:Words
+      qudt:numericValue: 1025
+    hasBeginning: 2018-12-08T10:00:00-5:00
+    hasEnd: 2018-12-08T15:00:00-5:00
```
elf-pavlik: What if for scheduling purpose Alice wants to just include estimated duration of work but not select specific beginning and end yet, maybe just interval during which it happens like your *before* & *after* intent.

examples/use-kits.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - '@id': ouikit:ff208521-9fec-4bb4-a6d3-197f6735cb26
+    '@type': RecipeFlow
+    action: use
+    resourceConformsTo: ouikit:a7e94dde-2997-451e-a7e1-8d364d5e6dee  # reusable eating kit
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    alsoQuantifiedAs:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 1
```
elf-pavlik: why not `time:hasDuration` ?

examples/use-kits.yaml
```diff
@@ -0,0 +1,89 @@
...
+    action: use
+    provider: http://www.ouikit.org/
+    receiver: http://www.ouikit.org/
+    involves: ouikit:9396d2d1-ded3-4b1c-8144-1c024e641039 # ouikits resource
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 29
+    alsoQuantifiedAs:
+      qudt:unit: unit:Hour
+      qudt:numericValue: 2.5
```
elf-pavlik: How application supposed to understand what `quantifiedAs` and `alsoQuantifiedAs` supposed to mean? Would anything change if we swap those two values?

examples/machining.yaml
```diff
@@ -0,0 +1,430 @@
...
+    conformsTo: machine:2d350055-acba-4e17-92f8-8b126272ee40
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  # plan
+
+  # Multiplier = demand / recipe batch size, M = 1000 / 80 = 12.5
+  # first pass, create plan from recipe assuming no limits on time (this is the optimum because of the least setup time)
+  # second pass, capacity planning to modify plan to fit due date and person/machine availability, this is the plan
```
elf-pavlik: Having two distinct scaling factors:
* number of machines and people
* duration of work and usage

Application could easily calculate 'on-the-fly' all the possible combinations, given specific number (1000 - or adjusted to full batches) it could even provide UI with a rectangle where one can drag a corner and ratio between factor changes while their products stays matching what it takes to produce that specific number.

If you create plan for 1 person and 1 machine from a recipe (duration 25h), how from there application supposed to help with modifying the plan to what you have below if the recipe seems to have the information needed for it.

examples/machining.yaml
fosterlynn: Yes, fractions would be perfectly valid.  An example where unit of use may not be a time unit is printing press impressions.

examples/translate.yaml
fosterlynn: Hmmm, commitments are I guess always estimates to some degree in terms of time.  Do you think hasBeginning and hasEnd cannot represent estimates?  I can see a lot of use for time:before in commitments, would work like "due".  Or are you meaning something like "6 hours"?  I don't think that helps much in terms of scheduling.  But I suppose whatever time: elements people want to use for scheduling they could do that, if they think that it will be supported by the receiving app.

examples/use-kits.yaml
fosterlynn: Because I want to use it as a quantity, with other quantities, in formulas that understand quantities and queries that understand quantities.  It doesn't matter that it could also be expressed as a duration.

examples/use-kits.yaml
fosterlynn: Yeah, quantifiedAs follows down from intents and commitments.  You could only fulfill a commitment if the event is using the same meaning of quantity and the same unit, compare planned vs actual, etc.  The alsoQuantifiedAs would be a different unit or at least a different functional quantity than found in intents or commitments before the event.

examples/machining.yaml
fosterlynn: Yeah, there could be all kinds of apps to help create plans from recipes, and I've seen also a lot of tweaking and hands-on adjustment of plans. 

An app could do what I did, which was to take the straight line plan (25 hours), which gives you minimum machine setup time, then scrunch it to fit the deadline just like you are suggesting with the dragging the corner of the rectangle to get different ratios.  At the same time they would have to take into account other demands on machines, and the schedules of the people.  Since it is only working with classifications/specifications, not specific machines, it can take all available substitutable equipment and skills into account.

examples/machining.yaml
elf-pavlik: Ok, and vehicle could have 'distance', still for scheduling usage will need to have duration.

examples/translate.yaml
elf-pavlik: So you see no problem to have on commitment `time:hasDuration` plus some topological temporal relations like *before* and *after* ?

examples/use-kits.yaml
elf-pavlik: `time:hasDuration` also references a quantitative value, one can use it with formulas that understand quantities and queries that understand quantities. Besides all that it has very clear meaning while *quantifiedAs*, *alsoQuantifiedAs*, *additionallyQuantifiedAs* have no meaning.

examples/use-kits.yaml
elf-pavlik: > You could only fulfill a commitment if the event is using the same meaning of quantity and the same unit, compare planned vs actual, etc.

and where application supposed to get the 'meaning of quantity?

> The alsoQuantifiedAs would be a different unit or at least a different functional quantity than found in intents or commitments before the event.

I see no reason why Intents and Commitments wouldn't have both number of kits and usage duration and require satisfaction and fulfillment of both. So neither 10 kits for 2.5 hour or 29 for 1 hour can fully satisfy or fulfill.

examples/machining.yaml
elf-pavlik: > An app could do what I did, which was to take the straight line plan (25 hours), which gives you minimum machine setup time, then scrunch it to fit the deadline just like you are suggesting with the dragging the corner of the rectangle to get different ratios.

This sounds like a good candidate for prototyping exercise. I could do prototype which does it directly from Recipe and can scale volume of equipment and people as well as duration. You could give a shot to your approach and we could discuss our implementation experiences.
At this moment I think doing it directly from Recipe should come simpler and we don't get this intermediate plan which no one actually would ever want to execute in at any point in described scenario.

examples/use-kits.yaml
fosterlynn: >I see no reason why Intents and Commitments wouldn't have both number of kits and usage duration and require satisfaction and fulfillment of both.

But why would they need that? The scheduling and coordination happens with the time fields.

fosterlynn: >and where application supposed to get the 'meaning of quantity?

Haven't really seen a problem in practice so far.  The resource specification knows the unit, so that clues people in.  And there is a description on a commitment that adds any detail not provided by the action.  If you see a form to fill in that says "work" "translate english to spanish" and "hours", you just fill in your hours and the time information.  In our UIs you can see on the same page the process information and the input document, etc.

For usage, you see the equipment, or at least its specification, the unit, so not confusing.  For consume and produce, the same.

I wasn't sure what to do with the taxi service stuff, but I don't actually like what I did in the current state of the example, I'm going to change it, but need to think a bit.

examples/machining.yaml
fosterlynn: If there's any way you could think about the core naming with me first, that would be very helpful.  I understand it might be different if you want many quantities and I want just one as the core.

examples/translate.yaml
fosterlynn: I'm ok with that if it is not a quantity field in a different vocab and format.  If there is already a quantity field, and this is used for scheduling, not reciprocity etc.

elf-pavlik: I think we agreed that in all UCs we can have variant where reciprocity gets handled based on %, so applications should make no assumptions if it gets handled based on quantity of involved resource, duration of flow or some other recorded quantity.

fosterlynn: Reciprocity is very different processing than scheduling, and reciprocity will be using various quantities to do its calculations - seems needlessly messy to ask that code to use different vocabularies for different quantities.

Since scheduling happens before the fact and is an estimate, perhaps that duration on a commitment is different that a time-based quantity on the event?

elf-pavlik: > seems needlessly messy to ask that code to use different vocabularies for different quantities

I don't see any difference if terms come from one or many vocabs

> Since scheduling happens before the fact and is an estimate, perhaps that duration on a commitment is different that a time-based quantity on the event?

duration of the event fulfills duration of the commitment

fosterlynn: >duration of the event fulfills duration of the commitment

No it doesn't.  Fulfillment fulfills quantities.  Why complexify it?  Durations are for scheduling.

>I don't see any difference if terms come from one or many vocabs

So if you are using a quantity in an equation (such as happens with distribution logic), sometimes it is numericValue, sometimes it is numericDuration, and the units you might be comparing are a unit or a unitType?  Ouch.

elf-pavlik: i don't think any case would involve comparing quantities of involved resources to duration of a flow.

PR #481, opened 2019-04-06
changing name of involves

fosterlynn:
involves > resourceInventoriedAs in the flows

much clearer name, and also groups better conceptually with resourceClassifiedAs and resourceConformsTo, clarifying these together give you what is known about the little-r resource

[edit If there is agreement, I'll also change the existing examples and add here, or do as a separate PR right away]

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -175,9 +175,9 @@ vf:recipeOutputOf
         vs:term_status      "unstable" ;
         rdfs:comment        "Relates an output flow to it's node in a recipe." .
 
-vf:involves
+vf:resourceInventoriedAs
         a                   owl:ObjectProperty ;
-        rdfs:label          "involves" ;
+        rdfs:label          "resource inventoried as" ;
         rdfs:domain         [ owl:unionOf (vf:Commitment vf:Intent vf:EconomicEvent) ] ; 
         rdfs:range          vf:EconomicResource ;
         vs:term_status      "unstable" ;


elf-pavlik:
What do you think about also using `resourceQuantity` - quantity of the the *little-r resource* and in next step we can figure out if we also consider *work* and *usage* as *little-r resource* or we just consider them as flows.

fosterlynn:
>What do you think about also using resourceQuantity - quantity of the the little-r resource and in next step we can figure out if we also consider work and usage as little-r resource or we just consider them as flows.

The problem I can see with resourceQuantity is that it sounds like the quantity of the resource itself, while in this case it is the quantity of the flow from (or to) that resource, although it might have been or will be part of the currentQuantity of the resource.

Where the involves, the resourceClassifiedAs, and the resourceConformsTo, I think of as a collective reference to the little-r resource, which exists (or will exist) outside of this flow as its own thing, independent of this flow although affected by it and others.  The quantity called involves is not about this referenced resource.

The way I think of work is that the little-r resource is a person's capability to work, perhaps with a certain skill.  And the quantity of work that is promised or happened, say 5 hours today, is the quantity on the flow.  Then it matches reasonably the way we think of a resource of 100 widgets, and 5 of them flowed today into this process.  I think the work capability little-r resource could potentially be instantiated too, with calendar as its currentQuantity or something like that.

fosterlynn:
>What do you think about also using resourceQuantity

Actually, re-reading your comment @elf-pavlik , I wonder if I misunderstood, because of the "also".  If I did not respond to what you meant, can you try me again?

elf-pavlik:
sorry, let's deal with quantities in already existing issues and PRs

PR #484, opened 2019-04-08
introduce multilateral exchange agreements

elf-pavlik:
In this paragraph I intend to explain that exchanges do not have to happen exclusively between two agents. VF supports multilateral exchange agreements and one between two agents comes as common special case of more general multilateral agreements.

### DIFF
--- a/docs/introduction/exchanges.md
+++ b/docs/introduction/exchanges.md
@@ -2,6 +2,8 @@
 
 Here we look at exchanges of resources from an independent or neutral viewpoint (not the viewpoint of one of the Agents in the exchange). For example, from one Agent's viewpoint, the exchange may be a Purchase, from the other Agent's viewpoint, it might be a Sale. From the neutral viewpoint, it is an exchange of resources, with usually at least two flows of resources, one from each direction. So for example, the seller might give some goods to the buyer, and the buyer might give some money to the seller. Or in a barter exchange, one agent might give the other some books, and the other agent might compensate with some cookies.
 
+Valueflows also enables multilateral exchange agreements. Any number of agents can commit to flows where they provide something and flows where they receive something. This way creating a reciprocal cycle in the flows graph. So for example, Alice can provide apples from her orchard to Bob, who can provide accommodation to Claire, who can provide tutoring to Alice's children. Such exchanges can happen in infinite number of possible ways, as long as all agents participating agree on specific reciprocal cycle in the flows graph.
+
 Exchange is ubiquitous on the internet today, with offers everywhere. In ValueFlows, we want to track not only the offers and promises, but also the actual flows of resources in networks, in all directions. And we support exchanges that don't involve money as well as those that do.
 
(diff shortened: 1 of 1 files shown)


elf-pavlik:
This one shows short reciprocal cycle which I would like to start establishing at some point:
* I provide software development services to OFN project (represented as process)
* OFN process outputs new versions of OFN platform (knowledge resource) each month
* La Huerta Lucia (also represented as farming process) - produces food - and people running it use OFN platform
* La Huerta Lucia provides organic food to Tashirat
* Tashirat provides housing serivces to me, which includes usage of a bungalow plus some other mainenance related services.

It has additional shorter cycle where I get food directly from the farm, but eventually they might want to provide more than i can consume and I could start creating longer reciprocal cycles based on that.

reciprocity-cycle

docs/introduction/exchanges.md
```diff
@@ -2,6 +2,8 @@
 
 Here we look at exchanges of resources from an independent or neutral viewpoint (not the viewpoint of one of the Agents in the exchange). For example, from one Agent's viewpoint, the exchange may be a Purchase, from the other Agent's viewpoint, it might be a Sale. From the neutral viewpoint, it is an exchange of resources, with usually at least two flows of resources, one from each direction. So for example, the seller might give some goods to the buyer, and the buyer might give some money to the seller. Or in a barter exchange, one agent might give the other some books, and the other agent might compensate with some cookies.
 
+Valueflows also enables multilateral exchange agreements. Any number of agents can commit to flows where they provide something and flows where they receive something. This way creating a reciprocal cycle in the flows graph. So for example, Alice can provide apples from her orchard to Bob, who can provide accommodation to Claire, who can provide tutoring to Alice's children. Such exchanges can happen in infinite number of possible ways, as long as all agents participating agree on specific reciprocal cycle in the flows graph.
```
fosterlynn: I like it.  Suggest moving it to the end of the section, maybe fits best with that last paragraph?

elf-pavlik:
moved

elf-pavlik:
Looking at image in https://valueflo.ws/introduction/concepts.html#another-little-value-flow

dual-exchange

It matches the first case on the diagram below (plus it has few steps of supply subchain which diagram below doesn't include at this moment)

exchanges

Do you have somewhere original of that coffee and olive oil diagram? I would like to make two variants to mach next two degrees of  multilateral agreements from my diagram. I could add rice and maybe potatoes.

bhaugen:
@elf-pavlik sorry, looked but did not find. I think it was a png export from an odp file (Libre Office Impress) but can't find it in either of those types or with "olive" in the name.

elf-pavlik:
Thanks @bhaugen I can easily reproduce it, just wanted to check in case you had done it with draw.io or google drawing.

fosterlynn:
Looks very interesting @elf-pavlik !  So, this would help us think about: multi-agent proposals and agreements?  I'm trying to start work on proposals and missing parts of exchange agreements now, so am interested in how you are structuring it in the data.... (while still keeping track of transfers and quantity fields as they continue to evolve)

PR #488, opened 2019-04-13
interim name change to quantifiedAs

fosterlynn:
I'd like to do this while we continue to discuss it related to the scheduling milestone.  Seems better than what we have now in any case, and i'm not sure when we will finish hashing out the quantities.

### DIFF
--- a/examples/fulfill-satisfy.yaml
+++ b/examples/fulfill-satisfy.yaml
@@ -16,17 +16,11 @@
     action: work
     receiver: https://manufacturing.example/
     resourceClassifiedAs: https://www.wikidata.org/wiki/Q192047 # machining
-    intendedQuantity:
+    quantifiedAs:
       qudt:unit: unit:Hour
       qudt:numericValue: 8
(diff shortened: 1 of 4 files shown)


elf-pavlik:
could you please update examples as well?

fosterlynn:
>could you please update examples as well?

done; and updated the times as well (oh maybe that was already done.... ah well, hopefully no yucky merge conflicts! if so, let me know, I'll fix ahead - was trying to not have to figure out rebase)

elf-pavlik:
I included alternative version of fulfill-satisfy in #492 
If we can't resolve it in next couple of days I think we can rename examples with some suffixes like `-lq` and `-pq` to include both versions while we continue figuring things out.

fosterlynn:
I'm closing this.  Had some push back on the name quantifiedAs, changed it to flowQuantity in the graphql reference for now until we can resolve quantity - or until next release of graphql spec basically.  And with all this time passed, there is no "interim".

PR #489, opened 2019-04-14
claim model

fosterlynn:
@elf-pavlik @bhaugen 
This is to replace the removed Reciprocity, which created Commitments from Economic Events.  It was observed that Claims are different than Commitments because commitments can only be committed by the provider, while claims are initiated by the hopeful receiver.

Model looks like this.  Claim has most of the properties of I-C-E- flows.  Claim basically references both the event(s) that triggered the claim and the event(s) that resolve the claim.  [edit: WRONG model, see following comments]
claim

I included a brief description in the flows page.  Probably a lot more could be said eventually.

The terms are straight out of McCarthy / REA.  

Note REA people also talk about claims and 'negative claims'.  Negative claims are things like issuing dividends.  I haven't defined those here, not sure if we want to do that, or consider those commitments.  My impression is the 'negative' term comes from enterprise perspective, not independent perspective.  And if an agent doesn't get their dividend as expected, they could issue a regular claim.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -117,8 +117,13 @@ vf:Fulfillment  a           owl:Class ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The quantity that the economic event fulfilled towards the commitment." .
 
-vf:Claim  a           owl:Class ;
+vf:Claim  a                 owl:Class ;
         rdfs:label          "vf:Claim" ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "." .
+
+vf:Settlement  a            owl:Class ;
+        rdfs:label          "vf:Settlement" ;
         vs:term_status      "unstable" ;
         rdfs:comment        "." .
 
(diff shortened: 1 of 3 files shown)


bhaugen:
@fosterlynn  @elf-pavlik
> This is to replace the removed Reciprocity, which created Commitments from Economic Events. It was observed that Claims are different than Commitments because commitments can only be committed by the provider, while claims are initiated by the hopeful receiver.
> 

Claims are an instance of the REA Duality relationshp. When an event that is expected to be reciprocated occurs, a claim exists for the reciprocal event. Whether or not it is materialized in something like a invoice, which is a claim for payment.

> Model looks like this. Claim has most of the properties of I-C-E- flows. Claim basically references both the event(s) that triggered the claim and the event(s) that resolve the claim.
> claim
> 

I think "materializes" is the wrong word. "Triggers" is better, with arrow going from the event to the claim. (And likewise I think the "settles" arrow goes from an event to a claim.) The claim does not materialize the event, the event has already happened. And the event does not materialize the claim, but it does trigger a claim.

> Note REA people also talk about claims and 'negative claims'. Negative claims are things like issuing dividends. I haven't defined those here, not sure if we want to do that, or consider those commitments. My impression is the 'negative' term comes from enterprise perspective, not independent perspective. And if an agent doesn't get their dividend as expected, they could issue a regular claim.

Positive and negative claims are trading partner views of the same claim. I understand that VF is using the independent or network view, in which case neither of those words applies.

fosterlynn:
>I think "materializes" is the wrong word. "Triggers" is better

I'm fine with triggers.

>arrow going from the event to the claim

arrow direction is per data reference, not the logical flow of how it happens.... but it is wrong anyhow, because I'm not explicitly resolving the M:M between event and claim, oops.  

will fix model....

fosterlynn:
Upgraded model.  I included all the attributes and relationships (inside to make it not so messy) so those are more apparent for discussion's sake.
claim

Decided it was safe to trigger a claim from only one event.

release-doc-in-process/all_vf.TTL
```diff
@@ -250,25 +255,32 @@ vf:inScopeOf  a             owl:ObjectProperty ;
...
-#        rdfs:domain         vf:Reciprocity ;
-#        rdfs:range          vf:Commitment ;
-#        vs:term_status      "unstable" ;
-#        rdfs:comment        "References the commitment that was fully or partially created because of the economic event, often based on a prior agreement." .
+vf:settles  a               owl:ObjectProperty ;
+        rdfs:label          "settles" ;
+        rdfs:domain         vf:Settlement ;
+        rdfs:range          vf:Claim ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "References a claim that fully or partially settled by the economic event.
```
elf-pavlik: Missing closing `"` which causes travis build to fail. I can setup pre-commit hook to use https://www.npmjs.com/package/turtle-validator to see problems with turtle file right away.

fosterlynn: thanks, fixed... you don't have to set up a hook, I can get in the habit of checking the travis build

fosterlynn:
pushed a simple example, a la Sensorica - needs update when we figure out transfer

PR #492, opened 2019-04-14
Pavlik quantities option 

elf-pavlik:
Alternative to #478 

I will start adding other examples later today.

I decided to include what I miss in your examples, flows here try to meet all the requirements discussed in #471  Which means that any UC can have variant where reciprocity happens based on % but we still need to manage quantities of involved resource and duration, which both affect coverage of fulfillment.

### DIFF
--- a/examples/translate.yaml
+++ b/examples/translate.yaml
@@ -1,0 +1,222 @@
+# Example: Part of planning and doing a translation - one step, proofreading
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - alice: https://alice.example/
+    translators: https://translators.example/
+    lingo: https://lingovocab.example/#
+
+'@id': rgh:valueflows/valueflows/master/examples/translate.yaml
+'@graph':
+
+  # recipe (fragment: proofreading process)
+
+  - '@id': translators:c404cff5-19c7-453d-b0d8-d8e96055bd0b
+    '@type': RecipeProcess
+    skos:note: Proofread a translated blog post
+
+  - '@id': translators:13cf44e5-eaa1-4ad0-88d5-e2910a73f377
+    '@type': RecipeFlow
+    action: work 
+    classifiedAs: https://www.wikidata.org/wiki/Q834191 # proofreading
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': translators:9f112621-3f13-4f48-a3e2-fd2ca99493cd
+    '@type': RecipeFlow
+    action: accept
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': translators:c7897c39-7f05-4a5d-a487-80e130a2414b
+    '@type': RecipeFlow
+    action: improve 
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  # resources
+
+  - '@id': translators:e1721a61-cd47-4556-84b9-8b1b81da15bf
+    '@type': EconomicResource
+    name: Summer of Commoning - A Tour of French Communities and Projects
+    classifiedAs: https://www.wikidata.org/wiki/Q17928402 # blog post
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  # plan
+
+  - '@id': translators:02b39a30-3e04-4305-9656-7f261aa63c84
+    '@type': Process
+    name: Proofread Summer of Commoning
+    hasBeginning: 2018-12-07T08:00:00-5:00
+    hasEnd: 2018-12-09T08:00:00-5:0
+
+  - '@id': translators:a25500e0-0106-43cd-8cbb-e74779488835
+    '@type': Intent
+    skos:note: Proofreading a French to Spanish translation
+    action: work
+    classifiedAs: https://www.wikidata.org/wiki/Q834191 # proofreading
+    receiver: http://www.translators.example/
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    time:hasDuration:
+      time:unitType: time:unitHour
+      time:numericDuration: 3 # estimate based on 350 words / hour
+    lingo:wordCount:
+      qudt:unit: unit:Words
+      qudt:numericValue: 1025
+    reciprocityQuantity:
+      qudt:unit: unit:Words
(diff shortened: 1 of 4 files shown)


examples/use-kits.yaml
```diff
@@ -0,0 +1,116 @@
...
+    hasBeginning: 2019-03-15T14:30:00-5:00
+    hasEnd: 2019-03-15T17:00:00-5:00
+
+  - '@id': ouikit:cc5a46e2-b903-4ff5-8c65-587272aba071
+    '@type': Fulfillment
+    fulfills: ouikit:bd5806d6-4a36-45b9-b3b6-3e7d361a5a13
+    fulfilledby: ouikit:30dad10d-606e-4188-80a2-4a1cf2454f58
+    fulfillmentCovers:
+      - coverageProperty: involvesQuantity
+        coverageQuantity: 20
```
fosterlynn: Is this 20 significant as opposed to 29?  Or typo?

elf-pavlik: just typo

fosterlynn:
@elf-pavlik just to let you know I'm following your examples, and will be able to comment better after seeing something that handles non-involves, non-duration quantities, like work as words.

elf-pavlik:
I've just added translate example, please note two things:
* it takes advantage from topological temporal relations, if process beginning end change, flows still stay defined as during the process and don't have to update *after* and *before*, also *work* gets positioned after *accept* and before *improve* which sounds like reasonable order for those flows.
* while it also uses word count for *reciprocityQuantity*, nothing stops others to use % and still record as well:
  * number of people involved in work
  * duration of work
  * world count

examples/translate.yaml
```diff
@@ -0,0 +1,222 @@
...
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    time:hasDuration:
+      time:unitType: time:unitHour
+      time:numericDuration: 3 # estimate based on 350 words / hour
+    lingo:wordCount:
+      qudt:unit: unit:Words
+      qudt:numericValue: 1025
+    reciprocityQuantity:
```
fosterlynn: Are you thinking there will be a way to generate this part of the plan from the recipe?

Also, how to give people a way to create formulas for reciprocity and other purposes on all these fields?

fosterlynn: Oh, just saw your comment in the conversations thread, answer is there.

elf-pavlik: Approach with `scalingFactor` and `factorProperty` doesn't even assume properties from `vf:` namespace. I think implementations may require to use properties with range `qudt:QuantityValue` and `time:Duration`, other than that even meaning of the property shouldn't matter for scaling, `factorVariable` and `factorFormula` should make all the scaling explicit and independent from meaning of the property.

Since you suggest only one quantity, I guess dedicated `reciprocityQuantity` should work for reciprocity requirement. At the same time other quantities can address other requirements.

fosterlynn:
>it takes advantage from topological temporal relations, if process beginning end change, flows still stay defined as during the process and don't have to update after and before, also work gets positioned after accept and before improve which sounds like reasonable order for those flows.

That is interesting, and I see the potential advantage of flows being intervals.

>I didn't know how you intended to scale recipe so I didn't include some quantities on it, it would also generate process with no duration and work flow with no quantities and require adding them directly on the plan.

I presume people will write different logic to generate plans from recipes depending on their needs.  I think we should make it as easy as possible for them.  So I was looking for how these extra named quantities could be actually made workable.

fosterlynn:
Trying to think of how to get closer together on this, as we are as far apart as we ever were.  Not sure.  

One possibility:  having multiple quantities, but defined as vf quantities and all having range of QuantityValue.  Define as few as possible, define the properties clearly based on the same criteria (not mixing format, usage, functional definition).

(Durations needed for scheduling would use hasDuration, so I'm not including that in the quantity discussion.  I hope that gets one thing out of the mix.)

Another possibility: move to using resources that reflect the quantity definition directly, usage and work resources with calendar schedule, availability and possession type resources like for rental, etc.  That might solve a lot of it, although possibly not all of it, would have to work it through.

Not making any proposals, just exploring.

Another alternative, wait until we have actual users with actual use cases to further develop this.  As of now, one quantity has been totally fine.  (Where I have seen deficiencies in our use of dates in the older software, now corrected in VF.)

fosterlynn:
One specific comment, on lingo:wordcount.  If we compare the difficulties for users, vf:quantifiedAs requires users to choose their unit in existing software.  Your proposal seems like it involves users in a domain getting together to create and publish a vocabulary, include it in their software, etc., which sounds like a pretty big deal tbh. Even a big deal to document some directions for people on how to do this.  Am I missing something?

elf-pavlik:
> Your proposal seems like it involves users in a domain getting together to create and publish a vocabulary, include it in their software, etc., which sounds like a pretty big deal tbh. Even a big deal to document some directions for people on how to do this. Am I missing something?

I think we should make it clear that operational applications will need more than what `vf:` defines and recommends from other general vocabs. This applies specifically to various domain specific applications. I don't see it in scope of VF to define things like: *shoe size*, *word count*, *engine power*, etc.

> One possibility: having multiple quantities, but defined as vf quantities and all having range of QuantityValue.
> [...]
> (Durations needed for scheduling would use `time:Duration`, so I'm not including that in the quantity discussion. I hope that gets one thing out of the mix.)

I think we can define it simply as `time:Duration` for any duration quantities and `qudt:QuantityValue` for everything else. I see no reason why to guess what that duration might get used for, even if we also use it in something like *reciprocityQuantity* `time:Duration` will work just as fine as if one used QUDT duration units.

> Another possibility: move to using resources that reflect the quantity definition directly, usage and work resources with calendar schedule, availability and possession type resources like for rental, etc. That might solve a lot of it, although possibly not all of it, would have to work it through.

I think we should see how this would work with `ouikts`, I can see the ouikits as resource but using them looks like a flow to me.

> Trying to think of how to get closer together on this, as we are as far apart as we ever were. Not sure.

In this PR, for every single UC one can easily handle variant where *reciprocityQuantity* gets handled with % while other quantities get used for planning (which includes scheduling) and reports and analysis (eg. words per hour translated). Could you explain how in your PR #478 you would handle variants of included there UCs where reciprocity gets handled with % but one also needs to meet other requirements (planning with scheduling + reports and analysis). I don't think we can have any of requirements discussed in #471 mutually exclusive, to show that PR meets this criteria it should showcase that each flow can meet all the discussed requirements.

fosterlynn:
>I think we should make it clear that operational applications will need more than what vf: defines and recommends from other general vocabs. This applies specifically to various domain specific applications. I don't see it in scope of VF to define things like: shoe size, word count, engine power, etc.

That's why all of that sort of thing should be user definable and configurable.  I don't think any of that should be in a vocabulary that is an ontology.  A shared taxonomy, sure.  But "data", not "software".  (Conceptually.)

fosterlynn:

Issue #493, opened 2019-04-14
essepuntato spec generation not working


fosterlynn:
Get 'internal server error' when trying to run it directly in the form on their site.  Tried an older branch to see if it had anything to do with removing the external vocab pieces. webvowl works for our current version, although it looks a bit disjointed now.  Seems like the site is still supported, at least some of the examples linked do work, and others get a much more verbose error message.

http://www.essepuntato.it/lode/owlapi/https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/all_vf.TTL

elf-pavlik:
> http://www.essepuntato.it/lode/owlapi/https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/all_vf.TTL

This one seems to work or you see problems with it?

fosterlynn:
Worked for me yesterday, tried it this morning, got:

Proxy Error
The proxy server received an invalid response from an upstream server.
The proxy server could not handle the request GET /apps/LODE/extract.
Reason: Error reading from remote server
Apache/2.2.22 (Ubuntu) Server at eelst.cs.unibo.it Port 80

I'm guessing it is up and down on their side.  Hmmm.

fosterlynn:
closing since I did a PR to save snapshot, for now

fosterlynn:
From the essepuntato github:

>The online version is now usable at another URL, starting with https://w3id.org/lode, e.g. https://w3id.org/lode/owlapi/http://purl.org/spar/cito.

fosterlynn:
And another comment from https://github.com/essepuntato/LODE/issues/21#issuecomment-529179453

>FWIW my mate @nicholascar has cloned LODE in Python - an API is described here https://kurrawong.net/pylode-online

fosterlynn:
Changed to the new LODE location.

PR #495, opened 2019-04-16
snapshot of the essepunto spec, otherwise working only intermittently

fosterlynn:
@elf-pavlik help? I can't remember how to test this in gitbook (works directly from my local drive) - but I think there is a way to do that.  I'm thinking we'll need to take snapshots by just saving in the browser when we add things.  And see if it is stays up more than it is down.

Ref. https://github.com/valueflows/valueflows/issues/493 which I will close and just leave this open.


elf-pavlik:
I would prefer not to add such manual step which will complicate all PR changing the turtle file. Let's better look at changing to something else dynamic.

fosterlynn:
>I would prefer not to add such manual step which will complicate all PR changing the turtle file. Let's better look at changing to something else dynamic.

Well, I would prefer that too, of course, just takes research.  Don't see anything from a cursory search.  Will look more later.

fosterlynn:
Update: on the essepuntato github, says they are looking for other server space, don't know when or if it will be back. https://github.com/essepuntato/LODE/issues/21.  It talks about the web version, I'll check to see if there is something that can be used locally to create a snapshot, understanding that is not the best option.

Also, did a search, doesn't look super promising, but there are some tools listed here https://www.w3.org/2011/prov/wiki/Generating_HTML_documentation_of_OWL.

Another option is to investigating generating doc from the graphql reference, or a json-schema reference (which I'll get to at some point).

fosterlynn:
Resolved.

PR #497, opened 2019-04-18
transfer examples

elf-pavlik:
https://github.com/valueflows/valueflows/issues/270#issuecomment-483028351

> * transfer some currency from one agent's account to another agent's account (no fee)

Transfer of R&R only happens on *increment* flow. I will add more commits with other examples.

### DIFF
--- a/examples/book-checkout.yaml
+++ b/examples/book-checkout.yaml
@@ -1,0 +1,57 @@
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - library: https://library.example/
+    bob: https://bob.example/
+
+
+'@id': rgh:valueflows/valueflows/master/examples/book-checkout.yaml
+'@graph':
+
+  # resource before
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': library:f47064eb-7120-4b31-b882-770165901fe6
+    '@type': Agreement
+    skos:note: renting Valueflows cookbook to Bob
+
+  - '@id': library:54b814ee-62dc-40c1-bb96-f8582aa4f771
+    '@type': EconomicEvent
+    clauseOf: library:f47064eb-7120-4b31-b882-770165901fe6
+    action: decrement
+    involves: library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
+    transfer:
+      - rights: use
+        provider: https://library.example/
+        receipient: https://bob.example/
+
+  - '@id': bob:98c2bfeb-9c70-4801-896c-4646b975a7d9
+    '@type': Commitment
+    clauseOf: library:f47064eb-7120-4b31-b882-770165901fe6
+    action: increment
+    involves: library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    before:  2019-05-17T18:00:00-5:00
+    transfer:
+      - rights: use
+        provider: https://bob.example/
+        receipient: https://library.example/
+
+  # resource after
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 0
--- a/examples/currency-transfer.yaml
+++ b/examples/currency-transfer.yaml
@@ -3,69 +3,175 @@
   - https://git.io/vf-examples-jsonld-context
   - bob: https://bob.example/
     alice: https://alice.example/
-    lets: https://lets.example/
+    abank: https://abank.example/
+    bbank: https://bbank.example/
+    notary: https://notary.example/
 
 '@id': rgh:valueflows/valueflows/master/examples/currency-transfer.yaml
 '@graph':
 
   # accounts before
 
-  - '@id': lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+  - '@id': abank:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
     '@type': EconomicResource
-    skos:note: Alice's account 
+    skos:note: shared account Alice has some credits on
     classifiedAs: wd:Q179941
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: -200
+      qudt:numericValue: 200
+    share:
+      - owner: https://alice.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 150
+      - owner: https://zoe.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 50
 
-  - '@id': lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+  - '@id': bbank:3be5259d-10f0-431c-9fec-9c0c15a461d3
     '@type': EconomicResource
-    skos:note: Bob's account 
+    skos:note: shared account Bob has some credits on
     classifiedAs: wd:Q179941
     currentQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 500
+    share:
+      - owner: https://bob.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 200
+      - owner: https://sam.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 300
 
-  - '@id': lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+  - '@id': abank:e0caf1e3-4a9b-4927-a19c-3f681d234529
     '@type': Process
-    skos:note: moving credits between accounts
+    skos:note: debit Alice's credits on account
 
-  - '@id': lets:0468e9b9-4d54-4eee-9d3f-bc76db11710a
+  - '@id': abank:0468e9b9-4d54-4eee-9d3f-bc76db11710a
     '@type': EconomicEvent
-    action: decrement
-    inputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
-    involves: lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
-    involvesQuantity:
+    action: split
+    inputOf: abank:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    resourceInventoriedAs: abank:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    provider: https://alice.example/
+    recipient: https://alice.example/
+    actsOnQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 100
 
-  - '@id': lets:33f11508-92b3-4095-9e03-6678cddf7adf
+  - '@id': abank:58c36336-09fa-4be2-b1ac-d26ea91bd982
     '@type': EconomicEvent
-    action: increment
-    outputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
-    involves: lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
-    involvesQuantity:
+    action: merge
+    outputOf: abank:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    resourceInventoriedAs: urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    provider: https://alice.example/
+    recipient: https://alice.example/
+    actsOnQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 100
-    transfer:
-      - rights: control
-        provider: https://alice.example/
-        recipient: https://bob.example/
 
-  # accounts after
-
-  - '@id': lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+  - '@id': abank:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
     '@type': EconomicResource
     skos:note: Alice's account 
     classifiedAs: wd:Q179941
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: -300
+      qudt:numericValue: 100
+    share:
+      - owner: https://alice.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 150
+      - owner: https://zoe.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 50
 
-  - '@id': lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+  - '@id': urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    '@type': EconomicResource
+    skos:note: credits in transit
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+    share:
+      - owner: https://alice.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 100
+
+  # actual ownership transfer happens only in this flow
+  - '@id': notary:da8f3360-3a6d-4ea8-b57f-4d4236fe2c3e
+    '@type': EconomicEvent
+    action: own
+    provider: https://alice.example/
+    recipient: https://bob.example/
+    resourceInventoriedAs: urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    actsOnQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+  - '@id': urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    '@type': EconomicResource
+    skos:note: credits in transit
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+    share:
+      - owner: https://bob.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 100
+
+  - '@id': bbank:57f01dbe-fd78-40d0-9278-60005debfba6
+    '@type': Process
+    skos:note: credit Bob's account
+
+  - '@id': bbank:33f11508-92b3-4095-9e03-6678cddf7adf
+    '@type': EconomicEvent
+    action: split
+    inputOf: bbank:57f01dbe-fd78-40d0-9278-60005debfba6
+    resourceInventoriedAs: urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    provider: https://bob.example/
+    recipient: https://bob.example/
+    actsOnQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+  - '@id': bbank:33f11508-92b3-4095-9e03-6678cddf7adf
+    '@type': EconomicEvent
+    action: merge
+    outputOf: bbank:57f01dbe-fd78-40d0-9278-60005debfba6
+    resourceInventoriedAs: bbank:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    provider: https://bob.example/
+    recipient: https://bob.example/
+    actsOnQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+  - '@id': bbank:3be5259d-10f0-431c-9fec-9c0c15a461d3
     '@type': EconomicResource
     skos:note: Bob's account 
     classifiedAs: wd:Q179941
     currentQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 600
+
+  - '@id': urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    '@type': EconomicResource
+    skos:note: credits in transit
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:numericValue: 0
+    share:
+      - owner: https://bob.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 300
+      - owner: https://sam.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 300
--- a/examples/currency-transfer.yaml
+++ b/examples/currency-transfer.yaml
@@ -159,13 +159,6 @@
     currentQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 600
-
-  - '@id': urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
-    '@type': EconomicResource
-    skos:note: credits in transit
-    classifiedAs: wd:Q179941
-    currentQuantity:
-      qudt:numericValue: 0
     share:
       - owner: https://bob.example/
         quantity:
@@ -175,3 +168,10 @@
         quantity:
           qudt:unit: unit:Number
           qudt:numericValue: 300
+
+  - '@id': urn:uuid:9f5181db-7bd2-4ddf-92ff-10f0e8fad844
+    '@type': EconomicResource
+    skos:note: credits in transit
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:numericValue: 0
--- a/examples/currency-transfer.yaml
+++ b/examples/currency-transfer.yaml
@@ -1,0 +1,71 @@
+
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - bob: https://bob.example/
+    alice: https://alice.example/
+    lets: https://lets.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/currency-transfer.yaml
+'@graph':
+
+  # accounts before
+
+  - '@id': lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    skos:note: Alice's account 
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: -200
+
+  - '@id': lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    '@type': EconomicResource
+    skos:note: Bob's account 
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 500
+
+  - '@id': lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    '@type': Process
+    skos:note: moving credits between accounts
+
+  - '@id': lets:0468e9b9-4d54-4eee-9d3f-bc76db11710a
+    '@type': EconomicEvent
+    action: decrement
+    inputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    involves: lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+  - '@id': lets:33f11508-92b3-4095-9e03-6678cddf7adf
+    '@type': EconomicEvent
+    action: increment
+    outputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    involves: lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+    transfer:
+      - rights: control
+        provider: https://alice.example/
+        recipient: https://bob.example/
+
+  # accounts after
+
+  - '@id': lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    skos:note: Alice's account 
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: -300
+
+  - '@id': lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    '@type': EconomicResource
+    skos:note: Bob's account 
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 600
--- a/examples/transfer-inventory.yaml
+++ b/examples/transfer-inventory.yaml
@@ -1,0 +1,219 @@
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - bob: https://bob.example/
+    alice: https://alice.example/
+    carol: https://carol.example/
+    dan: https://dan.example/
+
+'@id': rgh:valueflows/valueflows/master/examples/transfer-inventory.yaml
+'@graph':
+
+  # resource before
+
+  - '@id': alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    classifiedAs: wd:Q89 # apples
+    currentPossesion: https://alice.example/
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 230
+
+  # alice sells some of her apples at the farmer's market
+  # first she splits amount of apples
+
+  - '@id': alice:0468e9b9-4d54-4eee-9d3f-bc76db11710a
+    '@type': EconomicEvent
+    action: split
+    resourceInventoriedAs: alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 50
+
+  - '@id': alice:267a2f60-92e0-49c8-928b-601e89d5ddf9
+    '@type': EconomicEvent
+    action: merge
+    resourceInventoriedAs: alice:e55414b6-0b4d-42a5-b86c-d85d95979732
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 50
+
+  # then she transfer rights to them
+  - '@id': alice:267a2f60-92e0-49c8-928b-601e89d5ddf9
+    '@type': EconomicEvent
+    action:
+      - own
+      - possess
+    provider: https://alice.example/
+    receiver: https://dan.example/
+    resourceInventoriedAs: alice:e55414b6-0b4d-42a5-b86c-d85d95979732
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 50
+
+  - '@id': alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    classifiedAs: wd:Q89 # apples
+    currentPossesion: https://alice.example/
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 180
+
+  - '@id': alice:e55414b6-0b4d-42a5-b86c-d85d95979732
+    '@type': EconomicResource
+    classifiedAs: wd:Q89 # apples
+    currentPossesion: https://dan.example/
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 50
+
+  # alice loans a basket of apples to her friend to use in a display for a few days
+
+  - '@id': alice:33f11508-92b3-4095-9e03-6678cddf7adf
+    '@type': EconomicEvent
+    action: split
+    resourceInventoriedAs: alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 10
+
+  - '@id': alice:d5062396-c17f-4880-9cb4-d25e74fe1f63
+    '@type': EconomicEvent
+    action: merge
+    resourceInventoriedAs: alice:01595d68-e3d6-40ab-a77d-3a48b64fc5ba
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 10
+
+  - '@id': alice:e899998e-36ca-4cd6-9c54-655346ac6102
+    '@type': EconomicEvent
+    action: possess
+    provider: https://alice.example/
+    receiver: https://carol.example/
+    resourceInventoriedAs: alice:01595d68-e3d6-40ab-a77d-3a48b64fc5ba
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 10
+
+  - '@id': carol:d5062396-c17f-4880-9cb4-d25e74fe1f63
+    '@type': Commitment
+    action: possess
+    provider: https://carol.example/
+    receiver: https://alice.example/
+    resourceInventoriedAs: alice:01595d68-e3d6-40ab-a77d-3a48b64fc5ba
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 10
+
+  - '@id': alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    classifiedAs: wd:Q89 # apples
+    currentPossesion: https://alice.example/
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 170
+
+  - '@id': alice:01595d68-e3d6-40ab-a77d-3a48b64fc5ba
+    '@type': EconomicResource
+    classifiedAs: wd:Q89 # apples
+    currentPossesion: https://carol.example/
+    currentQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 10
+
+  # alice sells some apples FOB destination, and has them shipped (they are enroute)
+
+  - '@id': alice:469a7adb-c7c0-44d2-99a2-df4e4eff97a3
+    '@type': EconomicEvent
+    action: split
+    resourceInventoriedAs: alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    actsOnQuantity:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 100
+
+  - '@id': alice:12c25554-1d9f-47e7-9f23-038b8c3b1246
+    '@type': EconomicEvent
+    action: merge
+    resourceInventoriedAs: alice:7a3b3eda-bef6-4b15-bcec-8f0b172e4051
+    actsOnQuantity:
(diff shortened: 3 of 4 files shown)


elf-pavlik:
Last commit has alternative to #388 without trying to mange tracking identifiers which looks like requirement completely independent from transfers.

elf-pavlik:
> * checkout a library book (but no transportation - checkin to hotel if you prefer)

Last commit includes that case

> * transportation FOB source, 3rd party shipper
> * transportation FOB destination, 3rd party shipper

For those I'll update #386 which has 3rd party shipper variant https://github.com/valueflows/valueflows/pull/386/files#diff-ec6dd08b70d0091f4002433eceb7303eR163

FOB origin will provide interesting case where on *load* just 'possession' change Alice -> Claudia and on *unload* we have two transfers: 'possesion' Claudia -> Bob & 'control' Alice -> Bob

FOB destination will just differ that 'control' Alice -> Bob happens on *load*

elf-pavlik:
I have changed shipping to *FOB destination* in this commit: https://github.com/valueflows/valueflows/pull/386/commits/8fab722c19ea8c8175a3996aea70f402fcc96f6b

For the sake of transfer related discussion I'll include it here in comment:

```yaml

# Bob receives 100kg of apples from Alice and Claudia transports them (FOB destination)

  - '@id': claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    '@type': Process

  - '@id': claudia:fd399b37-0740-4a68-a184-1e655021ca21
    '@type': EconomicEvent
    inputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: load
    involves: alice:c7897c39-7f05-4a5d-a487-80e130a2414b
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights:
        - load
        - unload
        receiver: https://claudia.example/
    
  - '@id': claudia:57f1c1d0-432e-4bfa-9d32-002b8955a708
    '@type': EconomicEvent
    outputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: unload
    involves: bob:a3246484-e862-4a12-b729-23986da5c033
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights: control
        receiver: https://bob.example/
      - rights:
        - load
        - unload
        provider: https://claudia.example/
        receiver: https://bob.example/
```
This example assumes that Claudia looses *load* and *unload* rights after transferring them to Bob. Possibly we may find need to distinguish between just giving rights and giving and loosing oneself. I imagine that some rights one can give without loosing them oneself eg. *use*.

fosterlynn:
Thanks @elf-pavlik !

From one of your examples:
```
    tranfer:
      - rights: control
        receiver: https://bob.example/
      - rights:
        - load
        - unload
        provider: https://claudia.example/
        receiver: https://bob.example/
```
I'm wondering what vf:transfer refers to?  Like what are these 2 items?  Do they have an owl:Class?  Or, what would the all_vf.TTL changes look like?

Another possible property: vf:under?  To further define the rights and responsibilities.

fosterlynn:
Another question: What would a planned exchange agreement look like with say the currency transfer (say the LETS example) as one side and one of the FOB ownership transfers on the other?

elf-pavlik:
> I'm wondering what vf:transfer refers to? Like what are these 2 items? Do they have an owl:Class? Or, what would the all_vf.TTL changes look like?

I didn't give them any `@type` since I didn't see how would it matter for this example. We can always come up with name for the `rdf:Class` later on.

> Another possible property: vf:under? To further define the rights and responsibilities.

Since we have `vf:under rdfs:range vf:Agreement .` it would imply that those instances I didn't use any class for have `'@type': Agreement`. In some ways it makes sense since it only represents very agreement between agents, we could consider that.

> Although I'm curious how you would do it for a load on FOB source in the example here, because I think it would be the wrong resource.

I don't understand what you mean by 'wrong resource'? Transfer applies to the small-r resource not the inventoried resource. Based on action in the flow and other flows in I-P-O chain one should have a way in what (if any) inventoried resource that transferred small-r resource ends up. In case of transfer on *load* it would end up in a resource with matching *unload*,  which can happen next day or next week and transfer already happened long time ago.

fosterlynn:
I did some thinking about this last night, and hope to write some ideas up this afternoon, may make a little progress this morning.  They involve adding a resource to the "thing" that is the range of vf:transfer in your examples.  Basically I'm trying to use your pattern but get all the resources tied up so agents know what is in their namespace/inventory.

>Transfer applies to the small-r resource not the inventoried resource.

Not sure I totally understand that.  It applies to R&R for the inventoried (or not inventoried) resource.  I think a small-r resource could be inventoried or not, it covers all of "resource-ness".  Either that or we are creating a different resource for the R&R, which is another interesting avenue to try to detail out possibly.

examples/book-checkout.yaml
```diff
@@ -0,0 +1,57 @@
...
+        provider: https://bob.example/
+        receipient: https://library.example/
+
+  # resource after
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 0
```
fosterlynn: Interestingly, I think the library wants to still know that it *owns* a certain number of the VF Cookbook, even if one or more is checked out and therefore not available.

examples/currency-transfer.yaml
```diff
@@ -0,0 +1,71 @@
...
+    action: increment
+    outputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    involves: lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    involvesQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+    transfer:
+      - rights: control
+        provider: https://alice.example/
+        recipient: https://bob.example/
```
fosterlynn: This is an example place where it is not clear what Alice is giving up some rights to.  SInce this is simple, you could figure it out, and a machine could figure it out, although it would be a pain.  But if there were a service created, and a transfer fee, and whatever, it would start to become even more of a pain.

examples/book-checkout.yaml
fosterlynn: I tentatively think that the currentQuantity should refer to total "controlled", but maybe we need something else too - or maybe the availability can be calculated for now or later in the same way, by looking at commitments and events?

elf-pavlik: They have commitment that this resource will get incremented again, they also should have track that they at some point in the past received *control* rights to that quantity of that resource and never transfer that right to anyone else so they still have that right.

examples/currency-transfer.yaml
elf-pavlik: I think we only transfer R&R to products not services (including usage and work). In my examples transfer always happens over `involvesQuantity` of a resource specified either by `involves` or classification or specification.

fosterlynn:
Some conclusions I have of the moment:
* At the instant of transfer of full R&R, the information of 2 resources (one decremented and one incremented) needs to be very easy to find in the data.  (Unless the resource is not inventoried at all so there is nothing to decrement or increment.)
* There might be record of different REA resources for the same "thing" at the same time, which would involve different R&R and maybe realities (like possession automatically implies there are some R&R around).

Taking your example above and adding to the vf:transfer parts, first attempt for discussion:
```# Bob receives 100kg of apples from Alice and Claudia transports them (FOB destination)

  - '@id': claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    '@type': Process

  - '@id': claudia:fd399b37-0740-4a68-a184-1e655021ca21
    '@type': EconomicEvent
    inputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: load
    involves: alice:c7897c39-7f05-4a5d-a487-80e130a2414b
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights:
        - load
        - unload
        receiver: https://claudia.example/
        resourceyThing: claudia:sssss # the receiver resource, gives the ability to keep the shipping resource list
        #(do we need a time? can a load event take more than an instant?)

  - '@id': claudia:57f1c1d0-432e-4bfa-9d32-002b8955a708
    '@type': EconomicEvent
    outputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: unload
    involves: bob:a3246484-e862-4a12-b729-23986da5c033
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights: control
        receiver: https://bob.example/
        resourceyThing: alice:c7897c39-7f05-4a5d-a487-80e130a2414b # the provider resource
        #(do we need a time?)
      - rights:
        - load
        - unload
        provider: https://claudia.example/
        receiver: https://bob.example/
        resourceyThing: claudia:sssss # the provider resource
        #(do we need a time?)
```

fosterlynn:
Another example in the same experiment:
```
  - '@id': lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
    '@type': Process
    skos:note: moving credits between accounts

  - '@id': lets:0468e9b9-4d54-4eee-9d3f-bc76db11710a
    '@type': EconomicEvent
    action: issue # give?
    inputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
    involves: lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
    involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 100
    time:inXSDDateTimeStamp: 2018-10-14T8:24:34-0:00

  - '@id': lets:33f11508-92b3-4095-9e03-6678cddf7adf
    '@type': EconomicEvent
    action: receive # take?
    outputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
    involves: lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
    involvesQuantity:
      qudt:unit: unit:Number
      qudt:numericValue: 100
    time:inXSDDateTimeStamp: 2018-10-14T8:24:34-0:00
    transfer:
      - rights: control
        recipient: https://bob.example/
        resourceyThing: lets:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead #provider resource to decrement
```

examples/currency-transfer.yaml
fosterlynn: It doesn't matter, just trying to say it is not straight-forward to find the other resource in all cases.  Maybe the wrong place to make the comment since this is probably the simplest example.

examples/book-checkout.yaml
fosterlynn: Yes, always an inventoried resource can be derived, it does not have to be instantiated at all.  But since we so far always instantiate it in practice, and have defined VF that way too, we need to define the currentQuantity carefully.  Do you think it should be the quantity you can see and touch always for that kind if resource, irrespective of if the agent actually "owns" all of it?

fosterlynn:
Trying to get alice her inventory in transit.
```yaml
# Bob receives 100kg of apples from Alice and Claudia transports them (FOB destination)

  - '@id': alice:533f6e56-6c7d-4a5b-b9c9-1a8adafd8969
    '@type': Process

  - '@id': alice:fd399b37-0740-4a68-a184-1e655021ca21
    '@type': EconomicEvent
    inputOf: alice:533f6e56-6c7d-4a5b-b9c9-1a8adafd8969
    action: issue
    involves: alice:c7897c39-7f05-4a5d-a487-80e130a2414b
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100

  - '@id': alice:fd399b37-0740-4a68-a184-1e655021ca21
    '@type': EconomicEvent
    inputOf: alice:533f6e56-6c7d-4a5b-b9c9-1a8adafd8969
    action: receive
    involves: alice:the-inventory-in-transit-still-owned-by-alice
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100

  - '@id': claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    '@type': Process

  - '@id': claudia:fd399b37-0740-4a68-a184-1e655021ca21
    '@type': EconomicEvent
    inputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: load
    involves: alice:the-inventory-in-transit-still-owned-by-alice
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights:
        - load
        - unload
        receiver: https://claudia.example/
        resourceyThing: claudia:sssss # the receiver resource, gives the ability to keep the shipping resource list
        #(do we need a time? can a load event take more than an instant?)

  - '@id': claudia:57f1c1d0-432e-4bfa-9d32-002b8955a708
    '@type': EconomicEvent
    outputOf: claudia:633f6e56-6c7d-4a5b-b9c9-1a8adafd8960
    action: unload
    involves: bob:a3246484-e862-4a12-b729-23986da5c033
    involvesQuantity:
      qudt:unit: unit:Kilogram
      qudt:numericValue: 100
    tranfer:
      - rights: control
        receiver: https://bob.example/
        resourceyThing: alice:the-inventory-in-transit-still-owned-by-alice # the provider resource
        #(do we need a time?)
      - rights:
        - load
        - unload
        provider: https://claudia.example/
        receiver: https://bob.example/
        resourceyThing: claudia:sssss # the provider resource
        #(do we need a time?)
```

fosterlynn:
Something I noticed in the above example: The load doesn't want to decrement alice's in-transit resource.  In general, I am coming to the conclusion that load and unload don't decrement or increment anything.  (Although this relates to the discussion of what decrement and increment mean too, if it is a physical thing or a rights/availability/accounting thing.)

>Since we have vf:under rdfs:range vf:Agreement . it would imply that those instances I didn't use any class for have '@type': Agreement. In some ways it makes sense since it only represents very specific agreement between two agents, we could consider that.

That doesn't seem right to me.  In the model in general, there can be a commitment to transfer something (part of an agreement), and there can be (or not be) an actual transfer of something per the agreement.  (Something might happen in the meantime that precludes the actual transfer, since we still need R&R for something real and available.)

Maybe it is a vf:Transfer (not to be confused with the old one).

A separate idea not related to the name: Maybe it has to reference 2 process events as trigger, instead of one?  Then it really only needs the R&R part, although the time instant might get difficult to manage.  Or have vf:transfer on the 2 events referencing the same thing, if you like that direction better.  Just brainstorming, I'm not sure I like this really.

elf-pavlik:
I think we struggle because we attempt to model something without clearly define requirements. Once we clarify how we expect to use data we try to model we will have better way to see how different approaches meet those requirements.

Let's maybe take a step back and start more rigorous process

1. Select Use Cases - we already have few re-ocurring ones, we can also make variants of them
2. Define Requirements
3. Write YAML for Use Cases with explanation how it meets all the requirements, or why it meets some but can't meet other ones.

> Maybe it is a vf:Transfer (not to be confused with the old one).

I really would leave naming things as last formality and first focus on functional requirements.


--------------------

I have impression that we still try to make *ownership* a special case which requires moving resource between inventories. While for example possession doesn't move resource between inventories, it still stays inventoried by the owner.
I don't say that we should or shouldn't do it this way, I just see that we can't have it both ways, suggest that we want to model all different R&R the same way but still have special treatment for *ownership*.

examples/book-checkout.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': library:f47064eb-7120-4b31-b882-770165901fe6
+    '@type': Agreement
+    skos:note: renting Valueflows cookbook to Bob
```
fosterlynn: This is interesting, missed the Agreement earlier.  Curious how would you combine that with the other side of the transfer (the increment on bob's side when he checks out the book), which would have a Process?

elf-pavlik: I don't know if we need to assume any 'other side'. Bob can just have that commitment but don't keep an inventory. Also if inclusion in inventory implies ownership, Bob might just increment the same resource `library:21f361a6-2375-46bb-b192-c21b5ba833bf` at his place. I wouldn't see it as some 'other side' of the transfer of *use* rights to the book, R&R already happened on check-out from library, recording now location of the same book on his bookshelf doesn't have anything to do with transfer any more. He could as well just keep it in his backpack and return couple of days later, fulfilling the commitments and transferring the *use* rights back to library but never checking it in and out again from anywhere else.

fosterlynn: No, we don't need to assume it of course.  But we also need to know how to handle it if it is needed, so I'd like to "assume" it for purposes of at least one example.  And in particular, the "mostly duplication" of the give/receive or decrement/increment in a process that is basically the same event content with a different agent's resource, I think we need to experience first hand to see if it is acceptable.

fosterlynn:
I think the main problem to solve with this transfer model is that transfers by nature involve 2 resources in some way, even if you are trying to decrement/increment in the events and not the transfers.  Both agents need to know exactly when something was transferred from/to someone else to know what they are responsible for, when.  Like do they have to report that inventory in transit, or if something happens to the shipment, do they have to coordinate with the shipper on that.

Maybe there is a way to solve that with this basic model?  (I'm still working on the R&R-as-resources possibility for now, but don't know how that will work out.)

fosterlynn:
Regarding using actions for rights, it does occur to me that `load` and `unload` are actually more responsibilities than rights in this case.  The shipper is doing that as part of providing a service, and doesn't have much choice in the matter for themselves.  So maybe there is a way to use actions for both rights and responsibilities?  Would have to look in some more detail.

elf-pavlik:
> I think the main problem to solve with this transfer model is that transfers by nature involve 2 resources in some way

Do you mean *ownerhisp* transfer, which one could model as moving physical resource between two different inventories. Or you suggest that transfer of any rights involves 2 resources, even if you don't model those rights as resources themselves?

fosterlynn:
>> I think the main problem to solve with this transfer model is that transfers by nature involve 2 resources in some way

>Do you mean ownerhisp transfer, which one could model as moving physical resource between two different inventories. Or you suggest that transfer of any rights involves 2 resources, even if you don't model those rights as resources themselves?

No I mean that any transfer of any full or partial rights to a resource needs to be clearly known by both agents when it happens (an instant), and what resource it happened to (however currently identified by the agent).  It affects not only ownership but availability, etc.

elf-pavlik:
Both agents need to have access to records, possibly even none of them should be able to change those records. Still besides accessing the information in some specific way, I don't see where you see need for 2 resources involved? Transfer could possibly happen over quantity of not inventoried resource, also it could happen over quantity of specific inventoried resource. It seems to me that two resources come into play in your #498 where you have something like 'rights accounts' which you represent as resources, in that case you have two accounts between which you transfer the rights. The need for two resources - rights accounts seems specific to that specific approach to modeling it.

fosterlynn:
> I don't see where you see need for 2 resources involved? Transfer could possibly happen over quantity of not inventoried resource, also it could happen over quantity of specific inventoried resource.

Of course inventoried resources are not always there.  But that is always the case for any kinds of events.  So I don't see why it is relevant here, since sometimes there are inventoried resources.  Transfers have always had 2 resources involved, in the sense of one agent's resource and another agent's resource.  Don't both agents need to know about transfer of any kind of rights?  Not just that a resource of that classification has been decremented for an unknown reason sometime in the past, which connects in a value chain, but which may not be definitively implied to be the resource transferred if there is any complexity to the graph.

>where you have something like 'rights accounts' which you represent as resources, in that case you have two accounts between which you transfer the rights. The need for two resources - rights accounts seems specific to that specific approach to modeling it.

I don't think so - transfers have ever and always been about 2 agents/resources, in all of our iterations and everyone else's in REA.  (Of course with the usual exceptions, not inventoried, possibly some serialized resources, etc.)

elf-pavlik:
> Transfers have always had 2 resources involved, in the sense of one agent's resource and another agent's resource.

I think this only applies for 'transfer of ownership', where ownership gets implied by which inventory resource gets included in and to transfer ownership one has to remove resource from one inventory and add to another inventory.

If you transfer *use* rights to a 3d printer, where do you have two resources if you don't model those rights as resources themselves?

fosterlynn:
>> Transfers have always had 2 resources involved, in the sense of one agent's resource and another agent's resource.

>I think this only applies for 'transfer of ownership', where ownership gets implied by which inventory resource gets included in and to transfer ownership one has to remove resource from one inventory and add to another inventory.

Ok I see your point that there is a difference, but agents still need to know everything about partial rights transfers even if not decremented out of inventory because of availability.  And even if using this model with transfer info attached to a load or unload for example, not just if rights are treated as resources.

By the way, I do like a lot of things about this model, but see that one as needed resolution to make it work.

elf-pavlik:
I would like to redo the *book checkout* example. I see it very related to all the *reservation* / *use* related use cases. What I find interesting that while checkout and checking allow us to track current quantity of a resource, the 'value flow' (meaning use value) relates to the whole *rental period*.
If we would have rental as part of the exchange, we would exchange the whole period of rental for something, not *check-in* and *check-out*. The whole rental period would have *provider* and *receiver*. Thinking in terms of *providing (use) value*, it doesn't happen on *check-in* and *check-out*, it happens during the period between them.

I have impression that we actually might need to jump into composition. This whole lending / rental could probably work as *service*. When we think about transportation, agents also provide & receive transportation *service*. Specific *load* and *unload* of a resource at locations defines that service. Exchange wouldn't happen on *load* and/or *unload* but on the service. Also thinking in terms of *providing and receiving value*, it doesn't really happen on neither *load* or *unload*, it happens on the completed service.

fosterlynn:
Interesting thought process.  Yes, worth looking into further in detail.

>This whole lending / rental could probably work as service. When we think about transportation, agents also provide & receive transportation service. Specific load and unload of a resource at locations defines that service.

Yes.  But I'm not sure that the load and unload "define" the service.  There will be other inputs needed to provide the service also.  I picture the service being an output of the transportation process that has the load and unload as well as other inputs; and may even be a series of processes that are legs in the transportation, but only one service will come out in the end that can be exchanged.

Renting probably the same in some way, but not as clear to me.

Currency transfer I think you were right when you said it has a give, a take, and also creates a service, which can be exchanged for a fee.

Anyhow, I think you're right, we need some detailed yaml work.  I'll work on that too, starting from the POA perspective, and we can see what it all looks like.

fosterlynn:
To make sure I understand the implications to the model from your latest example, some questions:
1. So always resources of the same specification and scope (ownership?) that are possessed (or soon will be possessed) by different agents must be separated into another resource, right?  There is only one possessor per resource?
2. You can't use your transfer related actions (possess, own) in combination with another process related action?  Like combining `load` with `possess`....
3. Is there ever a time you would include something like 'right to use', same as 'reserve' possibly?  If so, how would that work?

elf-pavlik:
> So always resources of the same specification and scope (ownership?) that are possessed (or soon will be possessed) by different agents must be separated into another resource, right? There is only one possessor per resource?

Yes, since they supposed to represent tangible physical resources, different agents can own it but only one agent can posses it at any given time.

> You can't use your transfer related actions (possess, own) in combination with another process related action? Like combining load with possess....

Actually I see them totally duplicate, we should consider having them combined like:

```yaml
action:
  - load
  - possess
```

Since all the rest of the information stays exactly the same.

> Is there ever a time you would include something like 'right to use', same as 'reserve' possibly? If so, how would that work?

In various conversations I think we realized that we don't need to overgeneralize 'rights and responsibilities'. It looks like we have 3 pragmatic requirements
* full control (aka. ownership)
* possession
* rights to use

My last example only deals with first two, 'rights to use' IMO need further clarification of use cases and their requirements. When do we have 'exclusive usage rights' and multiple agents, it seems to require managing a schedule. From schedule milestone I see it doable using intents and commitments which represent plans. We also may have case of 'rights to use' with some quotas, which I would try to represent with Commitments which later get refined to keep clear schedule #464 . Last case which seems like 'pure rights', deals with *non-exclusive* 'rights to use', like with a laundry machine of our neighbors which gets used sporadically enough that without keeping schedule one doesn't need to worry about conflicts. While all the other *exclusive rights* UCs seem to require intents and commitments to *reserve* - block availability on the schedule or at least assign some quota, we don't seem to have way to handle that non-exclusive rights which don't require schedule.

I think we should stop talking about 'rights to use' in general and always discuss it around one of the UCs which has different requirement and provides different variant of 'rights to use'.

fosterlynn:
>I think we should stop talking about 'rights to use' in general and always discuss it around one of the UCs which has different requirement and provides different variant of 'rights to use'.

I'd like to define it as exclusive rights to use, with some specific times.  And yes, it will often involve commitments and schedules.  Or someone could also just check out something by dropping in, like in a library.  But in any case, seems like the non-exclusive use case and the non-scheduled quota use case is more around agreements, not transfers, if it is even in scope of VF?

If the name needs to be clarified to define it that way, we can try to think of something.  Although I've spent a lot of time on this name, not finding any one word one that is adequate.

fosterlynn:
>Yes, since they supposed to represent tangible physical resources, different agents can own it but only one agent can posses it at any given time.

And you figure out who owns it by looking at the past events?

fosterlynn:
Coming back over to this issue.  I hadn't fully appreciated that you had gotten the transfers down to one event.  (It's obvious, I just hadn't considered the implications.)  And I still have questions on how continuing to use just the reference to the provider's resource, or urn:uuid, would work.  But still a nice outcome if we can figure something out, including the possibility of using just the logical identifier fields.

But.... and addressing this to myself more than you because I was pushing it..... then we need to figure out how to do things like transfer fees.  Or other inputs and outputs that we have considered as useful to attach to a transfer process. I'll give this a try in either a new PR or comments in my POA one.

elf-pavlik:
> I'd like to define it as exclusive rights to use, with some specific times. And yes, it will often involve commitments and schedules. Or someone could also just check out something by dropping in, like in a library.

When I think here about agents valuing what they receive and provide that *exclusive availability*. I think it works pretty much what we discussed so far with *use*, with the only difference that it doesn't mean active use but a reservation. I think we should look at some agreements that involve reservation and incorporate reciprocity, I think the way agents value (verb) that reservation will stay proportional to its duration, for that I think just like *use* the flow should capture the temporal factor. We could revise our discussions from scheduling milestone on *reserve* - one flow vs. *check-in* and *check-out* two flows.
It seems to me that Commitment with verb like *reserve* could work here, later it gets gets fulfilled with event meaning *currently reserved* for it duration. While something stays *reserved*, possession can change, location can change etc. We should consider what happens when resource with *reserve* commitments gets split and merged.
If we use Commitments, we get to another interesting case, in some cases agent referenced as *recipient* should stay able to 'transfer' that right (of receiving something specific they value), to another agent. Similar agent referenced as *provider* may need to 'transfer' that responsibility of providing something specific to another agent. For example - commitment to use generic tool like an electric drill, both agents (providing and receiving) would often don't mind 'transferring' responsibility of providing and rights of receiving to someone else. 


> But.... and addressing this to myself more than you because I was pushing it..... then we need to figure out how to do things like transfer fees. Or other inputs and outputs that we have considered as useful to attach to a transfer process.

In case of 'pure ownership transfer', which doesn't increment or decrement the physical resource, it only acts as digital record, I don't think we will need notary services which will take fees for creating each record, but in that case they may have the same fees for creating commitments and agreement records, so not transfer specific. When we have process moving quantity between two different physical resources, we have transportation which would have service based on *load* and *unload* of the resource. It probably doesn't care about potential split & marge at origin and potential split & merge at destination. Another case of moving quantity of currency (virtual resource) between accounts, we again have transportation like situation which would have something to represent *withdraw* and *deposit*. I don't see it as 'pure ownership transfer' which on its own doesn't increment or decrement resource it involves. Service here, often part of an agreement incorporating reciprocity, would not relate to 'pure ownership transfer' but to 'virtual transportation' service.

fosterlynn:
>> Yes, since they supposed to represent tangible physical resources, different agents can own it but only one agent can posses it at any given time.

>And you figure out who owns it by looking at the past events?

Never mind, I see the answer in the POA issue.

fosterlynn:
>In case of 'pure ownership transfer', which doesn't increment or decrement the physical resource, it only acts as digital record, I don't think we will need notary services which will take fees for creating each record,

>Another case of moving quantity of currency (virtual resource) between accounts, we again have transportation like situation which would have something to represent withdraw and deposit. I don't see it as 'pure ownership transfer' which on its own doesn't increment or decrement resource it involves. 

What's the difference between those?  What is a more "pure" transfer than currency transfers?  It is only about digital records.

Or do we disagree that you need some kind of economic event on the observation layer to make a change of rights happen?  In the REA model, I don't see how to make it happen with just an agreement.  They mean agents agree, but maybe it will happen and maybe it won't.

elf-pavlik:
> What's the difference between those? What is a more "pure" transfer than currency transfers? It is only about digital records.

What I consider "pure ownership transfer" doesn't increment or decrement resource itself. Virtual resources like currencies make it bit confusing but for physical resources it looks much more clear. With physical resource like 400kg of apples which processes can handle as one tangible resource, 10 different people can handle different amounts of it. Still the whole 400kg can get transported as one process. I haven't tried to capture when flow involves quantity of that physical resource, which ownership apply to that involved quantity.

One could have one currency account, let's say in conventional bank, and still have fractional ownership of currency deposited on that bank account. When recording ownership transfers for quantities of currency on that bank account, it would not change amount on that bank account. When one *withdraws* from that bank account, it should also capture which ownership applied to that decrement.

elf-pavlik:
> I haven't tried to capture when flow involves quantity of that physical resource, which ownership apply to that involved quantity.

Actually *provider* and *recipient* reference on flows involving that resource should give enough information whose calculated ownership share gets affected.

fosterlynn:
For "containedIn" relationships, like virtual accounts, or people's owned apples part of a stock resource, I don't see any difference between the container and contained resources regarding this discussion.  A virtual account transfer gets recorded in an app as an event and a change in the account balance; a bank account transfer gets recorded in an app as an event and a change in the account balance.  The recognition by the state seems irrelevant - and i'm not even sure that is different.

elf-pavlik:
For virtual resources like currencies everything happens just as digital records. For physical resources we can actually load them in one place and unload in another place. While that physical process happens any number of agents can share ownership over various quantities of that resource. I would like to avoid vagueness of virtual resource like currencies resulting in lack of clear notion of physical resource which flows can involve. Especially that currencies have no 'use value' and we can consider their role in many ways secondary.

elf-pavlik:
https://github.com/valueflows/valueflows/pull/497/commits/b6fd7c86cda238265cfd6d097cb703f5f069587e has update to `currency-transfer.yaml`

recalling this diagram we can get all 5 variants

transfers

If we had flows where banks handle change of possession, we could specify in addition (possession bit silly for virtual resources, but same will work for transportation)
* origin: *own* flow happens when *possess* flow happen - where alive gives possession to abank
* mid-air: *own* flow happens after *possess* flow where alice gives possession to abank and before *possess* flow where bob gets possession
* destination: *own* flow happens when *possess* flow happen - where bob gets possession


In case of *pre-origin* and *post-destination* we don't even get that 'in transit' resource, transfer directly involves one of the accounts and reflects in calculated shares.

examples/currency-transfer.yaml
```diff
@@ -0,0 +1,177 @@
...
+  # accounts before
+
+  - '@id': abank:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    skos:note: shared account Alice has some credits on
+    classifiedAs: wd:Q179941
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 200
+    share:
```
fosterlynn: what @type is share?

examples/currency-transfer.yaml
```diff
@@ -0,0 +1,177 @@
...
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+    share:
+      - owner: https://alice.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 100
+
+  # actual ownership transfer happens only in this flow
+  - '@id': notary:da8f3360-3a6d-4ea8-b57f-4d4236fe2c3e
```
fosterlynn: Who is the notary in a real life example related to bank transfers?  Or in your example who would the notary be?

examples/currency-transfer.yaml
elf-pavlik: I didn't want to make up name since it doesn't seem to matter how we would call something like that, it pretty much plays role of [Qualified Relation](http://patterns.dataincubator.org/book/qualified-relation.html) between resource and agent. Flows should include all information necessary to calculate it.

examples/currency-transfer.yaml
elf-pavlik: I used 3rd party trusted notary so that we don't need to worry if transfer flow should have IRI in `alice:` or `bob:` namespace. Both of those personal namespaces will still have their digital record describing that *own* flow  - economic event. So notary has a copy as well as alice and bob have their copies.

examples/currency-transfer.yaml
```diff
@@ -0,0 +1,177 @@
...
+      qudt:numericValue: 0
+    share:
+      - owner: https://bob.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 300
+      - owner: https://sam.example/
+        quantity:
+          qudt:unit: unit:Number
+          qudt:numericValue: 300
```
fosterlynn: Should those shares be zero?

elf-pavlik: oops, pasted on wrong resource... fixed!

examples/currency-transfer.yaml
fosterlynn: But is there any such agent in real life in currency transfers?  I don't know of any.  Hard to figure out what is happening if it doesn't match any real life observations.  Is it just to support the urn:uuid?

examples/currency-transfer.yaml
fosterlynn: OK, just wanted to know if it was an EconomicResource or not.

examples/currency-transfer.yaml
elf-pavlik: I don't know details of how mainstream banks do clearings between themselves. I wouldn't expect those banks to adopt VF, they would probably just do the part of moving credits between accounts and VF records would get created outside of that banking system.
Instead of `notary:` namespace in this example I could use `urn:uuid` IRI to denote transfer flow as well. Which doesn't change that both datasets hosted on `alice:` and `bob:` namespaces can have their own digital resource which describes that flow.

fosterlynn:
To summarize a bit where we are at, seems like we are both working towards a one-record transfer event.  That record could be signed and counter-signed.  We are experimenting with different ways to handle this:
* Logical resource identifiers for the independent view, with agents adding their own resource identifiers.
* Transfer referring to the provider's resource, and receiver continuing to refer using that identifier; and/or urn:uuid identifier to bridge the provider and receiver.

Also, we are both trying out the possession - ownership - use rights transfer kind of idea.  In either case, the transfer information could be added to a transformation or transportation related event.
* New actions, allow multiple actions per flow.
* Separate property for these, keep one action.

Does this seem accurate to you?  Any other areas to focus on?

I'm experimenting with having a one-record transfer event be also part of a process, so we can support other inputs and outputs related to the transfer.  Not sure how I like it so far, but will continue with that.

elf-pavlik:
> Share as a separate resource.

I understand you explore it in #498, I have concern here that we would need a way to prevent having rights to rights to rights etc. While transferable rights seem in many ways similar to currencies, I don't think they require anything like 'currency' account, and IMO transferring rights doesn't need to increment or decrement anything.

> We are handling inventory quantities differently. We agree all are derivable. But perhaps disagree on what kinds of inventory concepts we need vocabulary for?

> *  Define inventory possessed only.
> * Define inventory possessed, owned/stewarded, available.

If we explore idea of only using `urn:uuid:` for IRIs denoting resources, in a way we get independent view inventory, possibly a global one.
I see possession tracked using flows, and only one agent can possess given resource at any point in time. I think you take more 'accounting record' focused resource where different quantities of it could stay possessed by different agents, without considering those quantities as resources (do we call it *in limbo* ?). Do you also plan to keep possession and location aligned, so one *inventoried resource* can have *in limbo* quantities possessed by different agents and located in different places? I try to take tangible resources focused approach, where physical resources can't stay possessed or located in different places at different times. If one wants to repossess or relocate quantity of some resource, it takes split & merge and we have another inventoried resource to represent quantity which someone possesses or stays at different location.
As for availability, I think panning it may require scheduling which I explored building from intents and commitments in *scheduling* milestone. This way one can for any time period calculate *available* quantity. 

> I'm experimenting with having a one-record transfer event be also part of a process, so we can support other inputs and outputs related to the transfer. Not sure how I like it so far, but will continue with that.

Which use case you work with?

fosterlynn:
>Which use case you work with?

Check this one. https://github.com/valueflows/valueflows/pull/499/files#diff-4c37f533c3014f38023f8c07e0f2c6f7

fosterlynn:
>> Share as a separate resource.

>I understand you explore it in #498, I have concern here that we would need a way to prevent having rights to rights to rights etc.

I thought these shares were like virtual currency accounts within one legal bank account.   So I wasn't thinking of them as right resources, just "normal" resources.  Not like shares, more like people's currency accounts managed by the group instead of the bank.  Sensorica does this; Enspiral wanted to do it, not sure if they did or not.  Saves on bank fees and administrative contact with the bank.

I think of shares more like on indivisible resources - like maybe the people who contributed to the 3D printer had shares instead of pieces of the printer, not sure.  For currency accounts they have pieces of the containing account, which is actually their own account, just managed internally.

I think of "virtual accounts" as just like serialized bike resources in a bikeshed resource, using vf:containedIn.

fosterlynn:
>If we explore idea of only using urn:uuid: for IRIs denoting resources, in a way we get independent view inventory, possibly a global one.

Can the same urn:uuid: be used for other transfers at a different time of the same resource?

fosterlynn:
>only one agent can possess given resource at any point in time

I agree with this.  (Except of course it could be part of a resource if it is split. But in essence I agree with you, 2 agents can't possess the same material thing at the same time.)

> I think you take more 'accounting record' focused resource where different quantities of it could stay possessed by different agents, without considering those quantities as resources (do we call it in limbo ?

I put all three "inventories" in the resource to try to make all of this clear on its face, so that nothing is in limbo, and nothing is focused over anything else, it is all just laid out.  It does mean that more than one agent can "have in inventory" the same resource.  But they will be different aspects, like the shipper might have possession but not ownership, the manufacturer might have ownership but not possession.

>Do you also plan to keep possession and location aligned, so one inventoried resource can have in limbo quantities possessed by different agents and located in different places?

Location relates to possession, yes, not to ownership or right-to-use.  Location and possession are materially based; ownership and right-to-use are not, they are rights based.

elf-pavlik:
> Can the same urn:uuid: be used for other transfers at a different time of the same resource?

Yes `urn:uuid:89964f76-e4eb-467d-a9fe-e5775fa755a2` works as valid IRI just as `https://alice.example/89964f76-e4eb-467d-a9fe-e5775fa755a2`, they only differ with `https:` scheme having defined way to dereference.

> I agree with this. (Except of course it could be part of a resource if it is split. But in essence I agree with you, 2 agents can't possess the same material thing at the same time.)

I would avoid having 'just split' and always having split & merge so that we have regular inventoried resource possessed by some agent and no *in limbo* resources.  

> Location relates to possession, yes, not to ownership or right-to-use. Location and possession are materially based;

I understand that you consider having one resource which quantities at given point in time can stay under possession of different agents and located in different places? I find it more straight forward to just have each of those quantities as normal inventoried resource, even if they stay 'in transit'. Split & merge seem to do it just fine and allows treating everything as inventoried resource, rather than keeping track of quantities of a resource in different places without treating those quantities as first class resources.

> I thought these shares were like virtual currency accounts within one legal bank account. So I wasn't thinking of them as right resources, just "normal" resources. Not like shares, more like people's currency accounts managed by the group instead of the bank.

I think we need to stay careful with basing to much on virtual resources like currencies. I would prefer to prioritize resources which have aspect of 'use value' and make sure that we model them well. I'll make another example with apples instead of credits. I consider those 'shares' just as computed attributes, didn't even give them `@type` or `@id`, I don't consider them as resources since flows wouldn't involve them. Flows would involve the actual resource and *provider* & *receiver* information on those flows would get use to determine which *computed shares* should get affected.

fosterlynn:
>I would avoid having 'just split' and always having split & merge so that we have regular inventoried resource possessed by some agent and no in limbo resources.

I don't understand.  Where do I have resources in limbo in my examples?  (And I didn't mean 'split' as an action, sorry if that was confusing.  I just meant that some event decremented some aspect of inventory in a resource, and that aspect of inventory appeared as a new resource.)

>I understand that you consider having one resource which quantities at given point in time can stay under possession of different agents and located in different places?

Not sure where this came from?  I tried to say clearly that a resource can only be possessed by one agent at a time.   Maybe your impression relates to something I said about ownership/accounting inventory?  That is not the same as possession, in the scheme I am trying to work with.  I am getting pretty fond of that separation actually, it removes some fuzziness in "transfer" and "inventory".

>Split & merge seem to do it just fine and allows treating everything as inventoried resource, rather than keeping track of quantities of a resource in different places without treating those quantities as first class resources.

But you ignore the requirement for ownership/accounting inventory.  I think it is a real requirement, even in the "next economy", because there is a lot of responsibility involved in ownership that isn't fair to put on to anyone who has possession of a resource for purposes of providing a service.

> I'll make another example with apples instead of credits. I consider those 'shares' just as computed attributes, didn't even give them @type or @id, I don't consider them as resources since flows wouldn't involve them.

If they are separable, why wouldn't flows involve them?  I should be able to come and get my "share" of the apples in the pile, and just take my resource which is then no longer vf:containedIn the larger resource.  If they aren't considered resources in their own right, we end up with yet another concept that seems overly complex and unnecessary.

I'm totally good with apple examples, but I think we need to support both, and it would be good if they can be as similar as possible.  I was focusing on currencies today because they seem harder in some ways, not because I think they are more important. I also don't see our modeling as a matter of personal emphasis, it needs to support any transitional experiments people want to try.  And I don't think it is up to us to decide, we are not that smart, as is no individual at this stage of transition.

elf-pavlik:
>  I just meant that some event decremented some aspect of inventory in a resource, and that aspect of inventory appeared as a new resource.)

I don't find those newly created resources in https://github.com/valueflows/valueflows/pull/499/files#diff-ac40e05f2342541218ee63a90dd9ae41

> Not sure where this came from?

Could you add location information about that 10kg of apples from
> # alice loans a basket of apples to her friend to use in a display for a few days

> But you ignore the requirement for ownership/accounting inventory.

I don't think I do, in my recent examples *own* flow moves ownership over given quantity of specific resource between agents. It can get reflected as those *computed shares* in a way representing qualified ownership relation between agent and resource.

> If they are separable, why wouldn't flows involve them? I should be able to come and get my "share" of the apples in the pile, and just take my resource which is then no longer vf:containedIn the larger resource.

I currently don't find it convincing that we need *containedIn*, one can simply *split* & *merge* quantity one owns which would represent actual physical separation. While resource stays physically as one, process can directly involve the whole resource, for example transport it as one resource between two places. Also ownership transfer of that resource quantities between agent, involves the actual resource directly and doesn't require virtual increments and decrements of two virtual resources. For *pure ownership transfer* we have two agents and one flow involving quantity of one resource. Merge and split come into play when we actually want to divide physical resource, one could also do directly *produce* to and *consume* from it.

> I also don't see our modeling as a matter of personal emphasis, it needs to support any transitional experiments people want to try. And I don't think it is up to us to decide, we are not that smart, as is no individual at this stage of transition.

I don't see it as personal emphasis. IMO if we focus to much on virutal resources we may not handle very well physical resources which provide actual *use value*. On the other side I see it likely that if we handle physical resources well, we should have straight forward way to apply it to virtual resources as well.

fosterlynn:

Issue #365, opened 2016-05-28
shares and 'logical stocks' of material resources


elf-pavlik:
I would like to discuss it based on [Shares: Peanuts deposited in community storage](https://github.com/valueflows/valueflows/blob/master/use-cases/shares.md) use case

> 1. Community of 20 people produced 1000kg of peanuts in one season
> 2. They allocated 200kg to their community kitchen
> 3. They distributed shares to the remaining 800kg using [resource distribution algorithm](https://github.com/valueflows/valueflows/issues/120#issuecomment-217671421)
> 4. All the peanuts stay stored in the same cellar
> 5. Inventory management system keeps track on the allocations
>    - 1 allocation of 200kg community kitchen
>    - 20 allocations adding up together to 800kg
> 6. Each person has a claim for her/his share of peanuts in one's own independent data space
> 7. People can transfer those shares to each
> 8. When someone requests donation of peanuts, shareholders can meet that requests by combining the total weight from any combination of fractions of their shares
### Agents

Let's just focus on 3 people, where one of them also manages the community kitchen:

``` yaml
'@context': https://w3id.org/valueflows/v1
'@graph':

  - '@id': wendy:identity
    '@type': foaf:Person

  - '@id': matthew:identity
    '@type': foaf:Person

  - '@id': joel:identity
    '@type': foaf:Person
    'skos:note': Joel manages community kitchen
```
#### Physical

``` yaml
'@context': https://w3id.org/valueflows/v1
'@graph':

  - '@id': commune:peanuts-2016
    '@type':
      - vf:MaterialResource
      - prodont:Peanut
    'vf:quantity':
      '@type': qudt:QuantityValue
      'qudt:unit': unit:Kilogram
      'qudt:numericValue': 1000
    'vf:currentLocation': commune:big-cellar
```
#### Virtual

``` yaml
'@context':
 - https://w3id.org/valueflows/v1
 - https://w3id.org/role/
'@graph':

  - '@id': commune:kitchens-peanuts-2016
    '@type':
      - vf:MaterialResource
      - prodont:Peanut
    'vf:partOf': commune:peanuts-2016
    'role:owner': commune:identity
    'role:custodian': joel:identity
    'skos:note': For use in community kitchen
    'vf:quantity':
      '@type': qudt:QuantityValue
      'qudt:unit': unit:Kilogram
      'qudt:numericValue': 200

  - '@id': commune:joels-peanuts-2016
    '@type':
      - vf:MaterialResource
[...]

PR #498, opened 2019-04-20
transfer using R&R as resources: experiment

fosterlynn:
First example, book checkout.  More to come as long as it seems like a viable possibility.

### DIFF
--- a/examples/checkout-book.yaml
+++ b/examples/checkout-book.yaml
@@ -13,7 +13,7 @@
     skos:note: Valueflows cookbook
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: 1
+      qudt:numericValue: 2
 
   - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
     '@type': EconomicResource
@@ -21,7 +21,7 @@
     underlyingResource: library:2402b57c-1c45-423a-acb2-f50961c97d65 # maybe needs different name
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: 1
+      qudt:numericValue: 2
 
   # check out
   - '@id': library:f47064eb-7120-4b31-b882-770165901fe6
@@ -32,7 +32,7 @@
     '@type': EconomicEvent
     inputOf: library:f47064eb-7120-4b31-b882-770165901fe6
     action: give
-    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #library's possession resource
     time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
     provider: https://library.example/
     receipient: https://bob.example/
@@ -44,7 +44,7 @@
     '@type': EconomicEvent
     outputOf: library:f47064eb-7120-4b31-b882-770165901fe6
     action: receive
-    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #??
+    resourceInventoriedAs: bob:99bc3e4e-79a6-4b52-a178-8075a268f0df #bob's possession resource
     time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
     provider: https://library.example/
     receipient: https://bob.example/
@@ -60,7 +60,7 @@
     '@type': Commitment
     inputOf: library:953b40f2-4e47-4486-afb2-402512f5269c
     action: give
-    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #the possession resource
+    resourceInventoriedAs: bob:99bc3e4e-79a6-4b52-a178-8075a268f0df #bob's possession resource
     quantifiedAs:
       qudt:unit: unit:Number
       qudt:numericValue: 1
@@ -70,7 +70,7 @@
     '@type': Commitment
     outputOf: library:953b40f2-4e47-4486-afb2-402512f5269c
     action: receive
-    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #the possession resource
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #library's possession resource
     quantifiedAs:
       qudt:unit: unit:Number
       qudt:numericValue: 1
@@ -82,7 +82,7 @@
     skos:note: Valueflows cookbook
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: 1
+      qudt:numericValue: 2
 
   - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
     '@type': EconomicResource
@@ -90,4 +90,4 @@
     underlyingResource: library:2402b57c-1c45-423a-acb2-f50961c97d65 # maybe needs different name
     currentQuantity:
       qudt:unit: unit:Number
-      qudt:numericValue: 0
+      qudt:numericValue: 1
--- a/examples/checkout-book.yaml
+++ b/examples/checkout-book.yaml
@@ -1,0 +1,93 @@
+'@context':
+  - https://git.io/vf-examples-jsonld-context
+  - library: https://library.example/
+    bob: https://bob.example/
+
+
+'@id': rgh:valueflows/valueflows/master/examples/checkout-book.yaml
+'@graph':
+
+  # resource before
+  - '@id': library:2402b57c-1c45-423a-acb2-f50961c97d65
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Possession rights and responsibilities for the Valueflows cookbook
+    underlyingResource: library:2402b57c-1c45-423a-acb2-f50961c97d65 # maybe needs different name
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  # check out
+  - '@id': library:f47064eb-7120-4b31-b882-770165901fe6
+    '@type': Process
+    skos:note: renting Valueflows cookbook to Bob
+
+  - '@id': library:54b814ee-62dc-40c1-bb96-f8582aa4f771
+    '@type': EconomicEvent
+    inputOf: library:f47064eb-7120-4b31-b882-770165901fe6
+    action: give
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
+    provider: https://library.example/
+    receipient: https://bob.example/
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': bob:40bb78fe-28df-41a6-b698-525fa9d411d6
+    '@type': EconomicEvent
+    outputOf: library:f47064eb-7120-4b31-b882-770165901fe6
+    action: receive
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #??
+    time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
+    provider: https://library.example/
+    receipient: https://bob.example/
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': library:953b40f2-4e47-4486-afb2-402512f5269c
+    '@type': Process
+    skos:note: returning Valueflows cookbook to library
+
+  - '@id': bob:98c2bfeb-9c70-4801-896c-4646b975a7d9
+    '@type': Commitment
+    inputOf: library:953b40f2-4e47-4486-afb2-402512f5269c
+    action: give
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #the possession resource
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    before:  2019-05-17T18:00:00-5:00
+
+  - '@id': library:18520fad-4443-4c55-8f76-67738b839950
+    '@type': Commitment
+    outputOf: library:953b40f2-4e47-4486-afb2-402512f5269c
+    action: receive
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #the possession resource
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    before:  2019-05-17T18:00:00-5:00
+
+  # resource after
+  - '@id': library:2402b57c-1c45-423a-acb2-f50961c97d65
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Possession rights and responsibilities for the Valueflows cookbook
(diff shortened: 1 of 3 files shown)


fosterlynn:
Added the currency transfer example

>I have impression that we still try to make ownership a special case which requires moving resource between inventories. While for example possession doesn't move resource between inventories, it still stays inventoried by the owner.
>I don't say that we should or shouldn't do it this way, I just see that we can't have it both ways, suggest that we want to model all different R&R the same way but still have special treatment for ownership.

On this one I just took out the R&R information.  As of now, I think that the way to differentiate is that give/receive that are of the underlying resource (in this case the LETS account, but for the library the books if they for example gave them away when they were worn out), there is not R&R defined, it is assumed to be "full R&R" or "primary R&R", or "ownership" kind of thing.  R&R resources would be partial R&R of some sort.

examples/checkout-book.yaml
```diff
@@ -0,0 +1,93 @@
...
+  - '@id': library:2402b57c-1c45-423a-acb2-f50961c97d65
+    '@type': EconomicResource
+    skos:note: Valueflows cookbook
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 2
+
+  - '@id': library:21f361a6-2375-46bb-b192-c21b5ba833bf
+    '@type': EconomicResource
+    skos:note: Possession rights and responsibilities for the Valueflows cookbook
```
elf-pavlik: How one can know what rights it represent? Would you have different resource to represent each different rights?

examples/checkout-book.yaml
```diff
@@ -0,0 +1,93 @@
...
+    underlyingResource: library:2402b57c-1c45-423a-acb2-f50961c97d65 # maybe needs different name
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
```
elf-pavlik: if after return, wouldn't it have 2 again

examples/transfer-currency.yaml
```diff
@@ -0,0 +1,72 @@
...
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+  - '@id': lets:33f11508-92b3-4095-9e03-6678cddf7adf
+    '@type': EconomicEvent
+    action: receive
+    outputOf: lets:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    resourceInventoriedAs: lets:3be5259d-10f0-431c-9fec-9c0c15a461d3
+    provider: https://alice.example
+    receiver: https://bob.example
```
elf-pavlik: What difference you would expect in application interpreting the data if only *give* had provider and receiver? So if application get your snippet here and another equivalent one but no provider or receiver on *receive*.
How do you prevent application from counting it as two different transfers on two different events, 100 LETS each?

examples/checkout-book.yaml
fosterlynn: >Would you have different resource to represent each different rights?

I think so.  At least everything that this agent would need for that underlying resource.  Could be combinations of rights.  Haven't thought more about how to define R&R, but leaning towards something like POA but improved, along with reference to agreements.

examples/checkout-book.yaml
fosterlynn: Right.  I think is just after checkout?

elf-pavlik:
Which actions one can perform on 'rights resources' ? It looks like only *give* and *receive*, so they behave very similar to currencies. In some ways we could consider rights as currency, one with clearly defined underlying resources.

examples/checkout-book.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+  - '@id': library:953b40f2-4e47-4486-afb2-402512f5269c
+    '@type': Process
+    skos:note: returning Valueflows cookbook to library
+
+  - '@id': bob:98c2bfeb-9c70-4801-896c-4646b975a7d9
+    '@type': Commitment
+    inputOf: library:953b40f2-4e47-4486-afb2-402512f5269c
+    action: give
+    resourceInventoriedAs: bob:99bc3e4e-79a6-4b52-a178-8075a268f0df #bob's possession resource
```
elf-pavlik: how does this 'possession resource' gets its underlying resource reference?

examples/checkout-book.yaml
```diff
@@ -0,0 +1,93 @@
...
+  # check out
+  - '@id': library:f47064eb-7120-4b31-b882-770165901fe6
+    '@type': Process
+    skos:note: renting Valueflows cookbook to Bob
+
+  - '@id': library:54b814ee-62dc-40c1-bb96-f8582aa4f771
+    '@type': EconomicEvent
+    inputOf: library:f47064eb-7120-4b31-b882-770165901fe6
+    action: give
+    resourceInventoriedAs: library:21f361a6-2375-46bb-b192-c21b5ba833bf #library's possession resource
```
elf-pavlik: since we have *give* on possession 'resource' we don't have any action on 'underlying resource' so if underlying resource has current physical location, and current quantity 2, it looks like both books still stay at the library.

How do we handle situation when library transfers rights to possess book on monday and bob doesn't excercise those rights until wednesday. So for two days Bob already has rights but underlying book still stays at the library, on wednesday we have another event where Bob picks up 1 copy of the book but we don't have transfer o rights any more since he had rights to do it already on monday.

examples/transfer-currency.yaml
fosterlynn: Why would only give have provider and receiver?

>How do you prevent application from counting it as two different transfers on two different events, 100 LETS each?

Because they are input and output of a process, each only affecting one resource.  The whole thing is a "transfer process", not just one event.  The events are giving R&R and receiving R&R.  But that duplication from 2 agents' perspectives is why I was trying to make transfer one event with 2 resources.  You wanted this I-P-O structure, and I'm trying to work with it. You got the transfer to one thing, but I though it was messy and maybe not possible in some situations to have to deduce the other side of the transfer without direct reference to it, with the transfer information attached to one event.

examples/checkout-book.yaml
fosterlynn: It is defined by the agent at resource creation.

examples/checkout-book.yaml
fosterlynn: > if underlying resource has current physical location

Yeah I ran into that in the shipping example I'm working on.  I didn't figure the library needed a location defined.  But still it is a valid question.  In the shipping example, I'm going to go back to an issue/receive to a new resource for the stuff that will go on the truck, which I initially removed.

>How do we handle situation when library transfers rights to possess book on monday and bob doesn't excercise those rights until wednesday

I'll give that a try.  I think that is an 'availability' right, since nobody else can check out the book in the meantime, it is reserved.

fosterlynn:
>How would you create two different plans where,
>*    one includes intent to borrow 'Valueflows cookbook' for 2 weeks - just use it in that time
>*    one includes intent to get a copy of 'Valueflows cookbook' - own it

>Later each of those plans gets a commitment which satisfies intent, how would those commitments look?

I can add that example.  In general, from experience people need to have logic to not double commit things, no matter how we handle transfers.  So the logic will need to know the implications of different different kinds of rights (wherever they are in the model), and look at all the commitments, etc.

fosterlynn:
>Which actions one can perform on 'rights resources' ? It looks like only give and receive, so they behave very similar to currencies. In some ways we could consider rights as currency, one with clearly defined underlying resources.

That is an interesting comparison.

And a good question, since someone given possession rights or rights to use an underlying resource, or however we define it, will sometimes want to record the actual `use` event for example.  Will put that into the thought process, right now it seems a bit messy.  Maybe defining rights as actions would help; but we can't ignore responsibilities.  Sometimes it is pretty much all responsibility, like shipping or legal custodianship at sensorica.  So the action method has to define how to deal with responsibilities before we can consider it imo.

examples/checkout-book.yaml
elf-pavlik: If we talk about rights, having rights to posses doesn't mean that one possesses something already, it just means that one can possess something while having those rights. Does your 'rights' imply actual possession or just rights to posses (which one may exercise at some point or not)

fosterlynn: Very true.  I pulled possession out of the rights at one point, probably need to do it again. 

And it seems like I need to connect the underlying resource and the rights more tightly in some way, something more like a bundle than an underlying resource.  You do need to have possession of some sort in order to for example use something.

fosterlynn:
>>Which actions one can perform on 'rights resources' ? It looks like only give and receive

>And a good question, since someone given possession rights or rights to use an underlying resource, or however we define it, will sometimes want to record the actual use event for example.

Forgetting the bad use of the term possession, maybe this means that the rights to use something are `consume`d when the machine itself is actually used.  Especially since the usage could be measured in hours (from the rights to use also having unit of hours).  Meaning the `use` the equipment event becomes a `consume` the usage rights event as input to a process.  That is, after the transfer of the rights; and the rights are transferred back later, after the usage.

Possibly getting ahead, since I'm not sure how this would work when there is no transfer, equipment is just "owned" and used by one agent.  There still could be a usage resource possibly, if they use a calendar anyhow.  Or not.  Anyhow, these 2 milestones do have some overlap.

fosterlynn:
Based on comments in the last day, I'm working on an idea that is a small extension of this PR, will have to find out the devils in the details of course.  I think it will help with our quantity quandaries too.  Maybe not totally solve, but help.

Basically, the underlying resource "comes along" with the R&R resource in (most) events. Or at least some quantity of the underlying resource.  But the action may handle the quantities differently on the flows.  Not all details ironed out below, but I think it has possibilities.  In this picture, I'm thinking of issue/receive as not changing any R&R, basically same provider and receive agent.  Give/take changes possession and transfers some rights.  If the give/take references the R&R resource, it is partial R&R.  If it references the underlying resource, it is full R&R, then gone from the provider's control.  This is work in progress, and will need some yaml to catch everything too.

R R

fosterlynn:
Made some additions to the flow table to consider various effects of actions.  Putting here too for readable version.

Action | Affect | I/O | Changes existence | Changes possession | Changes availability | Pairs with |
------ | ------ | --- | ----------------- | ------------------ | ---------------------| ---------- |
produce | Increment  | Output | Yes | N/A | Yes | N/A |
consume | Decrement  | Input | Yes | N/A | Yes | N/A |
use | No effect  | Input | No | No | Yes | N/A |
work | No effect  | Input | N/A | N/A | Yes | N/A |
cite | No effect  | Input | No | No | No | N/A |
load | Decrement  | Input | No | N/A | Yes | unload |
unload | Increment  | Output | No | N/A | Yes | load |
accept | No effect  | Input | No | N/A | Yes | improve |
improve | No effect  | Output | No | N/A | Yes | accept |
give | Decrement | Input | No | Yes | N/A | take |
take | Increment | Output | No | Yes | N/A | give |
issue | Decrement | Input | No | No | No | receive |
receive | Increment | Output | No | No | No | issue |
raise | Increment | N/A | No | No | No | N/A |
lower | Decrement | N/A | No | No | No | N/A |

elf-pavlik:
> Forgetting the bad use of the term possession, maybe this means that the rights to use something are `consume`d when the machine itself is actually used.  Especially since the usage could be measured in hours (from the rights to use also having unit of hours).  Meaning the `use` the equipment event becomes a `consume` the usage rights event as input to a process.  That is, after the transfer of the rights; and the rights are transferred back later, after the usage.

This sounds to me more like making commitment and fulfilling it. I think we need to clarify when we make commitments and when we transfer rights. I would say that one transfers rights to the person who receives them can now make commitments oneself. This again brings up #464 where one can get broad 'quota' like commitment and later refine it with more specific commitments. Rights could come helpful here if provider doesn't want to need to create those more refined commitments oneself.

> I'm thinking of issue/receive as not changing any R&R, basically same provider and receive agent. Give/take changes possession and transfers some rights. If the give/take references the R&R resource, it is partial R&R. If it references the underlying resource, it is full R&R, then gone from the provider's control. So both issue/receive and give/take are now process-related events and so can do a bit more than transfer of rights - change of possession or location without a whole transportation process (or we can consider that a transportation process if we want, but not load/unload, vehicle, etc.). This is work in progress, and will need some yaml to catch everything too.

I'll take a look at that. How would 'FOB pre-origin' variant work? So Alice just gives full rights to 30kg of apples (pure transfer) but underlying resource doesn't get affected in any way. As I mentioned in other issue I still don't clearly see 'FOB post-destination' variant. It would look something like 30kg of Alice's apples get transported and received into shared resource (Alice still has full rights to 30kg of apples in that shared underlying resource). Next day Alice gives those full rights to 30kg apples in that underlying resource to Bob, 'pure transfer' of R&R which doesn't affect the underlying resource in any way. I can also think of 'FOB mid-air' variant, it would work better with 3 tons of apples but let's stay with 30kg. Alice loads 30kg of apples on the truck going between bio-regions, instead relying on 'middleman' distributor in destination region, she uses online network to start exchanging full rights to fractions of those 30kg of apples already on board. By the truck arrives to destination (warehouse managing shared resources like in #365) Alice doesn't have any more rights to that 30kg, now Bob has rights to 10kg, Claudia to 15kg and Daniel to 5kg (again it would make more sense with metric tons), still no one opened trailer and the underlying resource arrives unchanged.

I think loosely coupled model of rights and underlying resources should handle all the three extensions to original 'FOB origin' and 'FOB destination'.

fosterlynn:
> I think we need to clarify when we make commitments and when we transfer rights.

Makes sense.  Will see how it all works for transformation related events, not sure there even should be any rights directly involved in recording transformation events, but more work to be done there....

elf-pavlik:
I think modeling rights using accounts - resources, might have a challenge related to rights to those accounts, which means rights to rights. I don't think we should allow it, otherwise we may have rights to rights to rights and so on.

fosterlynn:
>I think modeling rights using accounts - resources, might have a challenge related to rights to those accounts, which means rights to rights. I don't think we should allow it, otherwise we may have rights to rights to rights and so on.

Yeah, its hard to see that there would ever be partial rights to an account.  

If we want to "not allow" it, I think we will need to have some category / bahavior / classification defined in the knowledge layer for people to be able to implement that.  Which we have found we needed anyhow for ResourceSpecification. We had "account" and "work" and "other", and then some others got added for limiting dropdowns when we were experimenting with logging in a sentence structure.

fosterlynn:
>> I think modeling rights using accounts - resources, might have a challenge related to rights to those accounts, which means rights to rights. I don't think we should allow it, otherwise we may have rights to rights to rights and so on.

>Yeah, its hard to see that there would ever be partial rights to an account.

Re-reading, I may have been confused on what you meant.  I don't see that an account is an R&R resource itself, if that is what you meant.

fosterlynn:
>>Give/take changes possession and transfers some rights.

>How would 'FOB pre-origin' variant work? So Alice just gives full rights to 30kg of apples (pure transfer) but underlying resource doesn't get affected in any way

Yes, I'm having trouble with my statement.  Because not all transfers of rights do change possession, to your point.  Same old problem.  But change of possession almost always implies some R&R have been transferred.  Some way, we need to fit that in so there are not accusations of teleportation.

>> Forgetting the bad use of the term possession, maybe this means that the rights to use something are consumed when the machine itself is actually used.

>This sounds to me more like making commitment and fulfilling it. I think we need to clarify when we make commitments and when we transfer rights.

Yeah, getting ahead of myself, will stick to transfers for now, and see how the other stuff nets out.

elf-pavlik:
>  I don't see that an account is an R&R resource itself, if that is what you meant.

I didn't refer to 'currency accounts' but the R&R resources which one can increment and decrement pretty much act as 'accounts' for rights, they don't really represent any actual resource, just provide bucket to keep rights in it.

As for your diagram:

1. I would not use *produce* and *consume* on rights, those two 'change existence' and this doesn't seem meaningful for rights.
2. It looks to me tricky to keeps rights and actual resources they refer to in sync. If we *produce* quantity of a resource, do we actually *raise* quantity of full rights? Similar what if we *consume* quantity of a resource, do we *lower* quantity of full rights? It seems that we duplicate current quantity of a resource between resource record and rights record. On the other hand if we want to have possibility of transferring rights independently of flows affecting actual resources, possibly we can't avoid that duplication.

I'll try to focus on how rights should interplay with commitments. I still think that rights should pretty much only delegate capacity to create specific commitments, then one exercises those rights by creating commitments and fulfilling them with events (sometimes going directly to events).
When it comes for shipping, I don't think we need to 'transfer rights to load and unload', often those two may require scheduling but even if not it may suffice that agreement includes commitments to *load* and *unload* and doesn't require any transfer of rights.

PR #499, opened 2019-04-26
transfer - experiment with POA

fosterlynn:
Trying this out: POA flagged on give and receive events.  Will need to think of different names for some of this.  I am using "availability" differently than others and I don't mean the same thing as we mean when we say some resource is available next week because we can see the commitments.  And "ownership".... needs some discussion.....

Here is the transportation example:
poa


fosterlynn:
>Right now my definition of A is "rights to possess and/or use a resource" 

Thinking now that should be just "rights to use a resource", handling possession separately.  This might get the combinations of POA into something more concise.  I think all of these will move smoothly from commitment to event.  And I hope will remove overlaps with the commitment concept itself.
* P (for when a receiver agent will perform a service for the provider agent and needs P for that)
* O (for when ownership is transferred with possession happening at some other point in time, either because of transportation or because there is some other PA in effect like an apartment lease)
* PA (for possession plus rights to use)
* PO (retail, other transfers of O without transportation)

Thinking that use won't happen without possession - so the P and A will be committed and actually transferred together.

Thinking that we don't need to explicitly include A (rights to use) in PO because that can be determined by if there are other commitments that include A that will conflict or not.  I suppose it includes immediate A.  Maybe that should be POA.

Anyhow, if this proves out, I might try to make these into actions, if so they will have new names. If not, at least O and A need new names anyhow.

fosterlynn:
Could be something like:

P: issue/receive
O: confer/acquire
PA: grant/gain
PO(A): give/take

I think anything that's an O in it would decrement/increment, anything else would not, and people could figure out the status from the resource plus commitments.  So this restores the ability of actions to have an assigned resource effect, which could potentially be configurable for an app/group.

But.... this doesn't make things super clear when things happening piecemeal, like O then P(A).

If we leave them as attributes on give/take, then they could be something like:
O: stewardship (or we need both ownership and stewardship? like people steward their land, but own their toothbrush?)
P: possession
A: reservation

In that case, any give/take with an O or an S would decrement/increment, otherwise not.

Could be done like @elf-pavlik suggestion earlier, I think I like this better than individual flags:
```yaml
transfer:
  - possession
  - stewardship
```

fosterlynn:
One question in my mind is would we ever want to piggy-back transfer of POA/etc. onto a transformation or transportation related event?  It would have to be an event that occurs in an instant.  If so, what would be an example?

examples/transport-with-transfer.yaml
```diff
@@ -0,0 +1,169 @@
...
+  - '@id': alice:ad56a7ed-be3c-4937-a3fb-0f156bcd2c47
+    '@type': EconomicEvent
+    inputOf: alice:33e8933b-ff73-4a01-964a-ca7a98893083
+    action: give
+    resourceInventoriedAs: alice:21f361a6-2375-46bb-b192-c21b5ba833bf
+    provider: https://alice.example/
+    receiver: https://bob.example/
+    quantifiedAs:
+      qudt:unit: unit:Kilogram
+      qudt:numericValue: 30
```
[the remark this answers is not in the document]
fosterlynn: So, alice gives possession to claudia, claudia gives possession to bob.  I was thinking possession doesn't decrement/increment.  Then at the end, alice give ownership to bob, that decements and increments.

But that is somewhat confusing, I agree.  Alice needs to know she has the apples in her inventory (in the sense of ownership, part of her "assets", but also something she is responsible for still, in case of problems with shipping for example); but that they are not available because she has given possession of them.  So that is the "inventory in transit".  Available quantity could have its own attribute, currentAvailableQuantity or something.  Or it could be calculated for whatever time period someone needs to know, based on various commitments, transfer of possession, etc.

examples/checkout-book.yaml
```diff
@@ -0,0 +1,80 @@
...
+    resourceInventoriedAs: bob:99bc3e4e-79a6-4b52-a178-8075a268f0df
+    time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
+    provider: https://library.example/
+    receipient: https://bob.example/
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+    transferPossession: true
+    transferAvailability: true
+
```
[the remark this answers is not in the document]
fosterlynn: That is a good question.  Library needs to know both that they still own the book, and the the book is checked out.  (See other answer too, more or less the same issue.)

fosterlynn:
>How this approach would allow giving ownership on produce quantity of a resource?

Like in the case of produce having a different receiver than provider?  Maybe that is the answer to if we want to use (our version of) POA on other actions, which is fine with me, just embedding the transfer.

>I recall use case of social help organization giving food, but only with rights to transport and consume it, not to give to someone else. This way one can eat that food but can't exchange it for beer.

My opinion: the more complex rights should be outside the scope of VF (in the agreement, which could be a "smart contract" of some type), there are millions of examples.  So I don't feel a need to support that.  But if you want to do more experiments with using actions as part of the rights definition, feel free.  

Thinking a bit about it now (not very deeply), I'm thinking we need as much as is needed to support people's accounting, planning, and tracking/tracing, but not so much that we need to validate contracts or rights and responsibilities.  Not sure exactly where that line is for VF.

examples/checkout-book.yaml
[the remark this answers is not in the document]
fosterlynn: Actually, one difference is whether you expect your resource to come back into your availability or not.

fosterlynn:
Thinking a bit about "inventory" and the discussion where we found out that the traditional definition is mostly "ownership".  I think we want to explicitly broaden that for our purposes, and include 
* inventory for which the agent has primary responsibility (ownership/stewardship)
* inventory the agent is in possession of but does not have primary responsibility
* inventory the agent has rights to use but does not have primary responsibility
* inventory the agent has responsibility to manage as a service to the agent with primary responsibility (something like "custodianship")
* others?

What I'm not sure about is how to define this in the vocab, and how to manage it with decrements and increments.  It might be useful to carry the (derived) current quantities of these, not sure.  

Also, to define how to calculate "availability" of resources at some future time, something like:

inventory of primary responsibility - inventory that will be consumed - inventory that someone has rights to use, or plans to use - inventory that will be given + inventory that will be received - inventory that will be out of the agent's possession for reasons of transport etc.....

Thoughts?

fosterlynn:
On the more specific side, I'm leaning towards using some re-thought version of POA as "transfer of", something like this:

```yaml
transfer:
  - possession
  - stewardship
```

rather than inventing a bunch of new actions, that have to be in pairs to cover the give and take, and aren't named that clearly.  Then we can use the transfer property on any event/action.  And just have to rethink the decrement/increment issue.

examples/checkout-book.yaml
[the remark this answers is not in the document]
elf-pavlik: If they expect to get it  back they should probably have commitment for that. It sounds that they have plan for specific flow to happen.

examples/transport-with-transfer.yaml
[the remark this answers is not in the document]
elf-pavlik: I still think we may go to far from having a clear notion of physical observable resource at some location (or in transit) with certain observable `currentQuantity`. Probably we may need to solve first situation in #365 where we have one physical resource where different agents own different quantities of it. I think such physical resource could stay located in some place or stay on vehicle 'in-transit' and still processes that involve it shouldn't care who owns how much of it.

elf-pavlik:
If we talk about 'rights', what does it mean to transfer 'rights' to possession but not having actual change of possession yet? What commitments on that resource and what events on that resource agent can act having 'possession' rights?

fosterlynn:
>If we talk about 'rights', what does it mean to transfer 'rights' to possession but not having actual change of possession yet? What commitments on that resource and what events on that resource agent can act having 'possession' rights?

Well, I have pretty much thought from awhile back in this round of transfer discussions that "possession" is not a right, it is more a physical thing (whether movement or not) and does not imply rights, although it might imply some responsibilities. But it is intertwined quite a bit with transfers of rights, like transportation is but maybe more so.  I think it  makes some sense to "transfer possession" as a phrase, but maybe we don't want to say that, I don't know.  Still it doesn't transform and mostly doesn't transport, either at all, or maybe an inch or whatever, if someone picks up something to load it, or is handed a key to an apartment, or whatever.  But if it is part of a process, then why not?  I like calling out possession explicitly because it handles your "teleportation" criticism, which was kind of buried in "transfer of rights" before.

I have gradually changed from "availability" (or "reserve" or whatever new better name) meaning just rights to use, not rights to possess, although a transfer of possession will almost always happen too.

This makes the "rights to possess later" just a Commitment to transfer possession and maybe rights to use, not a thing in itself.

What do you think?

examples/transport-with-transfer.yaml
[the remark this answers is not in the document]
fosterlynn: Just to note another complication, an agent could have an inventoried resource that doesn't say its location because it might be in several places, but is identified with the agent and the resource specification/classification, not location.  That would have a currentQuantity.

But yes, the currentQuantity for sure needs to be clarified - there is clear observable quantity, there is quantity that is accounted for on the books, .....

elf-pavlik:
I think possession only changes together with some flow, often in some process. I think we can add mechanism to track change of possession on flows, similar as we need mechanism to track change of location on flows. I don't think it makes much sense to think of 'possession' as a right that agents might want to transfer, especially independently from flows on resources in various processes.

So far to 'rights' seem pretty clear to me:
* full control (aka. ownership) - agent having those rights can do any commitments and any events involving resource, this includes consume from a shared resource (actual physical resource)
* reservation (aka. usage) - agent having those rights can do commitments and events with action *use* involving those resources

Since exercising rights always happens as flows with some actions that involve the actual physical resource we have rights to, I think transfer of those rights mostly acts as optimization so that agent who receives them can make commitments and events oneself and doesn't have to have even simple fully automated CfA with the agent who had the rights originally.

Also once someone has rights, flows that exercise them don't seem to need *provider* & *receiver* since rights already were provided so it doesn't happen any more on each flow. This way we don't have in a way 'counting providing twice'. For example if Alice exchanges rights to use cider press with Bob, she doesn't *provide* any more when Bob uses that press, since now he also has rights to do it. They may still require to make commitments to plan the actual schedule, but since they both have rights they can create those commitments independently (just need to make sure that they don't create colliding ones). If Bob didn't have those rights, he would need to have some minimal (maybe fully automated) CfA with Alice so that she creates the commitment, also then each flow would have Alice as *provider*.

fosterlynn:
>I think possession only changes together with some flow, often in some process.

Agree.  What actions do you think always or sometimes change possession?

fosterlynn:
>>I think possession only changes together with some flow, often in some process.

>Agree. What actions do you think always or sometimes change possession?

Not sure if this was part of what you meant, but also it seems like there may be no "real" thing that changes besides possession.  I think we need flows that only change possession; as well as flows that change possession plus some rights; as well as flows that are doing some other process-related thing that also change possession.

> full control (aka. ownership) - agent having those rights can do any commitments and any events involving resource, this includes consume from a shared resource (actual physical resource)
    
Agree.  With small addition - do any commitments and events that don't conflict with other commitments already made.

>reservation (aka. usage) - agent having those rights can do commitments and events with action use involving those resources

I'd like to keep reservation and usage as separate things, with reservation being rights to use, but not implying any actual usage.  On the observation layer, the reservation event may happen without the usage event.  But I do agree with your statement in general, just not the "aka".

elf-pavlik:
> I think we need flows that only change possession

Do we talk about actions like *issue*, *receive* (eg. book from library) or *load*, *unload* ? I think in many cases change of possession will coincide with transportation (so location of the resource also will change), in some cases we may have just change of possession but I think most use cases will also have some kind of reservation for usage in that case.

fosterlynn:
>Since exercising rights always happens as flows with some actions that involve the actual physical resource we have rights to

Except this? https://github.com/valueflows/valueflows/issues/270#issuecomment-485518155

>I think transfer of those rights mostly acts as optimization so that agent who receives them can make commitments and events oneself and doesn't have to have even simple fully automated CfA with the agent who had the rights originally.

But this I totally agree with.  Transfers just facilitate real processes because the agents involved know their R&R, are coordinated with each other.

>Also once someone has rights, flows that exercise them don't seem to need provider & receiver since rights already were provided so it doesn't happen any more on each flow. This way we don't have in a way 'counting providing twice'.

I don't think we have a problem with having provider and receiver, you will need those to not be looking back on some chain and making implications.  You notice when alice transfers possession to claudia who later transfers possession to bob, those are all different vf:EconomicResources, in 3 different scopes/namespaces, and the actual or implied possession quantities are quite clear.  There is no double counting, it is just moving along its value chain/flow.  Although I still agree we need to define more clearly what is what kind of inventory in relation to decrementing and incrementing.

>For example if Alice exchanges rights to use cider press with Bob, she doesn't provide any more when Bob uses that press, since now he also has rights to do it

Are you thinking of some general blanked rights over some longer time period, not specific rights to use for one period where Alice won't have usage rights during that period?  Otherwise I don't understand.

elf-pavlik:
> Are you thinking of some general blanked rights over some longer time period, not specific rights to use for one period where Alice won't have usage rights during that period? Otherwise I don't understand.

I don't see that part very clearly yet. If we have that apple cider press owned by Alice but Bob also having 'usage' rights to, they very likely want to coordinate schedules so as they make commitments to use it they probably want to make sure they both know about it. Still I don't think alice should appear as *provider* on those commitments since she already provided rights to Bob. If we have something like bungalow, where we have 'exclusive rights' to use it, I don't think agent who provided us with those rights even has to know about our commitments to use it on specific dates, even less appear as *provider* on those commitments and events fulfilling them.

Ownership seems easier, since we always have 'exclusive ownership' over certain quantity of a resource, maybe except some resources which one would manage as serial and have fractional ownership without a way to *issue* your fraction of the resource since one can't divide it.

I think that 'exclusive use' may require clear scheduling and probably better to represent it with a specific commitment than transfer of rights. Probably subletting UC would come handy here, we can make variant of #383 where we sublet it to someone for a week (eg. to a friend as a gift when we don't use it). Non exclusive rights seem like a better candidate to consider it as 'transfer of rights', since they don't directly depend on scheduling, while in many cases all agents who has 'use rights' in the same period of time may want to coordinate their schedules. Actually in some cases it just works, we use laundry machine of our neighbors which they have outside in their garden, we don't need to coordinate schedules and on rare occasions one may try to use it and find it occupied for another hour.
Here actually even though we have rights, we didn't do exchange on those rights, we calculated how much they have to pay for water and electricity for each run and we pay them that with little extra in case they will need maintenance sooner. 

>  I don't think we have a problem with having provider and receiver, you will need those to not be looking back on some chain and making implications.

To understand reciprocity one would still need to see if exchange happens on transfer of rights (eg. bungalow rental) or on *use* flow (eg. laundry machine). I don't want to assume that agents have to agree on all details of reciprocity before events, I can see scenarios where agents check each week or month who provided something for them and plan for the next week or month to provide something back on them based on their intents. Here we need to have it clear when we count transfer of rights, or exercising those rights or sometimes both.

fosterlynn:
>> I think we need flows that only change possession

>Do we talk about actions like issue, receive (eg. book from library) or load, unload ? 

Could be either issue/receive or load/unload.  Although seems like either of those could change possession or not, depending on the scenario.

> in some cases we may have just change of possession but I think most use cases will also have some kind of reservation for usage in that case.

The cases I saw where there is just change of possession without any rights to use is when the receiving agent is providing a service for the providing agent, has no rights to use the resource themselves.  Like repairing my car, transporting my goods, etc.

fosterlynn:
>I don't see that part very clearly yet. If we have that apple cider press owned by Alice but Bob also having 'usage' rights to, they very likely want to coordinate schedules so as they make commitments to use it they probably want to make sure they both know about it. Still I don't think alice should appear as provider on those commitments since she already provided rights to Bob.

I think that a "blanket" general agreement without specific commitments is outside the scope of VF.  I think Agreement in VF should be operational.  I don't see we would have a Commitment to transfer blanket rights in VF, that is some other kind of agreement.

But if we are talking about specific Commitments, I think alice will need to appear on the commitment, and then again on the event.  The commitment promises specific rights which will be exclusive (nobody else can use it then), the event delivers those rights.

>If we have something like bungalow, where we have 'exclusive rights' to use it, I don't think agent who provided us with those rights even has to know about our commitments to use it on specific dates, even less appear as provider on those commitments and events fulfilling them.

This I agree with.  If we already have rights to use and we have possession, we will be the provider on any "sub-let" flows and any use flows.

>Ownership seems easier, since we always have 'exclusive ownership' over certain quantity of a resource, maybe except some resources which one would manage as serial and have fractional ownership without a way to issue your fraction of the resource since one can't divide it.

Agree.  The way they did the fractional thing at Sensorica with the 3D printer was that there was a multi-page complex agreement (not a VF Agreement), including things like people who had put in some money having priority over other members.  But people who had put in money were not considered "owners" in the sense of full rights, for the reasons you point out.  If there was an "owner", it was a group owner made up of those 12 people, who could make those kinds of decisions together.  But there wasn't even that, it was considered a temporary situation and when people were paid back, the printer would go to the Sensorica commons.

>I think that 'exclusive use' may require clear scheduling and probably better to represent it with a specific commitment than transfer of rights.

Agree

>Actually in some cases it just works, we use laundry machine of our neighbors which they have outside in their garden, we don't need to coordinate schedules and on rare occasions one may try to use it and find it occupied for another hour.

Great example, agree.  So there aren't commitments there, just events.  You have an agreement (out of scope of VF unless there some "pricing" terms that are agreed on, which we don't have modeled atm - but in any case, nothing "planned").  But not what I was trying to call an ExchangeAgreement, which involves specific planning, and I think (at the moment) implies exclusivity at any point in time.

I guess if we want to explicitly support those kinds of general terms or blanket agreements, we will need to model them and see how they compare to exchange agreements, which contain actual commitments.

>To understand reciprocity one would still need to see if exchange happens on transfer of rights (eg. bungalow rental) or on use flow (eg. laundry machine). 

Yes, which should be fine in our model because those are just different flows.

>I don't want to assume that agents have to agree on all details of reciprocity before events

Nor do I.  Plenty of room for a less formal gifting reciprocity.

fosterlynn:
added some text on this, incomplete but wanted to get some thoughts out there to see if there is disagreement

examples/transport-with-transfer.yaml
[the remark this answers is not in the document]
elf-pavlik: I think this again shows weak notion of actual physical resource and more focus on accounting records. Physical resource can't stay in multiple places at the same time, most likely each physical resource needed to get transported to one's current location. At the same time 'accounting records' or ownership rights don't seem to care that much about physical location of the resource.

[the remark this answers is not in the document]
fosterlynn: It is possibly more focused on accounting needs, but also just assumes people working in the operational side already know where things are located and don't need a computerized record at that level of detail.  It is a valid scenario operationally for there to be one recorded resource per resource specification even if there are separate piles or whatever in different locations, thinking of a space where no transportation is required and you can pretty much see everything easily.  So that applies to one agent scope.

On the topic more broadly, One thing I've discovered from this whole transfer exercise is that different agents may need different records of the same physical resource.  Like the "owner" and the "shipper".  So in different agent scopes, yes I'm seeing multiple records of one resource, some of them more operational, some of them more accounting or responsibility based.

fosterlynn:
I pushed an experiment for different kinds of inventory, see transfers.md and transfer-inventory.yaml.

fosterlynn:
Added some things to the transportation example:
* 2 transportation hops and service output (re. https://github.com/valueflows/valueflows/pull/497#issuecomment-490222903)
* payment for the transportation service in USD

In terms of the first, I don't see any reason to think about composition.  Although I do see we need to work out the "same resource in and out" missing piece in the model, irrespective of composition.  We have done it for translation/editing/proofreading processes.  Transportation is a new use case for it for me, and may need a different mechanism.

In terms of the second, ran into the continually annoying problem of having both a give and take event for a simple transfer, which is still one event in most ways.  Which to put as reference in the agreement of transportation service for USD?  give, take, or both?  I think both is probably required, but the give and take are rather headless since the Process doesn't reference the agreement, so they basically look like a duplicate or 2 payments.  Hmmm, maybe we need a service out of all currency transfers?  Except the service doesn't know the quantity of currency, which is what is being exchanged.  Your suggestion @elf-pavlik ?

(I realize of course that there are also valid reasons to have both events.  All trade-offs.)

fosterlynn:
>In terms of the second, ran into the continually annoying problem of having both a give and take event for a simple transfer, which is still one event in most ways.

Reminded me of another annoyed dev, who is so annoyed that he is re-writing what I did in NRP (2 events per transfer).  And NRP is less annoying because the Transfers are part of the Agreement (Exchange), so they can hide the duplicate events for the exchange UI.  But I was annoyed when I wrote it too. :sweat_smile:  https://github.com/FreedomCoop/valuenetwork/issues/490

Do you think we should reconsider?  I have no brilliant ideas yet, but might spend some time trying for one that keeps the basic I-P-O structure but doesn't duplicate transfer events when one event (or commitment) is expected, like in Agreements.  Or at minimum we need some clearly stated rules for how to do it.

fosterlynn:
Put the terms i've started using into the ttl file and cleaned up the examples.

elf-pavlik:
I have added this commit https://github.com/valueflows/valueflows/pull/497/commits/ec13de1cc788f2c3d11c84068d4724f4d2a72090 to my PR with alternative approach to your *transfer-invontory* example. Some main differences:

* it tries to keep stronger notion of physical resource
 * same resource can't stay in more than one place at the same time
 * possession should change for the whole resource but flow still should copy its current quantity
 * load / unload should happen on the whole resource but flow still should copy its current quantity
* ownership change happens only in accounting records
  * it can happen on any quantity of the resource
  * one needs history of *own* flows to know current ownership of resource quantities
* it avoids having resources 'in limbo'
  * split and merge happen together so we keep all the actual resources inventoried (i omitted that 'stock management' process which just splits and merges)
  * since we *load* the whole inventoried resource, one can split it during transport as well as transfer ownership rights to it while in transport, resource 'in-transit' gets inventoried the same way as stationary resource. 

Since physical processes act on tangible resources, I find it important to keep clear notion of that tangible physical resource and don't drift to much into accounting records. Process can *consume* quantity of a resource owned by different people, I still haven't looked how in such cases record quantities owned by who got consumed in the process. I think process should only need to have one *consume* flow on tangible resource and ownership information would just augment that process with rights related details.

fosterlynn:
I added an example to try out the logical identifier thing for transfer-inventory.  Not sure how much clarity it added really, would need to see if it holds together on other operations.  Especially things like transfer fees; does it still need a process?

fosterlynn:
Pushed some work on currency examples, simple to more complex.  Experimenting with the process that facilitates a transfer outputting the transfer itself as an event, as well as the service.

examples/transfer-cash.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  # accounts after
+
+  - '@id': alice:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
+    '@type': EconomicResource
+    skos:note: Alice's cash 
+    classifiedAs: wd:Q4917 # US DollarS
+    currentQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
```
elf-pavlik: which event supposed to cause decrement on that resource? same for decrement on another resource.

examples/transfer-currency-fee.yaml
```diff
@@ -0,0 +1,76 @@
...
+    outputOf: banka:e0caf1e3-4a9b-4927-a19c-3f681d234529
+    resourceClassifiedAs: wd:Q4917 # US Dollar
+    provider: https://alice.example
+    receiver: https://bob.example
+    quantifiedAs:
+      qudt:unit: unit:Number
+      qudt:numericValue: 100
+
+    # alice and banka record the exact resource (account) on their copies of the transfer
+    resourceInventoriedAs: banka:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead
```
elf-pavlik: would it get incremented or decremented? if any than which event stays responsible for that increment or decrement

examples/transfer-currency-fee.yaml
```diff
@@ -0,0 +1,76 @@
...
+    skos:note: Facilitate moving currency between accounts
+    inScopeOf: https://banka.example/
+
+  - '@id': banka:0468e9b9-4d54-4eee-9d3f-bc76db11710a
+    '@type': EconomicEvent
+    action: transfer
+    transferOf:
+    - possession
+    - ownership
+    outputOf: banka:e0caf1e3-4a9b-4927-a19c-3f681d234529
```
elf-pavlik: why output?

elf-pavlik: I don't think we need to connect *transfer* flows to services. IMO `vf:transfer` flow could happen just on quantity of currency without moving it between bank accounts. I see thank `bank:transfer` fee related to moving quantity of currency between bank accounts. I think it may need equivalent of *load* and *unload* to represent picking up some amount from one account and dropping it off on another account. I see here tricky part that if we just want to split 100 credits from the account and merge it into inventoried resource 'in transfer', we don't have it 'in bank account' but 'in transit', I think we would check how banking systems represent credits already debited from one account and not credited yet on another account, which can happen when `bank:transfer` bounces.

examples/transfer-cash.yaml
fosterlynn: transfer has to both decrement and increment, one in each trading partner view

examples/transfer-currency-fee.yaml
fosterlynn: > IMO vf:transfer flow could happen just on quantity of currency without moving it between bank accounts.

Maybe that's the difference between possession and ownership? But in any case, I think we do need some way to know what transfer the service fee is for. Maybe there is a better way....

The trouble with it being the same as load and unload is that load and unload represent real events that can be seen separately, and accounted separately, and exchanged separately (if that would ever happen). Give and take you can't tell which one to use in an exchange because it is really just one in many ways. But as always, trade-offs.

examples/transfer-currency-fee.yaml
fosterlynn: I think you would have to tell by if it alice is the provider, then it is a decrement to alice's account.  I admit, that is a bit fishy.  And complicated by the relationship between alice and the bank.

Wim has it so there is transfer in the independent view, and give and take in the trading partners' views, and has them modeled as subtypes.  That seems sort of where we could go, but I didn't like the 3 things having different actions in our model.

examples/transfer-cash.yaml
fosterlynn: if we can do transfers as the independent view using logical resource identifiers, then we don't need the complication of doing splits and merges.  So then the transfer can decrement and increment.  But also thinking about possession, ownership which can be separate in terms of increment and decrement.  (I have to start remembering to not just say "increment and decrement" in these contexts.)

examples/transfer-currency-fee.yaml
elf-pavlik: > Give and take you can't tell which one to use in an exchange because it is really just one in many ways.

If alice's bank account gets debited 100 units and next day bob's bank account gets credited 100 units IMO we have two separate flows. When moving units between accounts has 'bounce' alice's account gets debited 100 units first and then next day same account gets credited 100 units. Again IMO we have two separate flows not just two different perspectives on the same flow.

While this movement of units between accounts happen. Transfer of ownership can happen as one flow pre-origin, origin, mid-air, destination, post-destination. In my approach transfer of ownership happens as flow which involves quantity of a resource but neither decrements it or increments it, just affect how shares of that resource get computed at any given moment.

fosterlynn: >If alice's bank account gets debited 100 units and next day bob's bank account gets credited 100 units IMO we have two separate flows

which one do you use in an exchange?

elf-pavlik: None of them, agreement may include any of them as commitment but exchange would include *own* flow with specific amount and specific resource, it would happen as flow on its own anywhere: pre-origin, orign, mid-air, destination, post-destination relative to moving units between accounts. Agents which share same bank account could just have *own* flow as part of exchange and the bank account would never get incremented or decremented. Similar with #365 shares of apples or peanuts stored together but different quantities owned by different people.

elf-pavlik: Actually commitment at certain stage may not specify any specific resource at all, just *own*, provider * recipient, amount, resource classification (eg. MXN), which gets nicely represented with a single flow. Later details could get added about 'virtual delivery / pick-up' of that virtual resource.

fosterlynn:
I've been playing with the examples.  They are not consistent with each other at the moment, and have some embedded questions.  They are related to this comment https://github.com/valueflows/valueflows/issues/270#issuecomment-495746884.

examples/transfer-currency-fee.yaml
fosterlynn: Yes, true, most commitments don't specify any specific resource.

fosterlynn:
Added another example for bank transfer with delayed timing.  I'm sure it is missing a bunch of details, although I did put the central bank in to account for the delay.  But should be enough to look at possession and ownership related to resources.

examples/transfer-currency-timing.yaml
```diff
@@ -0,0 +1,120 @@
...
+  # transfer of currency from alice to bob - first banka sends the transaction to the central bank account
+
+  - '@id': urn:uuid:0468e9b9-4d54-4eee-9d3f-bc76db11710a
+    '@type': EconomicEvent
+    action: transfer
+    transferOf:
+    - possession
+    resourceClassifiedAs: wd:Q4917 # US Dollar
+    resourceInventoriedAs: banka:b75d8f6a-e2df-4e52-b36d-1a22a66f4ead # alice's account
+    toResourceInventoriedAs: centralbank:f47064eb-7120-4b31-b882-770165901fe6 # some central bank account
```
elf-pavlik: How would you do it if banka doesn't expose how it does clearing with bankb (central bank or whatever else), it just shares with Alice information that it decrements her account by 100 units and takes possession of those units.

Similar with bankb, it may only share with Bob information that it incremented his account by 100 units from Alice.

elf-pavlik:
Could you please add variant where that 100 unit bounces back to Alice's account?

fosterlynn:
Added bounce variant.  May not be done how they would do it.

examples/transfer-currency-timing.yaml
fosterlynn: This transfer wouldn't be visible to alice or bob, just between the banks and the central bank.  Alice already wrote a check or initiated a transfer in some way, and then she just waits.

fosterlynn: As I think of the bank stuff, the reality is that alice loaned banka the amount in her account, so really none of these transfers would involve alice or bob as agents, just their accounts.  The only thing with alice would be the money between her and banka, not shown in my examples.  Crypto would be the same in terms of the focus being on the accounts, but wouldn't have that weirdness of the banks actually 'owning' the money with a debt to the depositor.  (That's just a bank rip-off anyhow - but we should be able to support any kinds.)

fosterlynn:
Another thought about the reality.  I'd like to do my latest examples with a scenario where alice wrote bob a check.  So bob deposits his check, and the whole thing is initiated from bankb instead of banka.  Will change that next.

fosterlynn:
>Another thought about the reality. I'd like to do my latest examples with a scenario where alice wrote bob a check. So bob deposits his check, and the whole thing is initiated from bankb instead of banka.

Pushed.  This is probably more realistic, although I did skip a bunch of internal transferring on all sides.  The way I did it doesn't necessarily illustrate the modeling issue at hand.  I started with a Commitment which was the instruction to the central bank for what it needed to do.

Didn't do the bounce scenario, but I think that instead of transferring money, the central bank would just mark the Commitment as finished, and that's that.

Note that bankb could show bob's account as +100 if they want to (but not available) based on the check deposit / Commitment.

fosterlynn:
I'm thinking about ways to address the question of things that are like transfers but don't change ownership or possession.  Like a person can transfer currency amounts between their account and another person's account, but can also transfer between two accounts that are theirs.  Likewise for material transfers.

I think it would be OK to make those use cases also transfers.  But I'm leaning towards a different action called `move`, which would work for material or non-material resources, including currencies.  It would be moving from one location to another, or one account to another.  This would need the option of having a toResourceInventoriedAs just like transfers, at least in my current model here.

I'm going to do some examples and see how it works.  Also some examples where transferring currencies works like transferring material resources, as suggested at some point by @elf-pavlik .

fosterlynn:
I made some changes to examples and added a couple `move` examples (like transfer but without change of rights or possession).

@elf-pavlik if you get a chance to look at all the examples, that would be great.  I'm trying to get to where even if we don't agree on changes to the vocabulary yet, I can add it to the graphql reference, and go forward with a couple projects and see how it works for them.  So if you see anything you think is unworkable or just wrong (as opposed to a matter of opinion on a better way to model), please let me know.  And I'll go ahead and do some work on the ttl and docs so the whole thing is consistent.

PR #501, opened 2019-04-26
Plan (again)

fosterlynn:
I'm re-introducing Plan to the vocab, as another concept that we need sooner than later.  I've thought about the higher-level processes, and think that Plan should be distinct, as an operational entity, even if we end up going with the higher-level process route, which I'm not sure I like atm.

I have more work to do on it, and will also ask @bhaugen to help out.  But wanted to get the thought process going.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -23,6 +23,7 @@
 * [Transfers](introduction/transfers.md)
 * [Exchanges](introduction/exchanges.md)
 * [Proposals](introduction/proposals.md)
+* [Planning](introduction/plans.md)
 * [Scoping](introduction/scoping.md)
 * [Classification](introduction/classification.md)
 * [Recipes](introduction/recipes.md)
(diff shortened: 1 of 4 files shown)


fosterlynn:
Added an example, changed some names.

fosterlynn:
Made a picture to see if that helps clarify the example.  Experimental, not quite what this calls for, but inspired by: https://github.com/valueflows/valueflows/issues/502
 
plan

elf-pavlik:
I think we need to evaluate use cases of planning together with other composition related use cases. I think we should not define two different ways to handle the same use case. I would like to see how we see doing composition and why you think it wouldn't work for this use case.

How urgently do you need this feature? I would prefer trying moving forward with transfers and address those proposals related PRs before jumping into composition.

fosterlynn:
>I think we need to evaluate use cases of planning together with other composition related use cases. I think we should not define two different ways to handle the same use case. I would like to see how we see doing composition and why you think it wouldn't work for this use case.

>I would prefer trying moving forward with transfers and address those proposals related PRs before jumping into composition.

If there is a lot of discussion needed for Plan (which it seems like there is), then I totally agree with your sense of priorities! :+1:  We can go ahead tentatively with Plan (it has been used for awhile in the OCP VF api anyhow so some people are used to it, it's more a matter of leaving it in place).  We can get it cleaned up later as needed.  Things are moving more slowly than expected anyhow.... :)

fosterlynn:
Closing this in favor of https://github.com/valueflows/valueflows/pull/525.

Issue #304, opened 2018-12-10
Intents: "wholesale" vs "retail"


fosterlynn:
I think some of what is making this hard (for me at least :sweat_smile: ) is the broad range of use cases.  Part of this span is how people generally treat "wholesale" vs "retail".  There is probably a gray area there too, but there is a difference between people engaged in production for others and people who are consuming it themselves or in a family situation, or even extended living situation.  At least a continuum.

Agents producing in some quantity for trading tend to have their supply chains already agreed to, with processes, quantities, etc.  This would apply to agents that are aggregating for re-distribution too.  If there is an exception and something else is needed, generally they would talk to their network of suppliers before making some larger broadcast of an intent.

Agents closer to the "consuming" end tend to have more of the one-off broadcast intents, or also regular agreements like CSA's etc.

This issue is to gather some thoughts about these differences, so we can see how they might affect the model.

fosterlynn:
Without even getting to more industrial production (which we do still need of course for many things, think computers or laying cable across the atlantic) - here is a small example from my life.

I do produce catalog updates for a local buying club.  Even in this small local situation there are 2 levels of intents.  I get price lists from the veggie producers.  I aggregate those (the ones that meet the criteria of the buying club - as local as possible and organic), average prices, and load to the buying club software.

I'm pretty sure each of those veggie producers email their price list out each week to all the groups they have agreed to supply.  (Veggies available change fast, that is why it is that frequent.)  So they sell to this buying club, also to regional co-ops, and I don't know where else.  They don't tend to do "retail" (read smaller quantities).  One of them used to do a CSA on the side, but I bet that internally there was a transfer/exchange between the wholesale and the CSA sides of that farm.  So the price lists are the first set of intents.

Then I create another set of intents in the buying club software.  There is an agreement between members of the buying club to pay an additional percentage over wholesale costs to support the buying club expenses that are not covered by volunteer work.  So this set of intents is published inScopeOf this buying club and governed by this agreement.

(Re. less artisanal production, Bob has described the Saturn supply chain, which relied completely on agreements, and did not publish intents except to publish the production schedule, which the tiers of suppliers were expected to work towards.)

elf-pavlik:
That sounds like an interesting use case: a producer providing to multiple distributors.
I think producer would have a single intent, for example estimated for a single week - *issue* 200kg of carrots. Then different distributors would make partial commitments that *satisfy* that intent.

[...]

PR #503, opened 2019-04-27
structure of proposal

fosterlynn:
Starting a new PR for the basic minimum proposal to publish intents.  Will close the old one, too old.  (Old one is here for reference https://github.com/valueflows/valueflows/pull/302.)

This one looks something like this:

proposal

Will do an example.


fosterlynn:
Added a series of simple examples.  Again, I know this is missing a lot of things: some different quantities, a way to reference "unit pricing", and others.  I'm thinking to try to get the basic structure out there, and then keep plugging away at it as we work with users if not before.  And also I understand we will probably want other things than Proposal for more complex use cases, like multi-agent barter agreements, maybe budgeting, etc.

elf-pavlik:
I think we should keep it similar to how we model Recipe, so
* RecipeProces ~ Proposal
* RecipeFlow ~ ProposedIntent

With the difference that commitments in Agreement created from that Proposal need to satisfy already existing intents which Proposal specifies.
Relation from ProposedIntent to Intent I would probably call something like *requiresSatisfactionOf*, meaning that we require created commitment to satisfy specific intent.

I would also like to include right away an example where single intent gets referenced in multiple proposals. This way we minimize chance that someone may try to see Intent as specific for some proposal.

fosterlynn:
>I would also like to include right away an example where single intent gets referenced in multiple proposals. This way we minimize chance that someone may try to see Intent as specific for some proposal.

Good idea. I can do that.

>RecipeProces ~ Proposal

I don't see Proposal that way at all.  A process is about creating, moving, and now giving.  But I don't think the intents under a proposal are related in the same way that flows under a process are related.  You might have give and take, but you won't have give and give in the other direction.  That is an Agreement.  Agreements normally have at least 2 value flows, processes are part of just one.

I think of Proposal as more social, seeking a response, almost part of CfA.  More a way to get Intents out into the world on their way to agreements and commitments.

fosterlynn:
>> I would also like to include right away an example where single intent gets referenced in multiple proposals.

>Good idea. I can do that.

Done.  See if it makes sense.....

elf-pavlik:
> I don't see Proposal that way at all. A process is about creating, moving, and now giving. But I don't think the intents under a proposal are related in the same way that flows under a process are related.

If we forget about transfers of rights for now which I don't think will appear in recipes, recipe describes how to do things in terms of *physical* requirements. What it takes to get something specific done. RecipeProcess and RecipeFlows result in generating Process and Flows (in most cases application can safely generate intents).

> I think of Proposal as more social, seeking a response, almost part of CfA. More a way to get Intents out into the world on their way to agreements and commitments.

Yes, since Agreements define something purely based in social dynamics, not in physical ones. At the same time, looking at application logic we have something similar. Proposal and 'ProposedIntents', plus information from Intents result in reaching Agreement and Flows (eventually we should get commitments which satisfy original intents).
We still have an extra Conversation For Action aspect where we will very likely also need some kind of flow structure only created for the sake of that conversation to reach those commitments.
In case of fully automated system like most current online e-commerce shops. Proposal and 'ProposedIntents' practically work in the same way as Recipe, since we can automatically generate 'order' which usually gets auto approved (just to check for racing conditions between orders) and we get Agreement with Commitments.
If CfA doesn't allow negotiating quantities etc, the whole scaling logic most applications should share between recipes and proposals. We could take an example of Open Food Network 'shoping cart' and we will get 'per item' proposal which will lead to automated creation of 'per item' agreement. We can skip for now the part when those 'per item' agreements get bundled into a single order, which relates to composition and refining plans #464 In that OFN example we should see that creating agreements from proposals and creating processes from recipes shares a lot of logic, even that one works based on physical constraints and other based on social dynamics.

fosterlynn:
>In case of fully automated system like most current online e-commerce shops. Proposal and 'ProposedIntents' practically work in the same way as Recipe,

>We could take an example of Open Food Network 'shoping cart' and we will get 'per item' proposal which will lead to automated creation of 'per item' agreement. We can skip for now the part when those 'per item' agreements get bundled into a single order, 

I'm not sure we even need Proposals for fully automated shopping cart, maybe just Intents, depending on how we want to do "unit pricing".  Maybe we need to figure that out now, I was trying to put it off. :)

Or maybe 'per item' kind of thing, or ongoing e-commerce or whatever, is another choice in the proposal from this one https://github.com/valueflows/valueflows/issues/504?  (Top of head, haven't thought in depth about those use cases.)

>If we forget about transfers of rights for now which I don't think will appear in recipes, recipe describes how to do things in terms of physical requirements.

So, I'm not sure why you think there is a direct mapping?  Proposals are pretty much about transfers (of some POA) and creating exchange agreements.  They won't get too physical, except for possession, although they can certainly trigger planning for more physically based processes.  Or be based on already planned physically based processes, such as the example in the yaml.  Am I misinterpreting something?

elf-pavlik:
> I'm not sure we even need Proposals for fully automated shopping cart, maybe just Intents, depending on how we want to do "unit pricing". Maybe we need to figure that out now, I was trying to put it off. :)

I see Proposal acting as 'template (~recipe) for resulting agreement. It should provide way to define salable quantities on expected commitments. Just as one can further automate creating Process based on RecipeProcess, one should be able further automate creating Agreement from Proposal. In some way one doesn't need proposals to contact a farm and after short conversation agree on some kind of exchange of work for food and accommodation. This wouldn't work with very well for something like OFN online automated orders creation and here we need proposals for all offered products so that consumer alone can create order (bundle of 'per product' agreements).

> Proposals are pretty much about transfers (of some POA) and creating exchange agreements.

I agree that agreements (resulting from proposals) will often have transfer flows. At the same time if someone would want to micromanage everything just with commitments that would get included in transformation and transportation processes. One could run a big scale regional economies without a single transfer. Besides one special action not used with transfers - *work*, agreements can also include very specific *consume* and *produce* commitments which one can do instead handling things with transfers of rights.

>  Am I misinterpreting something?

I pretty much mean that RecipeProcess and Process resulting from it, stays grounded in physical constraints. While Proposal and Agreement resulting from it, stays grounded in social constraints.
In both cases we start with certain pattern of how to create codependent flows, only the reason for that codependency differs: physical vs. social.

elf-pavlik:
As for the proposed `vf:reciprocal`.

How do you plan to implement scenarios where for simplest proposal involving two intents:
* only one intent gets marked as *reciprocal*
* both intents get marked as *reciprocal*
* no intents gets marked as *reciprocal*

I think it supposed to relate to the *purpose* of the intent, which currently I see as 2 general cases:
1. to benefit directly from the 'use value' - mostly when agent wants to receive something
2. to take advantage from 'exchange value' - again seems mostly relevant when agent wants to receive

I think that proposed `vf:reciprocal` suggests the 2nd case. While all currencies only play role in the 2nd case, they have no 'use value', it seems rather straight forward that if resource gets marked as currency it only can get used with suggested `vf:reciprocal` intents.

Even for non currency related intents, one may stay open to accept more of something than one can benefit from 'use value' oneself, to play with the 'exchange value'. But that seems to go already in direction of multilateral exchanges. If I know that farm where I get produce from exchanges it for use of some specific equipment. I could exchange something I provide for rights to use that equipment knowing that I accept the 'exchange value' of it not directly the 'use value'. Later I could exchange those use rights for food. Or we could do the same with 3 way exchange. My openness to accept those 'use rights' fully rely on me knowing intent of that farm to actually *use* that equipment. If we would have the two step exchange instead of multilateral, i think the 'receive use rights' would get marked as proposed `vf:reciprocal` - just for the 'exchange value' not the 'use value'.

Can you think of 3-4 other examples where you would see use for `vf:reciprocal` other than situations involving currencies?

fosterlynn:
>How do you plan to implement scenarios where for simplest proposal involving two intents:

>    only one intent gets marked as reciprocal

matches are made only to the non-reciprocal intent

>    both intents get marked as reciprocal

invalid

>    no intents gets marked as reciprocal

valid: means the agent wants both to be matched

>Can you think of 3-4 other examples where you would see use for vf:reciprocal other than situations involving currencies?

One immediately comes to mind.  I know a woman whose husband makes maple syrup for selling through a local maple syrup cooperative and they keep some for themselves.  They have a couple kids and she advertises frequently for things in our local email classified ad service.  She always says she will trade maple syrup for whatever she wants.  She never offers maple syrup as the primary intent, and she never would want some matching software to find someone who basically is looking for maple syrup to buy or trade for anything besides what she specifically wants at that moment.  She is not trying to get rid of maple syrup, it is just a nice thing she can offer in exchange.

>>I'm not sure we even need Proposals for fully automated shopping cart, maybe just Intents, depending on how we want to do "unit pricing"

>This wouldn't work with very well for something like OFN online automated orders creation and here we need proposals for all offered products so that consumer alone can create order (bundle of 'per product' agreements).

Maybe so.  But what properties that are part of the Proposal would OFN need for their online order creation?  Or what i'm really asking is what would an agent need for ecommerce if they have a steady product line that they just offer all the time.  (Not sure exactly the scope of OFN for what you are thinking about here.)

fosterlynn:
I wanted to reply to some of your broader perspective thoughts @elf-pavlik .

>One could run a big scale regional economies without a single transfer.

I find this an interesting statement, and have been thinking about it.  I don't really see how.  For example, how would the people in a region distribute housing in a fair way, without transfers of rights (from the commons or whatever regional governance) to people for housing?  I only picture the alternative being maybe rifles (out here, handguns in the city).  How would people coordinate responsibilities for resources that are needed for the common good, especially when those need to get passed along to the next person?

>Can you think of 3-4 other examples where you would see use for vf:reciprocal other than situations involving currencies?

Besides the one that came to mind above, I want to say that even if it is only for currencies (which I don't believe), then still it is needed.  And sometimes if people are doing fundraising, the currency will be the primary reason for the intent.  So we need it.

>to take advantage from 'exchange value' - again seems mostly relevant when agent wants to receive

Where we live in the country, people don't much try to take advantage, people know each other, people try to be fair, as well as there is a large informal gift economy.  I don't think it is correct to think of things involving currency as evil somehow.  One of the functions of currency is to facilitate exchange.  In fact for me, the multilateral exchange idea has good possibilities but is often very cumbersome, I would immediately try to set up something around mutual credit rather than that, where you could get approximately the same effect but by just letting it work itself out instead of elaborate planning.  Of course, I support you wanting to do it too.  Just saying that fairness can be coordinated in many ways.

elf-pavlik:
> Where we live in the country, people don't much try to take advantage, people know each other, people try to be fair, as well as there is a large informal gift economy.

I didn't intend to use this definition:
> make unfair demands on (someone) who cannot or will not resist; exploit or make unfair use of for one's own benefit.

But this one:
> make good use of the opportunities offered by (something).

So my sentence didn't try to suggest any evilness just differentiate between when one aims at 'use value' and when at 'exchange value'.

> She never offers maple syrup as the primary intent, and she never would want some matching software to find someone who basically is looking for maple syrup to buy or trade for anything besides what she specifically wants at that moment.

Do you suggest that you see matching systems just take any 'proposed intent' not marked as proposed by you `reciprocal` and ignore all other 'proposed intents` and consider a 'match' any other proposal that just matches that one intent? I have impression that you assume this way of doing 'matching' and proposed *reciprocal* supposed to give it some constraints. Would you still see need for *reciprocal* if matching system would only create 'match' if all (not any) proposed intents can get commitments satisfying them?

Your use case sounds like she would like to have a way to advertise match my intent to *provide* maple syrup against any other of my intents to *receive* except the intent to *receive* USD. The way I see it she could for example create proposal for each of her *receive* intents (except USD) which would also always include the same *provide* maple syrup intent. This in my opinion already gives no ground to matching system to assume that she would want to exchange maple syrup for USD. For that she would probably have special proposal with audience restricted to that coop.

fosterlynn:
>Do you suggest that you see matching systems just take any 'proposed intent' not marked as proposed by you reciprocal and ignore all other 'proposed intents` and consider a 'match' any other proposal that just matches that one proposed intent? I have impression that you assume this way of doing 'matching' and proposed reciprocal supposed to give it some constraints. Would you still see need for reciprocal if matching system would only create 'match' if all (not any) proposed intents can get commitments satisfying them?

That might work for some few situations.  In most one-off situations, I think it should be looser, and bring back more options rather than less options, so yes I would just match on non-reciprocal intents.  Many examples I've seen, the reciprocal intent is not even defined except in text, so with your proposal, many positive matches would be missed.  In the maple syrup example, my friend would never find matches for her requests, because people wouldn't think of requesting maple syrup for say the clothing she needs.  On the other hand, most people would happily accept it for their used kids clothing, because who doesn't like local maple syrup.  I think the way you are thinking about it is too rigid to be workable for people just trying to find things they need and get rid of things they don't so someone can use them and they don't have them cluttering their space.

Also, I still understand it is important for all the existing systems I know to be clear up front in the data if they basically have an offer or a request.  Reciprocal flag is the bare minimum to provide that.  

Again, I support you and others in the use case you want to try, I think it would be a very interesting experiment in an alternative economy.  But that is not everyone.  Having the ability to mark intents as reciprocal does not infringe on people's ability to not mark anything reciprocal for a different use case.

Why don't you think it is valid for some reasonable use cases?

examples/simple-proposals.yaml
```diff
@@ -179,3 +180,83 @@
...
+    eligibleLocation: map:566a2c01-96f8-4b3c-adcf-f197eced3092 # Madison, WI
+
+  - '@id': farm:c7897c39-7f05-4a5d-a487-80e130a24345
+    '@type': Intent
+    name: Carrots case 25#
+    action: transfer
+    provider: https://farm.example/
+    resourceClassifiedAs: wd:Q81 # carrot
+    resourceSpecifiedAs: farm:2402b57c-1c45-423a-acb2-f50961c97d65 # 25 pound case organic carrots seconds
+    exchangeRateQuantity:
```
elf-pavlik: Intent itself shouldn't have any notion of 'exchange rate' since same intent can get referenced from any number of proposals. I guess you could try moving it into Proposed Intent which stays specific to single Proposal.

elf-pavlik:
I think we may need to have an explicit construct to create 'one side open' Proposal. If we take example of some e-commerce store, just because they proposing to exchange product for currency X, one shouldn't assume that they want to exchange that product for other things.
At the same time agent may want to publish an explicit proposal
* I stay open to converse exchanging product Y for anything you would offer

One can already create Proposal for providing product Y and any other intent to receive, but to avoid creating a proposal for each intent, we could consider another broad proposal:
* I stay open to converse exchanging product Y for anything you would offer if I have current intent to receive it

Unless agent publishes an explicit statement about *staying open to converse exchanging...* currently no one can make any assumptions if they do or don't. To avoid spam I think we could actually recommend to assume that they don't, otherwise they could have stated that interest, in a way *opt-in*. Otherwise we may need to provide a way to *opt-out* and have a way to express *not staying open to converse exchanging ...*. 

Imagine that with proliferation of cryptocurrencies, agent would start receiving hundreds of proposal offering all those cryptocurrencies in exchange for their products. That looks spam-my to me and many agents would probably don't go more open than: "I stay open to converse exchanging product Y for anything you would offer if I have current intent to receive it"

examples/simple-proposals.yaml
fosterlynn: I think maybe I named this badly.  This is only one side, it is not a "rate" in that sense.  This is for the "price list" use case, like 25# case carrots.  In the other intent would be $32.  So 25# case carrots is re-usable, and would be there as long as carrots are in stock for the season (every week a new price list), and also re-usable year after year.  The $ side might vary somewhat over the season, but not necessarily.

It is also a good portion of retail e-commerce, which list unit prices, and you can buy however many you want, within some large limit.  Also might be most of retail, although those usually aren't published outwardly facing on a computer.  But still it is the same data.

I don't totally mind putting it in the ProposedIntent, but think it might be more confusing to apps, and don't see the need.

Maybe "unitQuantity" - or I think you had "scalableQuantity" or similar....?

fosterlynn:
>I stay open to converse exchanging product Y for anything you would offer

Seems like notes could handle this?  Or would you prefer something explicit on the reciprocal Intent?  Or... ?

>I stay open to converse exchanging product Y for anything you would offer if I have current intent to receive it

Suggestion?

For this, it seems like different apps can interpret different ways and then people can use the app they prefer for how they want to do intent based exchange.  I can picture an app that looks at all open intents, and matches in the way you suggest.  Which would be used by a group that wants to organize their exchange in this way with each other - i.e. put all their offers and requests out there.  Do we actually have this use case anywhere yet?

Which reminds me, I didn't include "periodicity", but mostly because I don't know of a use case, so would prefer to get the basics out there and then work it through with the next batch of users and see what they want to do.

>To avoid spam I think we could actually recommend to assume that they don't, otherwise they could have stated that interest, in a way opt-in.

Agree.

examples/simple-proposals.yaml
elf-pavlik: What if for the same intent of providing *undefined* quantity of 25# boxes of carrot, farm would want to have two different proposals (one for bulk amounts)?
I think this kind of information stays related to proposal.
I guess intent could have information about 'atomic quantity', commitments satisfying it would need to always have quantity multiplied by positive natural number.

elf-pavlik:
> For this, it seems like different apps can interpret different ways and then people can use the app they prefer for how they want to do intent based exchange.

Person can only control which app they use themselves, all the other people in the ecosystem can use whatever apps they choose. IMO explicit information about what kind of conversation agent publishing a proposal would like to engage in should stay in data.

fosterlynn:
>Person can only control which app they use themselves, all the other people in the ecosystem can use whatever apps they choose. IMO explicit information about what kind of conversation agent publishing a proposal would like to engage in should stay in data.

OK.  Suggestion how to represent in the data?  Or wait for a use case?

examples/simple-proposals.yaml
fosterlynn: I think an intent for undefined bulk amounts is a different intent, and different proposal.  And even a different resource.

elf-pavlik:
It sounds like you already have a use case with that maple syrup where agent offers specific product and would like to opt-in to leave other side open. I'll look at how to represent it in data.

fosterlynn:
>It sounds like you already have a use case with that maple syrup where agent offers specific product and would like to opt-in to leave other side open. I'll look at how to represent it in data.

Actually that use case is that she makes occasional requests, usually for specific things for her kids.  She always offers maple syrup in return.  She never just offers maple syrup as the primary, and things are not left open. (Her husband makes and sells maple syrup as part of a local maple syrup coop he helped start. They just have extra around to offer in trades or give as gifts.)

But please still look at how you would represent your use case in the data, if you like.

examples/simple-proposals.yaml
fosterlynn: >I don't totally mind putting it in the ProposedIntent, but think it might be more confusing to apps, and don't see the need.

Actually, I think it needs to stay in Intent, otherwise the Intent is not re-usable - or it is re-usable but not usefully.  It also fits from a modeling perspective in the same entity that has the resource class/spec.  Just needs to be distinguished from other quantity(s).

fosterlynn:
By the way, I'm not tied to having availableQuantity, I haven't seen an actual use case for that, so if you prefer, we can get rid of that for now.

Another thought: I'd like to agree to make this work for all the different use cases we know about, instead of for the respective use cases we are focused on.  And I'm trying to get some more info from at least one of our prospective users over in holochain.  Any objections to these at this point?

So here's an attempt:
* timebank
* small ads (like @matslats credit commons)
* price lists (wholesale)
* retail e-commerce
* @elf-pavlik use case(s) for agent planned ongoing intents; and for multi-lateral barter arrangements

(Please clarify yours and any others you are aware of...)

I'm realizing I'm missing a "publish to" on Proposal, which we talked about some time ago.  Might need to be range owl:Thing, I can think of a number of ways people might want to do that.  But I'm also interested in getting whatever minimum we can agree on out there, happy to leave things off for now.

examples/simple-proposals.yaml
fosterlynn: Changed name to unitQuantity.

fosterlynn:
In terms of supporting different use cases, which can imply different meanings to the quantity field: So far, I've created different properties to handle quantity.  Other ways we could do it: A code property on Proposal; a code property on Intent.  I think it is possible people could want to use both unitQuantity and flowQuantity, although maybe not likely.

elf-pavlik:
> In terms of supporting different use cases, which can imply different meanings to the quantity field: So far, I've created different properties to handle quantity.

I think properties have to have universal meaning which doesn't change, if we want different meaning we need a different term.

In your examples I miss scalable quantities in `ProposedIntents` which would scale proportionally when creating `Agreement`. Since quantities of intents stay independent from all the proposals which could build around them, those proposals need to have all the information for scaling quantities.

fosterlynn:
>In your examples I miss scalable quantities in ProposedIntents which would scale proportionally when creating Agreement. Since quantities of intents stay independent from all the proposals which could build around them, those proposals need to have all the information for scaling quantities.

This is the same as the discussion starting here, right? https://github.com/valueflows/valueflows/pull/503#discussion_r292100968

Scalable quantities are in Intent itself, just like other flow quantities, name atm unitQuantity, just fixed in the examples.  I guess we still have the disagreement we have had for awhile about how to think about intents expressed using "unit pricing" (please take that as not necessarily money, that is just the standard term)?  Anyhow, unit quantities would scale when commitments are created.

To me, unit quantities are part of the intent, independent of proposals, the same as any other quantification of the intent.  If there is no unit quantity, there is no quantity at all, in most use cases.  That would make Intents a lot less re-usable in different Proposals than if the unit quantity is included.

fosterlynn:
>I think properties have to have universal meaning which doesn't change, if we want different meaning we need a different term.

OK, then I will stick with unitQuantity (or some different name from the flow quantity which is not scalable).

fosterlynn:
I think adding something like `totalQuantity` to Intent would be a start toward supporting your use cases @elf-pavlik .  

You'll be happy to know I talked to a guy the other day who has something like that in mind, or at least wants to have flexibility to list several possible reciprocal intents to his main intent, in this case fish, as he is a fisherman.  So money or several things he could use that people might have.  

We could re-think the and/or modeling decision (and's are part of the same proposal; or's are different proposals).  Or we could create a different thing than Proposal to group these kinds of multi-lateral intents; or a different kind of Proposal.  It all does seem like proposals to me, but somewhat differently structured by necessity underneath the proposal.

fosterlynn:
OK, I added a possibly crazy experiment to support multi-lateral exchange proposals.  Example included, one person publishing a (very abbreviated) set of intents for the year.  Trying to see if this can all be modeled using Proposal and Intent.  What do you think?

Note: I have syntax errors in the ttl file, took a stab at how to do an enum range, probably messed that up.  Can track it down later, first trying to get the ideas out there for your feedback.

The other thing I added (thus the enum) was a "and" vs "or" choice on the ProposedIntent.  This could support the multi-lateral use case, plus the one I mentioned earlier where someone just wants to express several options for reciprocal intents without doing different proposals.  This isn't complete logic obviously, but I think adds quite a bit of flexibility to design different use cases.

Very interested in your feedback.

bshambaugh:
Is this helpful?
https://github.com/researchstudio-sat/webofneeds/blob/master/webofneeds/won-utils/won-utils-conversation/doc/wonprotocolhighlevel/highlevelprotocolpackagesreturningagreementtriples.png
-Brent Shambaugh

GitHub: https://github.com/bshambaugh
Website: http://bshambaugh.org/
LinkedIN: https://www.linkedin.com/in/brent-shambaugh-9b91259
Skype: brent.shambaugh
Twitter: https://twitter.com/Brent_Shambaugh
WebID: http://bshambaugh.org/foaf.rdf#me

bshambaugh:
More information:

https://github.com/researchstudio-sat/webofneeds/tree/master/webofneeds/won-utils/won-utils-conversation/doc/test

-Brent Shambaugh

GitHub: https://github.com/bshambaugh
Website: http://bshambaugh.org/
LinkedIN: https://www.linkedin.com/in/brent-shambaugh-9b91259
Skype: brent.shambaugh
Twitter: https://twitter.com/Brent_Shambaugh
WebID: http://bshambaugh.org/foaf.rdf#me

elf-pavlik:
> Scalable quantities are in Intent itself, just like other flow quantities, name atm unitQuantity, just fixed in the examples. I guess we still have the disagreement we have had for awhile about how to think about intents expressed using "unit pricing" (please take that as not necessarily money, that is just the standard term)? Anyhow, unit quantities would scale when commitments are created.

This `unitQuantity` doesn't have anything to do with exchanges, it only provides information about multiplier for quantity of satisfying intent or fulfilling event.

> To me, unit quantities are part of the intent, independent of proposals, the same as any other quantification of the intent. If there is no unit quantity, there is no quantity at all, in most use cases. That would make Intents a lot less re-usable in different Proposals than if the unit quantity is included. [edit: well, they would be re-usable, but not that meaningful, and the quantity would need to be duplicated in all the different proposals.

If Alice wants to receive potatoes and bananas and provide faircoins, she can have one intent to provide faircoins (x in this month). Both proposals faircoins for potatoes and faircoins for bananas need to include 'scalable quantities' based on which one can generate agreement. Quantity on intent only provides information about how much needs to get satisfied (later fulfilled) with commitments appearing as clauses of agreements, plus discussed `unitQuantity` which constrains multiplication on those satisfying commitments. Each of those agreements could get created based on any number of proposals for that intent. With your example of providing fish, fisherman could have intent to provide certain amount of fish and two proposals fish for faircoin and fish for labor. Each proposal would need to provide scalable quantities to calculate how many faircoin for a kilogram of fish and how many hours of labor for a kilogram of fish. Commitments in both versions of agreements would fulfill the same intent to provide certain amount of fish (in total).

fosterlynn:
>This unitQuantity doesn't have anything to do with exchanges, it only provides information about multiplier for quantity of satisfying intent or fulfilling event.

Yes, it is the quantitative aspect of many intents.  So, we agree on this?  I can't tell if you think that is a property we can include?

>If Alice wants to receive potatoes and bananas and provide faircoins, she can have one intent to provide faircoins (x in this month).   

That is one use case, but not all use cases.  And it isn't any use case that I have seen in real life at this stage of people working with intents.  I'm happy to support it, but not happy to say people have to do it that way. 

>Both proposals faircoins for potatoes and faircoins for bananas need to include 'scalable quantities' based on which one can generate commitments in agreement.

So that is what `unitQuantity` is for.  I don't have any problem with people including both `flowQuantity` (or whatever name) and `unitQuantity` in one intent.  Would that work for you?

fosterlynn:
>Is this helpful? https://github.com/researchstudio-sat/webofneeds/blob/master/webofneeds/won-utils/won-utils-conversation/doc/wonprotocolhighlevel/highlevelprotocolpackagesreturningagreementtriples.png

@bshambaugh thanks, and sorry I missed this earlier!  I think it is helpful for the Conversations for Action piece more, so I'll put out a comment there to reference your comment.  Also, I looked briefly in your repo for data elements for Proposals, if there are such a thing, can you point me to them?  Thanks.

elf-pavlik:
> Yes, it is the quantitative aspect of many intents. So, we agree on this? I can't tell if you think that is a property we can include?

If we talk about satisfying and fulfilling intents in general (any verb eg. produce, consume, work, use) and `unitQuantity` means that satisfying commitment and/or fulfilling event hast to have quantity that divides by that `unitQuantity` without a reminder. In some way a 'batch size'. Again it doesn't come as anything specific to agreements and *provider* / *receiver*.

> That is one use case, but not all use cases. And it isn't any use case that I have seen in real life at this stage of people working with intents. I'm happy to support it, but not happy to say people have to do it that way.

Everyone can create intent and only use it with one proposal, at the same time applications have to anticipate that any number of proposals can reference the same intent, which means none of the information 'on' the intent stays specific to any of the proposals.


>> Both proposals faircoins for potatoes and faircoins for bananas need to include 'scalable quantities' based on which one can generate commitments in agreement.

> So that is what unitQuantity is for. I don't have any problem with people including both flowQuantity (or whatever name) and unitQuantity in one intent. Would that work for you?

`unitQuantity` only puts constrain on how quantity of commitment can get multiplied in any proposal which references that intent, or any intent in general even *produce* that commitment gets created from recipe. At the same time it doesn't have anything to do with what we could call 'exchange rate' which stays specific to proposal. In other words none of the quantities on the intent have anything to do with 'exchange rate' since it stays specific to a proposal. What you call here `ProposedIntent` stays specific to proposal and anything specific to proposal, like quantities needed to get something like 'exchange rate' could go there.

fosterlynn:
>If we talk about satisfying and fulfilling intents in general (any verb eg. produce, consume, work, use) and unitQuantity means that satisfying commitment and/or fulfilling event hast to have quantity that divides by that unitQuantity without a reminder. In some way a 'batch size'. Again it doesn't come as anything specific to agreements and provider / receiver.

Yes, commitment would have to have a quantity that divides by the unitQuantity, which is basically 1.  The unitQuantity would generally be the smallest quantity the provider wishes to provide, although maybe for work by the hour it could be smaller easily.  (I don't see it being used much if at all for a receiver.  They would want to say how much they actually want.)  I'm not trying to make it anything specific to agreements, so I am missing that point.  But it has everything to do with the provider on an intent.  The use case is wholesale price lists such as discussed here https://github.com/valueflows/valueflows/issues/304 (and elsewhere I pasted an example, can't find easily now), and most of new non-custom retail e-commerce, or maybe offline point of sale for that matter.

Do you object to supporting those use cases?  If not, how would you support it cleanly and simply?

>Everyone can create intent and only use it with one proposal, at the same time applications have to anticipate that any number of proposals can reference the same intent, which means none of the information 'on' the intent stays specific to any of the proposals.

It stays specific to all the proposals that are related to it, right?  I see unitQuantity working fine for that, in the use cases it is meant to support.

>unitQuantity only puts constrain on how quantity of commitment can get multiplied in any proposal which references that intent

Correct.  I don't see a problem with that.  Many intents won't make any sense with unitQuantity.  And in the use cases I'm talking about, putting it for example on ProposedIntent would make the Intent much less re-usable because now the farmer on my price list example can simply carry his intent forward each week to a new price list (Proposals) until the price changes, which is not very often, usually not even in several years.  Vs. having to enter the unit for sale and the price again every week, ouch.

>any intent in general even produce that commitment gets created from recipe

I don't see produce Intents needing a unitQuantity, and see that as part of very different use cases - perhaps job shops that do custom work for one agent kind of thing.

elf-pavlik:
>> any intent in general even produce that commitment gets created from recipe

> I don't see produce Intents needing a unitQuantity, and see that as part of very different use cases - perhaps job shops that do custom work for one agent kind of thing.

> Yes, commitment would have to have a quantity that divides by the unitQuantity, which is basically 1. 

For intents with action *produce* unitQuantity would work in exactly the same way, quantity of the commitment would have to divide by the unitQuantity without a reminder.

-------------

Can we agree that unitQuantity doesn't come as any factor in calculating 'exchange rate' for the proposal? unitQuantity does constraint how quantity of the commitment satisfying that intent has to divide, but it doesn't directly impact calculating 'exchange rate'. If we take proposal referencing one intent to provide and one intent to receive, we can have any number of different proposals each referencing any one of those two intents as well, for that reason intent itself may constrain how quantity of commitment satisfying it needs to divide, but can't take any exchange related responsibility so also doesn't provide any information needed for calculating 'exchange rate'.

fosterlynn:
>Can we agree that unitQuantity doesn't come as any factor in calculating 'exchange rate' for the proposal?

Well, I see unitQuantity used only as one side of an exchange rate - not as calculating the exchange rate, but just expressing it.  I see it only on Intent atm.  So I don't really understand your statement.  If you see it used beyond that, please propose a model and a definition.

Here is the local use case I am trying to speak to where I am quite familiar (but as mentioned before, this kind of unit pricing probably will also serve most of e-commerce).  Here are 2 weekly price lists from two suppliers for our buying club as well as regional food coops and I don't know what else.  Everyone who regularly buys from them gets the same price list.  They have a limited set of buyers, and know who they all are and about how much they tend to want, and they also deliver to them weekly.  The main thing that changes weekly is the availability, not the pricing or packaging of veggies.  So this is about established supply chains.
price-list-do
price-list-k

(The number available is a new thing on the second price list.)

How would you want to support this local use case?

elf-pavlik:
> Well, I see unitQuantity used only as one side of an exchange rate - not as calculating the exchange rate, but just expressing it. I see it only on Intent atm. It would be used to calculate the quantity on the Commitments though, like multiply each unitQuantity by 6 or whatever. So I don't really understand your statement. If you see it used beyond that, please propose a model and a definition.

That means that we don't have common definition of unitQuantity. I see it just as a way to express that quantity on commitment has to divide but the unitQuantity without a reminder. It doesn't have anything to do with exchanges.

> Here is the local use case I am trying to speak to where I am quite familiar (but as mentioned before, this kind of unit pricing probably will also serve most of e-commerce). Here are 2 weekly price lists from two suppliers for our buying club as well as regional food coops and I don't know what else. 

Let's take your use case and make variant of it. Coop has those intents for upcoming cycle (let's say a week).
1. provide 200kg of Broccoli
2. provide 100kg of Chard
3. provide 50 Faircoins
4. provide 800 LETS
5. receive 50 Faircoins
6. receive 800 LETS
7. receive 40h of work

Now they have a proposals for all possible combinations of provide & receive pairs, except of currency for currency: 1-5, 1-6, 1-7, 2-5, 2-6, 2-7, 3-7, 4-7. I think it shows clearly that those 7 intents can't include any 'exchange rate' information since it differs for each of the proposals that reference those intents. Since that 'exchange rate' information stays specific to proposal, it could go into ProposedIntent not the Intent itself.

At the same time, someone can choose to create multiple intents to receive the same classification of a resource in the same period (eg. some currency) and choose to reference it from only one Proposal. Still they need to put information defining 'exchange rate' in ProposedIntent so the same application logic can handle both approaches.

Does above clarify why 'exchange rate' stays Proposal dependent and why to propose same intent in multiple proposals, the intent itself can't take any responsibility for 'exchange rate'?

fosterlynn:
>That means that we don't have common definition of unitQuantity.

This is how I have done the definitions in this PR.  (Hmmm, gotta remember to fix that typo.)

* unitQuantity  "The base quantity of the offer or request, which can be multipied to create a commitment; commonly seen in a price list or e-commerce." .

* totalQuantity  "The total quantity of the intent, allowing multiple commitments for part of that quantity; commonly seen in multi-lateral agreement arrangements." .

Please help clarify the definitions if they are not clear.

I put `totalQuantity` and a separate example for it in the PR to cover the multi-lateral proposals.  We don't need `unitQuantity` for that use case.

>Let's take your use case and make variant of it.

Feel free.  Or let me know what you don't like about how I handled the multi-lateral use case in the PR.  But let's not take every use case and make it into that use case.  As far as I can tell, people have very different requirements and economic experiments that are all Intents, and I'd like to support all of them.

fosterlynn:
>Or let me know what you don't like about how I handled the multi-lateral use case in the PR.

One missing piece might be periodicity.

elf-pavlik:
> Feel free. 

I already did described variant in my previous comment, I hope it clearly shows why no 'exchange rate' information can go directly on Intent.

> unitQuantity "The base quantity of the offer or request, which can be multiplied to create a commitment; commonly seen in a price list or e-commerce." .

This doesn't work when one creates multiple exchange proposals for the same intent, 'exchange rate (ratio)' stays specific to proposal so it could go in ProposedIntent but not directly on intent.

> totalQuantity "The total quantity of the intent, allowing multiple commitments for part of that quantity; commonly seen in multi-lateral agreement arrangements." .

How do you see it differ from whatever we end up to define quantities on flows, i think your last approach uses *flowQuantity*, i think that supposed to mean 'total quantity' for flow (I, C, E) which can gets partially satisfied and fulfilled (respecting multiplier which I would consider something like unitQuantity).

fosterlynn:
>> totalQuantity "The total quantity of the intent, allowing multiple commitments for part of that quantity; commonly seen in multi-lateral agreement arrangements." .

>How do you see it differ from whatever we end up to define quantities on flows, i think your last approach uses flowQuantity, i think that supposed to mean 'total quantity' for flow (I, C, E) which can gets partially satisfied and fulfilled (respecting multiplier which I would consider something like unitQuantity).

The way I see it is that we need 3 quantities for different ways of expressing intents (not necessary for other flows).  From smallest to largest, naming can be changed of course:
* unit - smallest amount offered, assumed people can request multiples of those for commitments (example: the veggie suppliers above, hourly plumbing work, and any place multiples of the same thing are regularly produced and offered)
* flow - the exact amount offered or requested, assuming that quantity would in most cases continue on to commitments (example: I clean out my garage and want to offer some specific things, often used)
* total - more of a budgeted quantity for a resource classification for a period, which assumes that commitments will address part of that (example: your variant above, for proposing multi-lateral agreements that will span some period of time and support some portion of a person's life)

I think those will be mostly mutually exclusive on an intent, but not always.  But they do address different use cases in how we know and hope intents will be used.

fosterlynn:
Specifically:

>Let's take your use case and make variant of it. Coop has those intents for upcoming cycle (let's say a week).

I don't think this is a real use case.  Agents involved in this kind of supply chain don't have time to have conversations for action on everything all the time, especially involving more than 2 agents at a time.  They just make regular orders as needed, and things tend to be somewhat predictable.  If not, then they coordinate further, but not usually.

The solawi's (CSA's) who plan with their members at the beginning of the season might be a good use case for this, much better than the one I did as an example.  I believe they budget their inputs and outputs, and then work with the eaters to figure out what they should put in to support the farm; and work out a season's worth of planning, including all the commitments between the farm and the eaters.  But we could find out more too.

I don't believe there is any need for unitQuantity in this use case.

elf-pavlik:
Just to clarify, Intent can work in planing without ever having any proposal  referencing it. I think we should clearly separate responsibilities related to proposals for exchange, and responsibilities related to more general planning. 

> unit - smallest amount offered, assumed people can request multiples of those for commitments (example: the veggie suppliers above, hourly plumbing work, and any place multiples of the same thing are regularly produced and offered)

I see a place for it, but it seems enough here to just say that when used on intent, commitment just need to have quantity resulting from multiplying that unit. Again nothing specific to having intents used in proposals.

> flow - the exact amount offered or requested, assuming that quantity would in most cases continue on to commitments (example: I clean out my garage and want to offer some specific things, often used)
total - more of a budgeted quantity for a resource classification for a period, which assumes that commitments will address part of that (example: your variant above, for proposing multi-lateral agreements that will span some period of time and support some portion of a person's life)

I still don't see where you see difference here, if instead of considering satisfying intent with commitments, we would look at fulfilling commitment with events, how flow and total would affect fulfillment when used on commitment ?

fosterlynn:
>Just to clarify, Intent can work in planing without ever having any proposal referencing it. I think we should clearly separate responsibilities related to proposals for exchange, and responsibilities related to more general planning.

Agree. 

>I see a place for it, but it seems enough here to just say that when used on intent, commitment just need to have quantity resulting from multiplying that unit. 

This is its current definition.
```
vf:unitQuantity a           owl:ObjectProperty ;
        rdfs:label          "unit quantity" ;
        rdfs:domain         vf:Intent ;
        rdfs:range          qudt:QuantityValue ;
        rdfs:comment        "The base quantity of the offer or request, which can be multiplied to create a commitment; commonly seen in a price list or e-commerce." .
```
Should something be added?

>Again nothing specific to having intents used in proposals. 

What is it if not a proposal?  To me, it is just a slightly different way of expressing a proposal.  I'm open to defining something else, although I'm afraid of getting too many concepts in the Intent space.  But what is your suggestion?

>I still don't see where you see difference here

Would you like me to remove totalQuantity?  Fine with me.

>All above still doesn't touch in any way requirement for proposals to provide quantities for calculating 'exchange rates'.

Not sure what that means?  Unit exchange rates are of course optional.

elf-pavlik:
> The base quantity of the offer or request, which can be multiplied to create a commitment; commonly seen in a price list or e-commerce.

Wouldn't the same work for Commitment ? Also 'base quantity' doesn't say anything IMO, if it means that quantity of satisfaction and fulfillment need to come by multiplying that quantity we should state so. Also I don't think we should use terms like offer and request and just stick with flow.

> What is it if not a proposal? To me, it is just a slightly different way of expressing a proposal. I'm open to defining something else, although I'm afraid of getting too many concepts in the Intent space. But what is your suggestion?

Proposal provides a sort of a template for how to create an agreement with commitments that satisfy specific set of intents. In case of proposal for exchange it will group intents where specific agent appears as *provider* on some and *receiver* on some. 

Bare Intent only expresses what agent plans but doen't directly suggest anything on how they want to reach commitments with another agents.

> Would you like me to remove totalQuantity? Fine with me.

I think however we end up representing planned quantity in a flow it already represents the total. Discussed unitQuantity only adds information on required quantities of satisfying commitments and fulfilling events.

> Not sure what that means? Unit exchange rates are of course optional.

As I mentioned many times in this thread unitQuantity doesn't provide any information to calculate 'exchange rate', since no information on Intent stays specific to any given Proposal while 'exchange rate' stays specific to particular proposal. Quantities used for calculating 'exchange rate' could go ProposedIntent since they stay specific to Proposal. In other words, algorithm which calculates 'exchange rate' doesn't care at all about unitQuantity, it needs some other quantity on ProposadIntent.

In simple dual exchange (special case) one would get 'exchange rate' by dividing some reciprocity quantity from ProposedIntent A (referencing intent to *provide*) by some reciprocity quantity from ProposedIntent B (referencing intent to *provide*). Later to make commitments unitQuantity from those two intents (not specific to any proposal) would come in place to know all the possible valid quantities for commitments, and would need to get used together with those reciprocity quantities from both ProposedIntents (specific for this proposal).

Until we can agree that unitQuantity doesn't affect 'exchange rate' and reciprocity quantities which affect 'exchange rate' can't stay directly on Intent (could go on ProposedIntent), I think we understand those concepts differently.

examples/simple-proposals.yaml
```diff
@@ -0,0 +1,262 @@
...
+    publishes: farm:c7897c39-7f05-4a5d-a487-80e130a24345 # the intent
+    reciprocal: false
+
+  - '@id': farm:0f563083-7f05-4a5d-a487-80e130a24345
+    '@type': Intent
+    name: Payment
+    action: transfer
+    receiver: https://farm.example/
+    resourceClassifiedAs: https://www.wikidata.org/wiki/Q4917 # US Dollar
+    unitQuantity:
```
elf-pavlik: If this supposed to express $18 for 1 case of carrots it can't go directly on Intent. Here it would mean that agent only accept USD in amounts that divide by 18.

examples/simple-proposals.yaml
```diff
@@ -0,0 +1,262 @@
...
+      qudt:unit: unit:Number
+      qudt:numericValue: 18
+
+  - '@id': farm:f325a950-5737-488d-b122-8f21114d0eb0
+    '@type': ProposedIntent
+    publishedIn: farm:da7bfa38-a1b2-4f6f-a896-b9bf1ec8807a # the proposal
+    publishes: farm:0f563083-7f05-4a5d-a487-80e130a24345 # the intent
+    reciprocal: true
```
elf-pavlik: on could add here something like

reciprocityQuantity:
     qudt:unit: unit:Number
      qudt:numericValue: 18

Since this stays specific to the proposal but not the intent itself.

examples/simple-proposals.yaml
```diff
@@ -0,0 +1,262 @@
...
+    resourceSpecifiedAs: farm:2402b57c-1c45-423a-acb2-f50961c97d65 # 25 pound case organic carrots seconds
+    unitQuantity:
+      qudt:unit: unit:Number
+      qudt:numericValue: 1
+
+  - '@id': farm:a8236bbb-81e0-422d-9861-56d2417db0fb
+    '@type': ProposedIntent
+    publishedIn: farm:da7bfa38-a1b2-4f6f-a896-b9bf1ec8807a # the proposal
+    publishes: farm:c7897c39-7f05-4a5d-a487-80e130a24345 # the intent
+    reciprocal: false
```
elf-pavlik: This doesn't provide any information on how to calculate USD to case of carrots exchange rate. It would need to have something like 

reciprocityQuantity:
qudt:unit: unit:Number
qudt:numericValue: 1

examples/simple-proposals.yaml
```diff
@@ -0,0 +1,262 @@
...
+    eligibleLocation: map:566a2c01-96f8-4b3c-adcf-f197eced3092 # Madison, WI
+
+  - '@id': farm:c7897c39-7f05-4a5d-a487-80e130a24345
+    '@type': Intent
+    name: Carrots case 25#
+    action: transfer
+    provider: https://farm.example/
+    resourceClassifiedAs: wd:Q81 # carrot
+    resourceSpecifiedAs: farm:2402b57c-1c45-423a-acb2-f50961c97d65 # 25 pound case organic carrots seconds
+    unitQuantity:
```
elf-pavlik: This just means that all commitments that satisfy this intent need to have a natural number as quantity (divides by 1). Those commitments could get made based on any number of proposals referencing this intent.

examples/simple-proposals.yaml
elf-pavlik: As with any Intent satisfying commitments could get made based on any number of proposals referencing this intent. Not information directly on intent stays specific to any of the proposals.

fosterlynn:
>Until we can agree that unitQuantity doesn't affect 'exchange rate' and reciprocity quantities which affect 'exchange rate' can't stay directly on Intent (could go on ProposedIntent), I think we understand those concepts differently.

I guess we do.  I don't understand why you won't allow a new different property to represent another use case that is still conceptually a proposal of reciprocal intents.  So then, please detail how you would support the price list use case?  Understanding that in that use case, what I am calling unitQuantity will stay the same across many proposals and should be reusable.

examples/simple-proposals.yaml
fosterlynn: 18 whats?

examples/simple-proposals.yaml
fosterlynn: 1 what?

elf-pavlik:
I've made comments inline that unitQuantity which does not affect 'exchange rate' in any way can stay directly on Intent but reciprocityQuantity which only defines 'exchange rate' would need to go on ProposedIntent since it stays specific to proposal. I understand that by 'price list' use case you mean the one with 18USD for 12# case of carrots.

examples/simple-proposals.yaml
elf-pavlik: Intent referenced by publishes already deifnes what

fosterlynn: That is way too confusing and complex - nowhere else do we have quantity separated from quantity of what.  There is no need to calculate an exchange rate.  It is totally defined by the unitQuantity on the intents themselves.

fosterlynn:
>which only defines 'exchange rate' would need to go on ProposedIntent since it stays specific to proposal

what if you want to reuse it?  just like anything else

what if the unitQuantities are 1 case of carrots for 3 hours of help harvesting?  (they have a lot of carrots and can use a lot of help; and they also might use the 3 hours of help harvesting intent for multiple proposals for different veggies, since it is a work shift they use to schedule harvesting help?)

what if they grow 30 kinds of potatoes that are all the same price per case?

why do you think that a proposal for my 1 used bike (flowQuantity) for a morning of help with childcare (flowQuantity on a reciprocal intent) is so essentially different from a proposal for 1 case of carrots (unitQuantity) for $32 (unitQuantity on a reciprocal intent)?

examples/simple-proposals.yaml
elf-pavlik: Noting on intent can define 'exchange rate' since it stays specific to just one of proposals referencing that intent.

fosterlynn: Please define 'exchange rate' as you are using it.

elf-pavlik: In this case we have: 18USD for 1 case of carrots

While the very same intent to provide carrots could get used at the same time in proposals like: 1h of work for 2 cases of carrots. And in the same way, the very same intent to receive USD can get used at the same time with proposals like 30USD for 1 case of potatoes. In the example of 'price list' you shared, one should have possibility to have just one intent to receive USD and reuse it in all the proposals for each kind of produce they provide.

Last but not least, 'exchange rate' while sufficient for 'price list' use case oversimplifies things a little. We can have proposals like 500g bar of chocolate for 80MXN + 30LETS (1 providing & 2 receiving), so this non-dual exchange would still scale proportionally each reciprocityQuantity from each ProposedIntent.

fosterlynn: If I offer my used bike (intent with Lynn provider), and request in return $40 (intent with Lynn receiver), is that an 'exchange rate'?  Still trying to make sure I understand your definition.

elf-pavlik: Yes, 1 (specific) bike for 40USD fits what i would call 'exchange rate' for that specific proposal.

> And in that case, would you put the quantity $40 on Intent or ProposedIntent?

Classification (wd:Q4917 # US Dollar) on Intent and unit:Number & value 40 on ProposedIntent. Agent could use the same intent to receive (wd:Q4917 # US Dollar) with 100 other proposals, and each proposal needs its own specific exchange rate.

fosterlynn: So then exchange rate applies whenever there is a reciprocal intent of any sort, correct?

elf-pavlik: I wouldn't call intent itself as 'reciprocal' since on itself it just represents some isolated part of agent's plan. Reciprocity appears when we have proposal for exchange, so proposal which references at least one intent where agent publishing proposal wants to *provide* and at least one intent where agent publishing proposal wants to *receive*. Again, intents directly (or flow of any stage in that matter) have no notion of reciprocity, it comes into play first in a proposal and later in agreement made based on that proposal.

fosterlynn: that's what i meant, of course - intent as part of a proposal - I'd still like to understand if that is how you think about it, I'm trying to think of a compromise way to model this, and hoping to come up with something simpler

> isolated part of agent's plan

that can be a little grandiose, don't you think? it might be what happened to me recently, my mom was asked to move to a smaller memory care unit from her assisted living apartment, and I gave away all her extra furniture when she moved, as quickly as I could so she could stop paying rent on her old apartment

I think it would be good to expand your view of the use cases, to cover what people are actually doing in these transitional times, in addition to ideas for how to put together some new over-arching economic methods...

elf-pavlik: > I think it would be good to expand your view of the use cases

As long as we don't put any proposal specific information directly on intent my approach seems to work just fine with all considered so far use cases.

> I'm trying to think of a compromise way to model this, and hoping to come up with something simpler

Simple check on any alternative approach. Will it work if in your farmer's produce UC agent providing that produce has one and only one intent to receive USD (for example intended USD income for current week long operating cycle).

fosterlynn: See gitter.

elf-pavlik:
Quick note on "to match or not to match", I think we could recommend matching software this approach:

1. Only match if proposal exist, never directly on intents - proposals act as invitation to match
2. Proposals can have three degrees of specificity:
  * all intents (provide & receive) specified
  * either provide or receive marked as 'catch all' agent's provide or receive intents respectively, other side specified
  * either provide or receive marked as 'open' (least specific possible), other side specified

This will create clear opt-in matching, so agents who want more or less specific matching need to create more or less specific proposal to invite engines to match.

examples/simple-proposals.yaml
fosterlynn: >Please also keep in mind that once quantity on intent (I think you call it flowQuantity) gets fully satisfied in commitments, other agents should consider all proposals that reference that intent as finished. In your approach, how do you represent how much of the flowQuantity on intent already got satisfied with commitments?

Since we didn't get to that in our chat, here are some cases where there is no availableQuantity (or flowQuantity) on offers.  All make-to-order scenarios (Dell back in the day, machine job shops, anything custom).  Any producer or distributor where they know they can make or purchase something if they do run out of stock. Any producer where they have known receivers and know they won't run out so it is not relevant to their proposal (like the first price list above). 

Also things change, especially estimates, and that is what the finished flag is for, there will be very few situations where software can make that assumption.  

Generally I think these are some of many reasons to not make assumptions about how intents will work for people, especially until we have way more experience with them.

fosterlynn:
>either provide or receive marked as 'catch all' agent's provide or receive intents respectively, other side specified
>either provide or receive marked as 'open' (least specific possible), other side specified

Don't understand this?  (I probably get the general idea, but not the data structure it uses.)

fosterlynn:
I'm experimenting with a model that I think will speak to both of our different emphasis on proposals/intents.  Basically it leaves the Intent itself to be where it is (part of a plan, large or small, etc.), and pulls out the pieces more relevant to a specific proposal.  The ProposedIntent can also pull the data it needs from the Intent.  Although I suppose we would need to include it there too.  But then the Proposal and ProposedIntent are what goes to apps that publish proposals.  Or maybe all 3 can go, but only parts of the Intent not overridden by the ProposedIntent.

On the left, the model currently in my PR; on the right, the direction I'm thinking of in this comment.

prop-intent

Might be some problems with how this gets implemented in distributed architecture, what is the independent view.

fosterlynn:
Unfortunately that doesn't give the user re-use of the pieces of their proposal, like in the price list use case.  Ah well, just looking for a workable compromise.

fosterlynn:
Another way of dealing with it perhaps is keeping Intent, Satisfaction with the model and definitions we have, thus pointing Intent forwards towards Commitments, and including only leaf-level operational items.  And adding something else behind Intent that is more part of larger planning, budgeting, etc.  Proposals would be for all intents that need matching; not for anything that is part of a plan but no publication for matching is required.

Something like this (approximately, and please ignore the bad naming).  The EstimatedFlow would have some kind of periodicity to it.  The Intent would not, other than what flows tend to express for times, reflecting how known they are, not periodicity per se.

prop-intent2

This assumes a lot of detail to be worked out when we get back to nesting of processes, higher level planning, etc.

fosterlynn:
>Another way of dealing with it perhaps is keeping Intent, Satisfaction on the leaf level operational side, thus pointing Intent forwards towards Commitments, and including only leaf-level operational items. And adding something else behind Intent that is more part of larger planning, budgeting, etc

I've thought more about this, and my preference would be that we think of things in more this way.  So the progression of flows would be something like:

estimated flow > intent > commitment > event

Intents would become something only used when seeking a match so that a commitment can be made, they would not be used for overall planning.  So sometimes things could progress directly from estimated flow to event, if there is no need for intent or commitment.  This would happen for example when I go up the hill to the local Amish farmstand to get some veggies.  They might be part of my overall plan, and even part of the overall community plan, but I just go get enough to eat from what is available.  If what I have a taste for isn't available, I get something else or go to the root cellar, otherwise I'd have to travel farther, not worth it.  

Just trying to give the flavor of flexibility here, I don't picture a planned economy from top to bottom - more something that we can make work within planetary and local limits, minimizing waste, with some nice level of personal choice as well as personal compromise.

The thing something like "estimated flow" is more like a budget or higher level agent plan, and might (or might not) be at a higher level in the taxonomy of resource classifications. This is part of the "composition" discussion, and maybe others.  I definitely support the requirement(s), but would like to separate it out.

This leaves us the possibility of simplifying the proposal/intent model, and defining it as only needed as initiation of a new agreement, with or without conversation for action involved.  This way, it will match the (exchange) agreement model better.  And do one thing well, instead of combining concepts, which we have been trying to do for some time.  This will especially help use cases which don't need both concepts.  And means that Intents can't be re-used in different proposals.  But they can support any form of expression of intent we can think of, adding quantity properties as needed, because they are not tied to a previous form of intents.

It also leaves us with Intents always being conceptually part of something proposed, but not necessarily needing a Proposal; only needing it for situations where the information about the proposal needs to be communicated because it is not already agreed on by the agents involved.

@elf-pavlik what do you think of this direction?  Do you see anything we will make impossible to do, if we move your use cases up to a previous step and a different part of the model?  Naming to be nailed down.

fosterlynn:
@bhaugen this one is not ready yet for review/merge.

Issue #486, opened 2019-04-09
Adding optional locations to `EconomicEvent`; location properties


pospi:
It seems like a good idea to enable this to me- it would be particularly pertinent in logistics in order to more accurately inform calculations of GHG emissions as a by-product of transporting resources from A to B.

pospi:
I think the only curly requirements are to provide for multiple `location`s at the same coordinates, with different names- so that you can have "under the sink", "in the top drawer" as separate `Place`s. The spec looks pretty comprehensive, I'm not concerned about limitations. The bigger concern is pairing down what we really need to service as a minimum implementation vs the fully-featured version.

If there's an existing issue where this is better discussed, feel free to direct me that way. I know the group has been discussing new location ontologies for some time.

elf-pavlik:
I recall recent comment on this old issue: https://github.com/valueflows/forum.valueflo.ws/issues/27#issuecomment-461827306

> I think the only curly requirements are to provide for multiple locations at the same coordinates, with different names- so that you can have "under the sink", "in the top drawer" as separate Places. 

I think 'same coordinates' depends heavily on their precision. https://wiki.openstreetmap.org/wiki/Indoor_Mapping shows maps inside of buildings.

Document I linked in my previous comment mentions https://www.w3.org/2017/sdwig/bp/#spatial-relations
>* Topological relations describe the relationships between geometric objects that are invariant to rotation, translation and scaling. As such, topological relations can support qualitative spatial reasoning without reference to the geometries themselves; for example, to assert that object A touches object B. These relations, also known as “spatial predicates”, include concepts such as: equals, disjoint, intersects, touches, within, contains, overlaps and crosses.
>* Directional relations specify the relative direction between object and reference. Examples include: left, in front of and astern.
>* Distance relations specify how far the object is from the reference object. Examples include: at, nearby and far away.

We already rely on OWL-Time for [Topological temporal relations](https://www.w3.org/TR/owl-time/#topology), so we can define that some flows or processes have to start at the same time (while not providing specific beginning for any of them), or that some flow happens before or after another one.

I think we may need at leas to specify that one place contains another place.

fosterlynn:
Seems like we got a couple decision tracks emerging from the discussion.

[...]

PR #510, opened 2019-05-07
added observedLocation, put other Location refs back

fosterlynn:
Added observedLocation to EconomicEvent, per discussion https://github.com/valueflows/valueflows/issues/486.  Also put vf:Location back for now.  We can swap in another vocab for this if we want.  I'll add a few property references from other vocabs, unless you get there first @elf-pavlik .

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -378,12 +378,12 @@ vf:primaryLocation a        owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The main place an agent is located, often an address where activities occur and mail can be sent. This is usually a mappable geographic location.  It also could be a website address, as in the case of agents who have no physical location." .
 
-vf:observedLocation a       owl:ObjectProperty ;
-        rdfs:label          "observed location" ;
-        rdfs:domain         vf:EconomicEvent ;
+vf:atLocation a             owl:ObjectProperty ;
+        rdfs:label          "at location" ;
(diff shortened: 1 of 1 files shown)


elf-pavlik:
We started moving away from creating different properties for different types of flows. Can we also here just use same property that would reference a place from Intent, Commitment and Event? Maybe simple `vf:atLocation` ?

fosterlynn:
>Maybe simple vf:atLocation ?

Good idea.  Changed and pushed.

release-doc-in-process/all_vf.TTL
```diff
@@ -367,17 +367,24 @@ vf:definedQuantity a        owl:ObjectProperty ;
...
         rdfs:label          "primary location" ;
         rdfs:domain         foaf:Agent ;
-        #rdfs:range          vf:Location ;
+        rdfs:range          vf:Location ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The main place an agent is located, often an address where activities occur and mail can be sent. This is usually a mappable geographic location.  It also could be a website address, as in the case of agents who have no physical location." .
 
+vf:atLocation a             owl:ObjectProperty ;
+        rdfs:label          "at location" ;
+        rdfs:domain         [ owl:unionOf (vf:EconomicEvent vf:Commitment vf:Intent) ] ;  ;
```
elf-pavlik: somehow you have two semicolons at the end of the line

fosterlynn: aargh, thanks for catching!  fixed

PR #511, opened 2019-05-10
draft location spec

fosterlynn:
Getting this out here so we have something to eventually do on it.  @elf-pavlik please don't make it a priority, but if you could later on fix up the external spec part to be the right spec, that would be helpful.  And just double check it all.... thanks!

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -383,6 +383,7 @@ vf:url a                    owl:DatatypeProperty ;
 
 vf:name a                   owl:DatatypeProperty ;
         rdfs:label          "name" ;
+        rdfs:range          xsd:string ;
         vs:term_status      "unstable" ;
         rdfs:comment        "An informal or formal textual identifier for an object. Does not imply uniqueness." .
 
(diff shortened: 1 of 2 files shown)


fosterlynn:
P.S. I decided to keep vf:Location because we will have extra properties.  If using geo:SpatialThing is better practice, we can change it.

release-doc-in-process/all_vf.TTL
```diff
@@ -526,6 +527,13 @@ vf:realizationOf a          owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "This economic event occurs as part of this agreement." .
 
+vf:mappableAddress a        owl:ObjectProperty ;
+        rdfs:label          "mappable address" ;
+        rdfs:domain         vf:Location ;
```
elf-pavlik: nothing stops us from defining domain as `geo:SpatialThing`

fosterlynn: Ok, I changed it all to SpatialLocation in the ttl file; will need to also update in other places once we have this.

fosterlynn:
@elf-pavlik once you have hopefully fixed up the external-terms.md file (thanks!), maybe this is good?  I'll fix up other files that need to be in sync once the properties are confirmed.  Could be another PR; and the graphql is a separate repo, so that will be another PR.

release-doc-in-process/all_vf.TTL
elf-pavlik: `geo:SpatialThing` not `vf:SpatialThing`

fosterlynn: thanks, duh!   fixed

PR #520, opened 2019-07-03
clarify resource spec vs classif

fosterlynn:
I think this clarifies the resource classification and resource specification difference.  I got this idea from reading some of McCarthy's recent writings.  This seems a lot cleaner conceptually than something sort of like "it is on the web or it is either part of a recipe or not on the web".  From his doc:

* typify - the association between a concrete entity and the abstract specification of its essential and computed properties.  Typifying a concrete object, person, or occurrence attempts to specify properties that capture its archetypal essence.  This definition derives from Platonic philosophy.
* group - the association between a concrete entity and another concrete class that defines its aggregated and computed properties.  Typing and grouping are abstraction mechanisms that at the margin are very close to each other.  In the REA ontology, we consider grouping as that set of aggregation associations that lacks the archetypal essence validation.

RDF question:  I put ResourceSpecification and owl:Thing both on the range of resourceConformsTo.  Seems not quite right, but I do want to call out ResourceSpecification.  Also wondering if instead of owl:Thing, we should be using a uri for this and for resourceClassifiedAs?

### DIFF
--- a/docs/introduction/resources.md
+++ b/docs/introduction/resources.md
@@ -10,27 +10,30 @@ And we want knowledge to be freely available.
 
 Also, we prefer to think of use value, but economic resources also often have exchange value.
 
-#### The difference between a resource and its specification or classification
+#### The difference between a resource and its specification/classification(s)
 
-An economic resource is observable.  Its specification or classification defines what type of thing the economic resource is.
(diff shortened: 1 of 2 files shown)


fosterlynn:
I'll want to add process to this too, ref. https://github.com/valueflows/valueflows/issues/521.

PR #522, opened 2019-07-07
Abbreviated transfer as one event

fosterlynn:
I removed the POA experiment from my other transfer PR, trying to create a fairly vanilla version of transfer that moves beyond give-Transfer-receive.  I tried to do the minimum that would basically work, and that I hope we can more or less agree on as a direction.  And that is better than what we have. I think there will be a lot of situations where this is plenty.

Note that currentQuantity is possessed quantity in this PR.  So there may be a gap in reporting a standard balance sheet, but it could possibly be figured out with some digging into agreements.  And is probably less important at the moment.

@elf-pavlik do you think this is getting close to something we can start to move forward with?


fosterlynn:
Another idea: Thinking in terms of this simpler PR, I would definitely consider (possibly haven't thought through all the implications yet) having `transfer` and `possess` as actions, idea taken from your PR.  Transfer would be more for "ownership" or other transitional and future kinds of transfer that don't have or imply a commitment to return the resource, like stewardship or anything that completely removes it from the inventory of the providing agent.  Possess would not imply rights, although a referenced agreement could specify rights or not.  So that could cover "rights to use" potentially, when we get there.

Anyway, it would cover up one possible hole in this PR in terms of figuring out inventory possessed vs owned for operational and accounting needs - coordination, different responsibilities (like if a resource is in your warehouse or in transit), balance sheet, etc..

fosterlynn:
>Thinking in terms of this simpler PR, I would definitely consider (possibly haven't thought through all the implications yet) having transfer and possess as actions, idea taken from your PR.

Nah, then we have to decide if to allow more than one action (too much additional complexity imo) or have 2 events just for a retail transfer, also not the best.  So never mind this last comment, but I'd still like to get some minimum transfer in, even with the holes, it will be better than what we have, and has room to plug the holes by just adding on, I think anyway.

docs/introduction/flows.md
```diff
@@ -57,27 +57,26 @@ All flows use an action property to designate what the flow is doing and how it
...
+produce | Increment  | Output | Yes | N/A | Yes | N/A |
+consume | Decrement  | Input | Yes | N/A | Yes | N/A |
+use | No effect  | Input | No | No | Yes | N/A |
+work | No effect  | Input | N/A | N/A | Yes | N/A |
+cite | No effect  | Input | No | No | No | N/A |
+load | No effect  | Input | No | N/A | Yes | unload |
+unload | No effect  | Output | No | N/A | Yes | load |
+accept | No effect  | Input | No | N/A | Yes | improve |
+improve | No effect  | Output | No | N/A | Yes | accept |
+transfer | N/A | N/A | No | N/A | N/A | N/A |
```
bhaugen: Suggest removing possession and availability columns until issues resolved, and than
transfer | Decrement+Increment | N/A | No | N/A |
or
transfer | Decr.+Incr. | N/A | No | N/A |

PR #533, opened 2019-08-04
added possess action

fosterlynn:
This adds `possess` events.  Instead of allowing more than one action per event, or having a separate property (like `transferOf`) on transfers to designate transfer of possession and/or ownership, I used `triggeredBy` to put the events inline in a value flow, similar to what we did with transfers FOB in shipping.

A little messy - all the options so far are a little messy!  

This does not include how to name inventory, but probably should.  I.e. have a separate property for something like `currentQuantity` (possessed) and `currentAccountingQuantity` or similar.

Please review @bhaugen and see what you think.


fosterlynn:
Another question here: I think it would be OK in say the transfer-pos example to just use the transfer events and imply possession change, to keep things simpler when it works for the agents involved.  What do you think?

PR #552, opened 2019-08-12
resourceQuantity and effortQuantity

fosterlynn:
Do not merge yet!  Need discussion on unitQuantity.

Making 2 quantities for all flows, plus flow relationships (to simplify):
* effortQuantity (work and use, probably cite)
* resourceQuantity (everything else, including services, non-inventoried resources)

This should satisfy the drills and ouikit use cases where we might need to document using X quantity of resources for X time, or high level intents that X number of people are needed for X time.  But doesn't allow unlimited proliferation of quantity names in different vocabularies.

Note that effortQuantity does not have to be time based, although mostly it will be.  Could also be machine cycles, words of translation, etc.

Example yet to do, as well as any clarifications and additions to the text.


fosterlynn:
Put in a solution for unitQuantity.  Ready to merge when reviewed and approved.

PR #566, opened 2019-08-18
Fixes spec

fosterlynn:


fosterlynn:
Explanation: To get the generated protege doc in a place it could be served, linked from the gitbook, we put it out on the server for mikorizal.org, temporarily.  This is a hack on a hack, but should work for now.  We still need to look for a better way to generate a rdf-based spec from the ttl file.

PR #590, opened 2019-10-28
Time

fosterlynn:
* Added vf domain classes to the time elements.  
* Added vf:hasPointInTime, equivalent to time:inXSDDateTimeStamp.  Please review....
* Removed time:before and time:after.
* Added vf:due for due date/times.  (Was using time:before for that, incorrectly.)

### DIFF
--- a/examples/checkout-book.yaml
+++ b/examples/checkout-book.yaml
@@ -46,7 +46,7 @@
     resourceQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
-    time:before: 2019-05-17T18:00:00-5:00
+    due: 2019-05-17T18:00:00-5:00
 
   # resource after check out
 
--- a/examples/checkout-book.yaml
+++ b/examples/checkout-book.yaml
@@ -29,7 +29,7 @@
     realizationOf: library:f47064eb-7120-4b31-b882-770165901fe6
     action: transfer-custody
     resourceInventoriedAs: library:2402b57c-1c45-423a-acb2-f50961c97d65
-    time:inXSDDateTimeStamp: 2019-04-18T8:00:00-5:00
+    hasPointInTime: 2019-04-18T8:00:00-5:00
     provider: https://library.example/
     receiver: https://bob.example/
     resourceQuantity: 
@@ -72,7 +72,7 @@
     resourceQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
-    time:inXSDDateTimeStamp: 2019-05-14T10:20:00-5:00
+    hasPointInTime: 2019-05-14T10:20:00-5:00
 
   - '@id': urn:uuid:6f438393-7f87-4914-806c-e23a4fd15e89
     '@type': Fulfillment
--- a/examples/plan-from-recipe.yaml
+++ b/examples/plan-from-recipe.yaml
@@ -62,7 +62,7 @@
     '@type': Plan
     name: Delivery fiber
     skos:note: Make the optical fiber that makes the connection between the device and the transducer. 
-    time:before: 2019-04-08T17:00:00-5:00
+    due: 2019-04-08T17:00:00-5:00
     dcterms:created: 2019-04-01T10:35:00-5:00
 
   - '@id': sensor:a8236bbb-81e0-422d-9861-56d2417db0fb
--- a/examples/simple-plan.yaml
+++ b/examples/simple-plan.yaml
@@ -11,7 +11,7 @@
     '@type': Plan
     name: Anti climate change campaign materials
     skos:note: A set of printed materials for the campaign by Students for Peace and Justice
-    time:before: 2019-04-08T17:00:00-5:00
+    due: 2019-04-08T17:00:00-5:00
     dcterms:created: 2019-04-01T10:35:00-5:00
 
   - '@id': printer:02b39a30-3e04-4305-9656-7f261aa63c84
--- a/examples/workflow-stage-state.yaml
+++ b/examples/workflow-stage-state.yaml
@@ -162,7 +162,7 @@
   - '@id': mfg:8e5fe80d-a769-4bd5-89e5-2136d33eab9f
     '@type': Plan
     name: Buckets run number 381
-    time:before: 2019-04-08T17:00:00-5:00 # due
+    due: 2019-04-08T17:00:00-5:00 # due
 
   - '@id': mfg:b52a5815-fae9-43bf-be95-833b95dc0adb
     '@type': Commitment
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -55,7 +55,7 @@ vf:ProcessSpecification a   owl:Class ;
 vf:RecipeFlow a             owl:Class ;
         rdfs:label          "vf:RecipeFlow" ;
         vs:term_status      "unstable" ;
-        rdfs:comment        "The specification of a resource inflow to, or outflow from, a recipe process.." .
+        rdfs:comment        "The specification of a resource inflow to, or outflow from, a recipe process." .
 
 vf:ScenarioDefinition a     owl:Class ;
         rdfs:label          "vf:ScenarioDefinition" ;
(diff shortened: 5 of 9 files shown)


pospi:
Are we leaving `before` and `after` on `EconomicEvent`? I think there was a use-case there for recording loosely that an event happened before / after some time, rather than the exact time that it happened. Just checking whether that is still intended or if it was only ever intended to be used for due dates.

fosterlynn:
@pospi thanks for reviewing!
>Are we leaving before and after on EconomicEvent? I think there was a use-case there for recording loosely that an event happened before / after some time

We took time:before and time:after off because it was being used in a different way than the owl time ontology uses it.  Rather than putting something like that back on in the vf vocabulary that wouldn't connect with owl time for its comparisons of times, two of which are time:before and time:after, I started with just vf:due because it doesn't need the comparisons and we know we need that on the planning layer.  And I could picture needing both a due date and a more exact scheduling of commitments and intents.

You are right that there was a use case for looser time periods on events (in FairCoop when people logged general coordination or server admin responsibility for a month instead of 5 minutes here and half hour there - more like a "salary").  My opinion is that we can use hasBeginning and hasEnd for that on observation and planning, and make sure the definition is clear.  In owl time, they just delineate a temporal entity, without comment on what it used for.  (@elf-pavlik disagreed with that, so please speak up if you still disagree and we'll look for a different solution!)

examples/checkout-book.yaml
```diff
@@ -46,7 +46,7 @@
     resourceQuantity:
       qudt:unit: unit:Number
       qudt:numericValue: 1
-    time:before: 2019-05-17T18:00:00-5:00
+    due: 2019-05-17T18:00:00-5:00
```
jvanbockryck: We don't have "due" in VF. This should be:
---
before: 2019-05-17T18:00:00-5:00
---
this is "vf:before", not "time:before".

examples/plan-from-recipe.yaml
```diff
@@ -62,7 +62,7 @@
     '@type': Plan
     name: Delivery fiber
     skos:note: Make the optical fiber that makes the connection between the device and the transducer. 
-    time:before: 2019-04-08T17:00:00-5:00
+    due: 2019-04-08T17:00:00-5:00
```
jvanbockryck: Should be "before" instead of "due".

examples/simple-plan.yaml
```diff
@@ -11,7 +11,7 @@
     '@type': Plan
     name: Anti climate change campaign materials
     skos:note: A set of printed materials for the campaign by Students for Peace and Justice
-    time:before: 2019-04-08T17:00:00-5:00
+    due: 2019-04-08T17:00:00-5:00
```
jvanbockryck: Should be "before" instead of "due".

examples/workflow-stage-state.yaml
```diff
@@ -162,7 +162,7 @@
   - '@id': mfg:8e5fe80d-a769-4bd5-89e5-2136d33eab9f
     '@type': Plan
     name: Buckets run number 381
-    time:before: 2019-04-08T17:00:00-5:00 # due
+    due: 2019-04-08T17:00:00-5:00 # due
```
jvanbockryck: Should be "before" instead of "due".

release-doc-in-process/all_vf.TTL
```diff
@@ -323,19 +323,57 @@ vf:containedIn a            owl:ObjectProperty ;
         rdfs:comment        "Used when a stock economic resource contains units also defined as economic resources." .
 
 vf:hasBeginning a           owl:DatatypeProperty ;
-        rdfs:domain         time:TemporalEntity ;
+        rdfs:domain         [ owl:unionOf (
+                                time:TemporalEntity
```
jvanbockryck: Should remove "time:TemporalEntity"

release-doc-in-process/all_vf.TTL
```diff
@@ -323,19 +323,57 @@ vf:containedIn a            owl:ObjectProperty ;
...
 
+vf:hasPointInTime a         owl:DatatypeProperty ;
+        rdfs:domain         [ owl:unionOf (
+                                time:Instant
+                                vf:Commitment
+                                vf:EconomicEvent
+                                vf:Intent) ] ;
+        rdfs:range          xsd:dateTimeStamp ;
+        vs:term_status      "unstable" ;
+        owl:equivalentProperty  time:inXSDDateTimeStamp ;
```
jvanbockryck: Should remove this line and replace with:
---
owl:propertyChainAxiom (time:hasEnd time:inXSDDateTimeStamp) ;

release-doc-in-process/all_vf.TTL
```diff
@@ -323,19 +323,57 @@ vf:containedIn a            owl:ObjectProperty ;
...
+        rdfs:comment        "The planned or actual time of a flow; can be used instead of hasBeginning and hasEnd, if so, hasBeginning and hasEnd should be able to be returned with this value." .
+
+time:hasDuration
+        rdfs:domain         [ owl:unionOf (vf:RecipeProcess vf:ScenarioDefinition) ] .
+
+vf:due a                    owl:DatatypeProperty ;
```
jvanbockryck: I thought we had "vf:before" instead of "vf:due", as we also have "vf:after"?

release-doc-in-process/all_vf.TTL
fosterlynn: You don't think we still want to have that "tunnel" into owl time?

release-doc-in-process/all_vf.TTL
fosterlynn: I wasn't sure the best way to do it.  My thinking was that if we can use hasBeginning and hasEnd to interface directly with owl time, and if we specify that those have to be returned if hasPointInTime is what was entered, then we could be good with either hasPointInTime or inXSDDateTimeStamp.  But I'm the first to admit I don't know what I'm doing on this one technically.

release-doc-in-process/all_vf.TTL
fosterlynn: I added vf:due as a totally separate thing from owl time stuff.  I don't actually know of much of a use case for vf:after so thought best to wait and see if there is one.  But maybe due isn't the best name anyhow, it is somewhat more specific.

release-doc-in-process/all_vf.TTL
jvanbockryck: OK then, i'm not religious about it :).

release-doc-in-process/all_vf.TTL
jvanbockryck: OK, I see. "vf:due" seems to have a more clear semantic. The we should either keep vf:before and vf:after and see how the user community will use it, or not.

fosterlynn: Yes, I agree on seeing how the user community thinks about it.  I plan on working as closely as possible with the current groups implementing VF so I can get feedback as close to the source as possible.  (Small note: We never had vf:before and vf:after, we just used time:before and time:after.)  I'm not clear which you prefer, so let's talk about this next time we chat.

fosterlynn: Decided to keep vf:due for now, pending more user involvement in current projects.

release-doc-in-process/all_vf.TTL
fosterlynn: done

pospi:
> My opinion is that we can use hasBeginning and hasEnd for that on observation and planning, and make sure the definition is clear.

Just for clarity- you mean that the appropriate usage of the event fields when recording things that happened in a loosely defined time period is to use `hasBeginning` and `hasEnd` to define the range of time in which the event is said to have occurred? And this is as inexact as we will support? (i.e. no unbounded "before" or "after")

fosterlynn:
>Just for clarity- you mean that the appropriate usage of the event fields when recording things that happened in a loosely defined time period is to use hasBeginning and hasEnd to define the range of time in which the event is said to have occurred? And this is as inexact as we will support? (i.e. no unbounded "before" or "after")

For events, yes.  For commitments or intents, yes, but due is also enough.

Thinking HoloREA for example, I"m thinking though that we don't necessarily want to require that both hasBeginning and hasEnd are recorded if one is recorded, although apps can of course do so.

Basically I think that different groups will have very different requirements for how strictly they schedule their work, also different for different kinds of work.  We can see how it works out in practice.

PR #593, opened 2019-10-29
Defined Actions as individuals, added configuration properties.

fosterlynn:
Action no longer a subclass of Property, so we can have configuration / metadata kinds of information for each action, which is now defined as a NamedIndividual.

I think the examples work as they are.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -18,13 +18,26 @@
     dcterms:status <http://purl.org/adms/status/UnderDevelopment>.
 
 
+#################################################################
+#    Classes
+#################################################################
+
 # ABSTRACT / CLASSIFICATION CLASSES (KNOWLEDGE LAYER)
 
 vf:Action  a                 owl:Class ;
         rdfs:label           "vf:Action" ;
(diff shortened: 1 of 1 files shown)


fosterlynn:
Fixed `accept` to be an input, thanks @pospi

release-doc-in-process/all_vf.TTL
```diff
@@ -666,109 +680,247 @@ vf:inverseLabel a           owl:DatatypeProperty ;
...
+
+vf:consume a                owl:NamedIndividual ,
+                            vf:Action ;
+        vf:inputOutput      vf:input ;
+        vf:pairsWith        vf:notApplicable ;
+        vf:resourceEffect   vf:decrement ;
+        rdfs:comment        "For example an ingredient or component composed into the output, after the process the ingredient is gone." ;
         rdfs:label          "consume" ;
-        vs:term_status      "testing" ;
-        rdfs:subPropertyOf  vf:decrement ;
```
jvanbockryck: This line should be removed:
---
rdfs:subPropertyOf  vf:decrement ;

fosterlynn: It's gone, the red and the '-' is a github convention for removed.

PR #595, opened 2019-10-30
removed state/stage from recipeflow and commitment

fosterlynn:
Thanks to @pospi for questioning this, these are in any case derivable, but keeping them on EconomicResource and removing from recipe/planning seems right.  @bhaugen please review.


pospi:
Please note I went ahead and updated this in https://github.com/valueflows/vf-graphql/commit/ad79c2c33377e4523bf269844ce854e7d649b8d3

Issue #299, opened 2018-11-27
Finish recipe


fosterlynn:
After PR https://github.com/valueflows/valueflows/pull/296 we will still need to add some features:
- ability to fork recipes
- integrate transfers as much as it makes sense
- support workflow recipes
- ???

Please feel free to add....

elf-pavlik:
> integrate transfers as much as it makes sense

I think in many cases we can use transportation as well. This keeps us away from 'changing rights' and just have resources moving between stock along all those transformations. Possibly wherever we can include transportation we can insert 'transfer' (teleportation) as well.

fosterlynn:
I thought about integrating exchanges, since that is needed now.  I think we can do it very simply, thanks to the simplification of transfers into individual events.  In the below diagram, I just added RecipeExchange and the relationship to RecipeFlow.

![exch-recipe](https://user-images.githubusercontent.com/3776081/69007909-c30d3600-0909-11ea-85b2-0352bbab8602.png)

Notes:
* This will create an Agreement.
* I don't see the need for a "specification" since it seems only useful as part of a recipe.
* In NRP, we had some configuration elements, but I propose to wait and see what is actually needed.  All of it was basically internal to an app and used to simplify the UI, I don't think needed as part of the vocabulary used to communicate between agents or apps.

@jvanbockryck @bhaugen I would love some feedback.

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

> https://lab.allmende.io/valueflows/valueflows/-/issues/299.

If you have not done so, you are very welcome to register at https://lab.allmende.io and join the ValueFlows organization there.

PR #600, opened 2019-11-17
added exchange into the recipe

fosterlynn:
The RecipeExchange would create what we currently call an Agreement.   It can be part of a flow which included processes, or not.  This includes a simple example.  I will try to add a more complex one that includes Intents generated from recipe, as we discuss this.  See also https://github.com/valueflows/valueflows/issues/299#issuecomment-554744775.


release-doc-in-process/all_vf.TTL
```diff
@@ -73,6 +73,11 @@ vf:RecipeFlow a             owl:Class ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The specification of a resource inflow to, or outflow from, a recipe process." .
```
bhaugen: RecipeFlow is now being used in recipe exchanges as well as recipe processes.

docs/examples/ex-exchange.md
```diff
@@ -55,3 +56,11 @@ Logged work triggers a claim for future income distribution for that work.
 
 [import, lang:"yaml"](../../examples/claim.yaml)
 
+
+#### Plan from recipe with production and exchange
+
+This shows a simple recipe that includes both a process and the exchange of work that is input to the process for currency.
+
+![process-exchange diagram](https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/process-exchange.png)
```
bhaugen: diagram seems to be missing.

PR #602, opened 2019-12-01
adding resource primary agent

fosterlynn:
Added accounting agent to the resource.  This is "owner" and whatever owner becomes in a less owner-oriented economic system, so I'm avoiding the term.  It is the agent that includes the resource in their accounting reports or UI as part of their inventory. 

Looking for the best name, suggestions welcome.  I have it as primaryAccountable.  Have also toyed with primaryAccounting, primaryResponsible.  Thought I had a better name earlier in discussion with @jvanbockryck but can't find it now.

This is a simple version of agent-resource relationships, user defined, such as we have used in NRP/OCP.  Right now I don't think we want that whole mechanism.  If we add something like that later, this will still be useful as a direct filter on resources.

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -344,7 +344,7 @@ vf:primaryAccountable a      owl:ObjectProperty ;
         rdfs:domain         vf:EconomicResource ;
         rdfs:range          vf:Agent ;
         vs:term_status      "unstable" ;
-        rdfs:comment        "The agent currently with primary rights and responsibilites for the economic resource. This agent will normally do the accounting for the resource." .
+        rdfs:comment        "The agent currently with primary rights and responsibilites for the economic resource.  It is the agent that includes the resource in their accounting reports or inventory." .
 
 vf:hasBeginning a           owl:DatatypeProperty ;
         rdfs:domain         time:TemporalEntity ;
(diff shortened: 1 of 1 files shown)


release-doc-in-process/all_vf.TTL
```diff
@@ -339,6 +339,13 @@ vf:containedIn a            owl:ObjectProperty ;
         vs:term_status      "unstable" ;
         rdfs:comment        "Used when a stock economic resource contains units also defined as economic resources." .
 
+vf:primaryAccountable a      owl:ObjectProperty ;
+        rdfs:label          "primary accountable" ;
+        rdfs:domain         vf:EconomicResource ;
+        rdfs:range          vf:Agent ;
+        vs:term_status      "unstable" ;
+        rdfs:comment        "The agent currently with primary rights and responsibilites for the economic resource. This agent will normally do the accounting for the resource." .
```
bhaugen: The description in the overview of this PR is more clear. Suggest changing
> This agent will normally do the accounting for the resource.

to
>  This is the agent that includes the resource in their accounting reports and/or as part of their inventory.

Or something like that...maybe mention accountingQuantity? "Do the accounting" could mean "be the bookkeeper or accountant".

fosterlynn: @bhaugen please check the updated definition.

Issue #610, opened 2020-02-10
Pass/Fail and States


fosterlynn:
This is related to https://github.com/valueflows/valueflows/issues/603 but I think involves a different decision point.  That one is about how to define measurement and test events themselves; this is to reconsider how to define the implications of measurements and tests in terms of economic events and economic resources.

Noting that we currently have `pass` and `fail` actions, which become the state of the resource referenced by the event.  And noting that we have an active use case of a beef supply chain which includes longish transportation and various quality or validity tests, some of which go beyond pass or fail.  For example, various tolerances of temperature readings may not "fail" the resource, but change the expiration date, which changes the price (and possibly the product, but that is part of this discussion, and we don't have full details on how they look at it).

So some questions: 
* What actions do we want?  
* Do we want states, and if so how and by whom are they defined? 
* What are the increment/decrement processing rules for something that fails or tests not optimal in some way?  
* What causes a new resource?  
* What causes a new resource specification?

fosterlynn:
Thinking out loud through some scenarios. 

Production:  I think this issue applies pretty much to "workflow" sequences of processes, where the same resource is input and output if all goes well.  But am including `produce` for completeness.  If something fails during production, it becomes split off into a separate resource with mostly a different resource specification.  Like, you drop one of 30 pies, you output 29 pies, and 2 lbs of compost.  So, you get "co-products" or "by-products".

Transportation:  The beef use case fits in here.  There are temperature readings taken throughout transportation in different parts of a container.  All or part of a resource might be failed as unconsumable (if it is a frozen product that thaws); or might be down-graded but still tradeable (if it is a cooled product that become too warm, but within some tolerances).  These are not separate quality testing processes, they are part of the transportation process, with `pickup` and `dropoff` actions.  Or in ground transportation, for example, sometimes one package of many will disappear from the truck, so there is no `dropoff` or a `dropoff` with less quantity.  We think that splitting off of failed or partially failed resources will usually happen on receipt, after the drop off (unloading).  But either way, I think it basically gets split into separate resources that probably have different resource specifications.

[...]

PR #611, opened 2020-02-26
Removed pass/fail as actions; made state a string

fosterlynn:
Ref. https://github.com/valueflows/valueflows/issues/610, https://github.com/holo-rea/holo-rea/issues/130.

This is not a final solution, just removes what we have found is too limited.  The hope is to get some more clarity during current implementations, and come up with a more stable and robust solution from that.


pospi:
This looks good, and makes sense. I'd like to propose a `newInventoriedResourceState` field be added to `EconomicEvent` that just accepts a string, and assigns that value to `EconomicResource.state` if present. That way, we've got everything needed to have non-VF business logic affecting resource states in any arbitrary manner.

Issue #615, opened 2020-04-09
Question on canonical schema(s) (JSON vs OWL)


orthecreedence:
Hello! In an effort to start playing with VF in code more, I built a rust library that converts the vf-json-schema into (de)serializable rust structs: https://docs.rs/vf-rs/. After doing so, I realized there is also an [owl schema](https://raw.githubusercontent.com/valueflows/valueflows/master/release-doc-in-process/all_vf.TTL). I'm fairly unfamiliar with owl (and ontology in general). Is the owl schema more "canonical" than the JSON or are they kept in sync pretty well? I would like to figure out how to parse the owl into structs because I'd like to include OM measurements in the library (or as a sister library) and OM uses owl as well, but if the JSON and owl schemas are in parity I might save that project for another time.

Also, I noticed that `vf:cite` shows up in the owl schema twice. Is that purposeful? Thanks!

orthecreedence:
Ok, after looking through the generated code for the rust lib, I'm seeing what appears to be an issue. Various structs (for example [EconomicEvent](https://docs.rs/vf-rs/0.1.2/vf_rs/economic_event/struct.EconomicEvent.html)) have a field `action` which seems like it should be a type [Action](https://docs.rs/vf-rs/0.1.2/vf_rs/action/struct.Action.html) but is being labeled in the JSON schema as a string. Should the `action` fields in `EconomicEvent`, `Commitment`, `Intent`, etc be `$ref` types that point to `https://raw.githubusercontent.com/valueflows/valueflows/master/json-schemas/Action.json`? If not, and they're supposed to point at the Action type's [Label](https://docs.rs/vf-rs/0.1.2/vf_rs/action/enum.Label.html) enum then does it make sense to break `Action.label` out into its own type in the schema?

fosterlynn:
Hi @orthecreedence very nice to see you here!  And apologies for the json-schema gaps. 

The owl schema is actually the "system of record" "canonical" model, but it only includes the VF namespace, so is missing a number of things that are required for VF but we are attempting to re-use from other specs like OM (they are [all here](https://valueflo.ws/specification/external-terms.html)).  The json-schemas should include everything that makes the VF model hang together.  However, they are not being actively used, so we have let them get a bit behind; and also because of that they have never been really tested.  (Some day we'll automate creating them, but somehow need to include the non-VF-namespace elements.)  In any case, I will fix them if you want to use them!  And will create a separate issue anyhow so we don't forget, thanks.

[...]

PR #619, opened 2020-04-13
add PairsWith as a class, include some actions

fosterlynn:
Ref. https://github.com/valueflows/valueflows/issues/615#issuecomment-613091091

Actually, I like this solution a lot better now that I have realized we only want a few actions included.  So thanks @orthecreedence !


orthecreedence:
This looks great! Exactly what I was thinking. Thanks so much!

PR #622, opened 2020-04-16
Discussion PR: remove `vf:notApplicable` and `vf:PairsWith` in owl schema

orthecreedence:
Hi! After [this pr](https://github.com/valueflows/valueflows/pull/619) and the [included discussion](https://github.com/valueflows/valueflows/issues/615#issuecomment-613091091) I started working with the vf model more (converting to rust) and realized something I'd like to bring up for discussion. Keep in mind, these modifications took all of 30s to make, so if ultimately the decision is made to reject this PR I will not be hurt at all.

Effectively I think it makes sense to remove `vf:notApplicable` and `vf:PairsWith`. When converting the rdf spec into rust, I ran into issues where essentially I was hardcoding logic like "if this contains `vf:notApplicable` then the type should be able to return NULL (`None` in rust)" and "if we encounter `vf:PairsWith` ignore it entirely because in reality it's a subset of `vf:Action` which is already defined."

Once logic like this starts sneaking, my first instinct is to ask "am I doing something wrong?"

Then I realized a way this could be done without hardcoding. Remove all instances of `vf:notApplicable`:

```ttl
vf:transfer-custody a       owl:NamedIndividual ,
                            vf:Action ;
        vf:inputOutput      vf:notApplicable ;
        vf:pairsWith        vf:notApplicable ;
        vf:resourceEffect   vf:decrementIncrement ;
        rdfs:comment        "Give physical custody and control of a resource, without full accounting or ownership rights." ;
        rdfs:label          "transfer-custody" ;
        vs:term_status      "unstable" .

vf:use a                    owl:NamedIndividual ,
                            vf:Action ;
        vf:inputOutput      vf:input ;
        vf:pairsWith        vf:notApplicable ;
        vf:resourceEffect   vf:noEffect ;
        rdfs:comment        "For example a tool used in process; after the process, the tool still exists." ;
        rdfs:label          "use" ;
        vs:term_status      "unstable" .
```

becomes 

```ttl
vf:transfer-custody a       owl:NamedIndividual ,
                            vf:Action ;
        vf:resourceEffect   vf:decrementIncrement ;
        rdfs:comment        "Give physical custody and control of a resource, without full accounting or ownership rights." ;
        rdfs:label          "transfer-custody" ;
        vs:term_status      "unstable" .

vf:use a                    owl:NamedIndividual ,
                            vf:Action ;
        vf:inputOutput      vf:input ;
        vf:resourceEffect   vf:noEffect ;
        rdfs:comment        "For example a tool used in process; after the process, the tool still exists." ;
        rdfs:label          "use" ;
        vs:term_status      "unstable" .
```

Given that the `ObjectProperties` for `vf:pairsWith` and `vf:inputOutput` are defined on the enum, the implementor can now say `Action::TransferCustody` should be able to return a `pairs_with()` or `input_output()` value *but they are not defined in all cases* therefor it makes sense to allow the return types to be nullable (which is effectively what the hardcoded `vf:notApplicable` logic was doing anyway).

By extension, if `vf:notApplicable` becomes redundant, `vf:PairsWith` outlives its usefulness as well (since it only exists to union `vf:notApplicable` with a subset of `vf:Action`).

The only part that I don't know is if this is idiomatic owl/rdf/etc or if this would hurt the undertanding of others trying to parse the schema. If you define an `ObjectProperty` on an enum value, is it "ok" for the enum values to not return a value for that property?

Hopefully this all makes sense, I can expand further if needed, and like I said, want this to be more of a discussion. I thought a PR instead of an issue might be appropriate since I know exactly what change I'm envisioning.

Thanks!

### DIFF
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -347,7 +347,7 @@ vf:containedIn a            owl:ObjectProperty ;
 vf:primaryAccountable a      owl:ObjectProperty ;
         rdfs:label          "primary accountable" ;
         rdfs:domain         vf:EconomicResource ;
-        rdfs:range          vf:Agent ;
+        rdfs:range          foaf:Agent ;
         vs:term_status      "unstable" ;
         rdfs:comment        "The agent currently with primary rights and responsibilites for the economic resource. It is the agent that is associated with the accountingQuantity of the economic resource." .
 
--- a/release-doc-in-process/all_vf.TTL
+++ b/release-doc-in-process/all_vf.TTL
@@ -33,11 +33,6 @@ vf:Action  a                 owl:Class ;
         vs:term_status       "testing" ;
         rdfs:comment         "An action verb defining the kind of flow and its behavior." .
 
-vf:PairsWith a              owl:Class ;
-        rdfs:comment        "The action that should be included on the other direction of the process, for example accept with modify; also includes not applicable." ;
-        rdfs:label          "vf:PairsWith" ;
-        vs:term_status      "unstable" .
-
 vf:InputOutput a            owl:Class ;
         rdfs:comment        "The action is an input or output of a process, or not related to a process." ;
         rdfs:label          "vf:InputOutput" ;
@@ -849,7 +844,6 @@ vf:accept a                 owl:NamedIndividual ,
 vf:consume a                owl:NamedIndividual ,
                             vf:Action ;
         vf:inputOutput      vf:input ;
-        vf:pairsWith        vf:notApplicable ;
         vf:resourceEffect   vf:decrement ;
         rdfs:comment        "For example an ingredient or component composed into the output, after the process the ingredient is gone." ;
         rdfs:label          "consume" ;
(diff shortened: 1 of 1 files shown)


fosterlynn:
@orthecreedence thanks much for this!  The same question flitted across my mind on the PairsWith PR.  I don't know, actually.  Your suggestion doesn't raise any flags for me generally, and what you are saying makes sense.  It might raise flags in semantic web land because theoretically you never know if a missing thing has a specific meaning - you have to just assume you don't know.  But I don't have enough experience in the semantic web.  But I do have a feeling that VF has to have more stringent requirements and definition than many in the semantic web - the "anyone can say anything about anything" doesn't really apply to economic activity imo.

Pinging @mayel @ivanminutillo @pospi - have you implemented Action yet, and if so, do you have input? Which reminds me, I didn't change the graphql to match the last fix on Action in VF core, which was a bug in the rdf definition.... what we are talking about is this https://github.com/valueflows/valueflows/pull/619.

cc @elf-pavlik @jvanbockryck if you're watching and are willing, please offer up a rdf/owl opinion!

pospi:
Holo-REA has action, but it doesn't impact anything if it keeps the enumerated values internally. We might want to change some of the GraphQL though? We could certainly get rid of the `notApplicable` values there as well, and just use nulls.

almereyda:
We have moved the ValueFlows organization from GitHub to https://lab.allmende.io/valueflows.

This issue has been closed here, and all further discussion on this issue can be done at

> https://lab.allmende.io/valueflows/valueflows/-/issues/622.

If you have not done so, you are very welcome to register at https://lab.allmende.io and join the ValueFlows organization there.

